PR #2, opened 2024-08-01
fix: grabbing endpoints without an authorization header

ATNoblis:
Fixing application if user information is not able to be returned to the user, provide and empty user without breaking the application with an error page.

### DIFF
--- a/src/views/Title/Title.tsx
+++ b/src/views/Title/Title.tsx
@@ -18,9 +18,20 @@ interface LoaderData {
   preferred_username?: string
   groups?: string[]
 }
+const emptyUser: LoaderData = {
+  email: '',
+  name: '',
+  preferred_username: '',
+  groups: [],
+}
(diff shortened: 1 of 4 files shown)


src/views/Title/Title.tsx
```diff
@@ -18,9 +20,19 @@ interface LoaderData {
...
+  email: '',
+  name: '',
+  preferred_username: '',
+  groups: [],
+}
 const Title: React.FC = (): JSX.Element => {
   const loaderData = useLoaderData() as string
-  const userInfo = jwtDecode(loaderData) as LoaderData
+  const userInfo: LoaderData =
+    loaderData === 'unauthorized\n'
```
talentedmrjones, line 32: Can you check against the response status instead? If it is not 200 then its an error

talentedmrjones (changes requested):

talentedmrjones (approved):
LGTM

PR #20, opened 2024-08-15
feat: Questionnaire Popup Modal & System Table

ATNoblis:
## Summary
Adding a questionnaire popup modal for users to view their questionnaires for data calls.
Closes: #24 #20 #26

### Added
Questionnaire Popup Modal.
Table that displays systems

### Changed

Changed to SPA without any additional pages. 

### Deprecated

None

### Removed

Removed library to access graphql since backend utilizes REST.
Removed routings to pillar pages. The breakdown of each pillar of selected system


src/router/authLoader.ts
```diff
@@ -1,21 +1,33 @@
...
  * @see {@link dashboard/Routes}
  */
 
+const emptyUser: userData = {
+  userid: '',
+  email: '',
+  fullname: '',
+  role: '',
+  assignedfismasystems: [],
+}
```
talentedmrjones, line 16: do you really need this? it exposes the shape of api data to the public do I would rather avoid it

src/types.ts
```diff
@@ -61,6 +61,61 @@ export type FormField = {
   value?: string | number | boolean | null
   component: React.ElementType
 }
+export type userData = {
```
talentedmrjones, line 64: and what about these types? do they need to be declared here out in the open?

src/views/Title/Title.tsx
```diff
@@ -1,41 +1,34 @@
...
+const emptyUser: userData = {
+  userid: '',
   email: '',
-  name: '',
-  preferred_username: '',
-  groups: [],
+  fullname: '',
+  role: '',
+  assignedfismasystems: [],
 }
```
talentedmrjones, line 22: and here

talentedmrjones (changes requested):

talentedmrjones (approved):

Issue #32, opened 2024-08-29
feat: Prepare question with previous answer, if applicable


ATNoblis:
Pre-populate the radio button with the previous answer to questions if applicable.

PR #34, opened 2024-09-03
Prepopulate Questionnare

ATNoblis:
## Summary
Prepopulate the questionnare if the user had previously answered the question. Check to corresponding radio button to the previous answer as well as the notes that was previously written.

Closes: #32 

### Added

None

### Changed

Changed the questionnaire modal file, types, and removed unused methods/dependencies from the title.

### Deprecated

None

### Removed

None

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -111,6 +111,14 @@ export type QuestionOption = {
   score: number
 }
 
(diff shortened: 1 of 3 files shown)


talentedmrjones (approved):
LGTM

Issue #49, opened 2024-09-09
bug: API responses are not working with changes from the backend


ATNoblis:
returned API responses are encapsulated in another object which needs to be handled by the front-end

PR #50, opened 2024-09-09
Fixing api calls with changes with the backend 

ATNoblis:
### Summary

API responses are encapsulated in another object thus the frontend is not getting the intended data from the API to use for UI.

closes: #49 

### Added

None

### Changes

3 files where fetching the API

### Deprecated

None

### Removed

None


talentedmrjones (approved):
:shipit:

PR #52, opened 2024-09-11
fix: reroute user to login via a button on the page. 

### LINKED ISSUE #44: bug: redirect user to login if endpoint status isn't 2xx
Users should be redirected to login as it is most likely that their session has expired. This causes endpoints to return anything but code 2xx which application is stuck in a loading state.

ATNoblis:
### Summary

Fixing the infinite looped caused by trying to grab a valid user. If the user is not valid, allow them to login via a button

closes #44
### Added

None

### Changed

Added a button to home page. If an invalid response code was returned, allow the user to bedirected to login

### Removed

None

### Deprecated

None

### DIFF
--- a/src/router/authLoader.ts
+++ b/src/router/authLoader.ts
@@ -14,7 +14,7 @@ const emptyUser: userData = {
   role: '',
   assignedfismasystems: [],
 }
-const authLoader = async (): Promise<unknown> => {
+const authLoader = async () => {
   try {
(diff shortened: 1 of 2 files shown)


talentedmrjones:
Let's hide the user icon in the top right until logged in

ATNoblis:
> Let's hide the user icon in the top right until logged in

user icon is now hidden

talentedmrjones (approved):

Issue #97, opened 2024-11-15
ui: provide summary of each pillar as tooltip on hover


talentedmrjones:
in the questionaire model, when user hovers over function title info icon

Issue #81, opened 2024-10-21
any api calls resulting in 401 or 403 pushes to login


talentedmrjones:
Currently only the `/user/current` path pushed to login, but _all_ URIs returning 401 or 403 should push to login. This will happen when the session cookie expires while logged in.

PR #99, opened 2024-11-21
feat: redirect users to login when endpoint throws a 4XX status

ATNoblis:
### Summary

Refactoring code to redirect users to sign in on all api calls such that if status is a 4XX number to a sign in page.
Added function description for each function in pillar with a info icon that users can hover.

closes: #97 
closes: #81 

### Added

Login.tsx - a component that contains a reference to login endpoint. Moved previous sign in to a reusable 

### Changes

Files that call an endpoint

### Test

Open questionnare modal to see the info icon. Hover to see description of function.

### DIFF
--- a/src/views/QuestionnareModal/QuestionnareModal.tsx
+++ b/src/views/QuestionnareModal/QuestionnareModal.tsx
@@ -95,6 +95,7 @@ export default function QuestionnareModal({
   const [questionId, setQuestionId] = React.useState<number | null>(null)
   const [categories, setCategories] = React.useState<Category[]>([])
   const [options, setOptions] = React.useState<QuestionOption[]>([])
+  const [datacallID, setDatacallID] = React.useState<number>(0)
   const [loadingQuestion, setLoadingQuestion] = React.useState<boolean>(true)
   const [questionScores, setQuestionScores] = React.useState<questionScoreMap>(
     {}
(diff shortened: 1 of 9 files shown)


src/views/QuestionnareModal/QuestionnareModal.tsx
```diff
@@ -253,10 +289,24 @@ export default function QuestionnareModal({
...
           console.error('Error fetching data:', error)
+          navigate(Routes.SIGNIN, {
+            replace: true,
+            state: {
+              message: 'Your session has expired. Please log in again.',
+            },
+          })
         })
       axiosInstance
         .get(`scores?datacallid=2&fismasystemid=${system.fismasystemid}`)
```
talentedmrjones, line 300: should not hard code datacall id

src/views/QuestionnareModal/QuestionnareModal.tsx
```diff
@@ -253,10 +289,24 @@ export default function QuestionnareModal({
...
         .then((res) => {
+          if (res.status !== 200 && res.status.toString()[0] === '4') {
+            navigate(Routes.SIGNIN, {
+              replace: true,
+              state: {
+                message: 'Your session has expired. Please log in again.',
+              },
```
talentedmrjones, line 307: code is very repetitive. As it is, if the message needed to be changed, you would have to do it in many places. I would move all of these state message objects to a single function that gets called. It would accept a parameter to indicate whether to prepend "your changes were not saved"

talentedmrjones (changes requested):

talentedmrjones (approved):
LGTM

PR #245, opened 2025-03-27
feat: email modal and email sent modal

ATNoblis:
### Summary

email modal that allows an admin to sent out emails to groups associated with the application

When an admin sends an email, a button will show up and the admin can check the list of users the email has been sent to

### Test

Click on the kebab menu on the top right corner. Click on send email and 3 required fields must be field before submitting. After a valid email has been sent, a button will appear to allow an admin to check on the list of users the email has been sent out to.

### DIFF
--- a/src/components/EmailModal/EmailModal.css
+++ b/src/components/EmailModal/EmailModal.css
@@ -1,0 +1,22 @@
+.group-label {
+  margin-top: 0;
+  margin-bottom: 0;
+}
+
+
+.subject-label-text {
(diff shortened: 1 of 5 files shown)


talentedmrjones (changes requested):
the form should not default to an error state before even giving a user the chance to provide the information. Required fields can be marked with a "*" or "required" note in gray, but should only turn red and if they are not provided at the time the user presses send

talentedmrjones (approved):

PR #261, opened 2025-04-10
feat: text to indicate current datacall

TranScripted:
### Summary

Provide user context of which datacall they are currently viewing. Refactored code across application including using latest api endpoint, removing multiple calls to api, and ways of handling errors coming from api


eschwein-usds (approved):
Looks good to me.

Issue #259, opened 2025-04-04
ui disables answering questions when datacall deadline has passed


talentedmrjones:
When viewing the questionnaire after the data call deadline has passed, radio buttons should be disabled, and a note displayed saying "The deadline for this datacall has passed and can no longer be updated."

jsos3-cms:
Needs Checking/Confirmation

PR #264, opened 2025-04-11
feat: disable options and notes passed the deadline

ATNoblis:
### Summary

Disable the questionnaire options and notes when the current date is passed the deadline. In addition, pressing the next button does not trigger an update to the backend unless deadline

Refactored all the api endpoints to consider looking at error number `401`, `403`, and anything other error to be handled accordingly across the application.

### Test

Open questionnaire for a fisma system
if the deadline has been passed today, there will be a message above the function that indicates and the options and notes will be disabled


jsos3-cms:
This will close #259 . Need to rebase and update.

PR #265, opened 2025-04-17
fix: questionnaire resolves issue where next is disabled on last question

talentedmrjones:
## Added
none

## Changed
next button says complete when on last question and pushes user to first question


eschwein-usds (approved):
This looks right, and I like the changes to the error message.

PR #286, opened 2025-04-28
chore: remove dependabot

talentedmrjones:
## Added
none

## Changed
removed dependabot because it just creates a ton of PRs that all fail and never get merged


eschwein-usds (approved):
Thanks

PR #330, opened 2026-02-19
Add SDL sync toggle for CMS data lake export control

danielbowne:
## Summary

- Adds frontend support for the `sdl_sync_enabled` field on FISMA systems, allowing admins to control whether a system's data is exported to the CMS Snowflake data lake
- Admins can toggle SDL sync on/off from both the Edit System modal and the System Detail edit view
- The System Detail read view displays the current SDL sync status as a read-only chip (On / Off / Not configured)
- New systems default to sync disabled; existing systems reflect whatever value the backend provides
- Extracts a shared `SdlSyncToggle` component and centralizes SDL-related constants to reduce duplication

## Additional changes

- Table action icon updated from edit pencil to view icon — clicking now navigates to the system detail page in read mode rather than jumping straight to edit mode
- Removed the `?edit=true` query param auto-edit flow from `SystemDetailPage` (dead code after the navigation change)

## Files changed

| File | Change |
|---|---|
| `src/types.ts` | Add `sdl_sync_enabled: boolean \| null` to `FismaSystemType` |
| `src/constants.ts` | Add SDL sync description strings and switch styling constant |
| `src/components/SdlSyncToggle/SdlSyncToggle.tsx` | New shared toggle component |
| `src/views/EditSystemModal/emptySystem.ts` | Default `sdl_sync_enabled: false` for new systems |
| `src/views/EditSystemModal/EditSystemModal.tsx` | Add toggle + include field in PUT/POST payloads |
| `src/views/FismaTable/FismaTable.tsx` | Update action icon to view, navigate to detail page |
| `src/views/SystemDetailPage/SystemDetailPage.tsx` | Include field in PUT payload, pass toggle handler, remove dead `?edit=true` effect |
| `src/views/SystemDetailPage/SystemDetailEditView.tsx` | Add Data Lake Export card with toggle |
| `src/views/SystemDetailPage/SystemDetailReadView.tsx` | Add read-only Data Lake Export card with status chip |

## Test plan

- [x] Verify FismaTable loads without the SDL Sync column (removed by design)
- [x] Open system detail page via table view icon — confirm it opens in read mode with SDL sync status displayed
- [x] Click Edit, toggle SDL sync on/off, save — confirm value persists on reload
- [x] Create a new system — confirm SDL sync defaults to Off
- [x] Verify READONLY_ADMIN users see the status chip but cannot access the toggle
- [x] Confirm no regressions on decommission flow or other system edit fields

### DIFF
--- a/src/components/SdlSyncToggle/SdlSyncToggle.tsx
+++ b/src/components/SdlSyncToggle/SdlSyncToggle.tsx
@@ -1,0 +1,41 @@
+import { Box, FormControlLabel, Switch, Typography } from '@mui/material'
+import {
+  SDL_SYNC_DESCRIPTION_ON,
+  SDL_SYNC_DESCRIPTION_OFF,
+  SDL_SYNC_SWITCH_SX,
+} from '@/constants'
+
+interface SdlSyncToggleProps {
+  checked: boolean
+  onChange: (checked: boolean) => void
+}
+
+export default function SdlSyncToggle({
+  checked,
+  onChange,
+}: SdlSyncToggleProps) {
+  return (
+    <Box>
+      <FormControlLabel
+        control={
+          <Switch
+            checked={checked}
+            onChange={(e) => onChange(e.target.checked)}
+            sx={SDL_SYNC_SWITCH_SX}
+          />
+        }
+        label={
+          <Typography variant="body2" sx={{ fontWeight: 500 }}>
+            Sync to SDL (Snowflake)
+          </Typography>
+        }
+      />
+      <Typography
+        variant="caption"
+        sx={{ display: 'block', ml: 4, color: 'text.secondary' }}
+      >
+        {checked ? SDL_SYNC_DESCRIPTION_ON : SDL_SYNC_DESCRIPTION_OFF}
+      </Typography>
+    </Box>
+  )
+}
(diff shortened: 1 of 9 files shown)


jsos3-cms (approved):
So, conditional approval. First off, this looks good! I like the implementation and the modal.

One question I had was around the per-system sync. Is there ever a case where a system is pushed to both SDL and HHS? I'm thinking in a migration of the backend storage, it could be useful to have both sync processes firing. 
Another question is; is there any notification to the user if the system is not configured to store/sync information? Asked another way, is it possible to have a misconfigured sync, or no sync at all, during the data call?

danielbowne:
> So, conditional approval. First off, this looks good! I like the implementation and the modal.
> 
> One question I had was around the per-system sync. Is there ever a case where a system is pushed to both SDL and HHS? I'm thinking in a migration of the backend storage, it could be useful to have both sync processes firing. Another question is; is there any notification to the user if the system is not configured to store/sync information? Asked another way, is it possible to have a misconfigured sync, or no sync at all, during the data call?

This is exclusively a manual place holder until this is automated on the condition we do integrate with HHS and they do not have an SDL or security datalake or in the event a CMS system owner does not want their information reported on in SDL.  It can only be enabled or disabled by an Admin.

PR #346, opened 2026-04-08
Remediate dependency vulnerabilities and modernize toolchain

danielbowne:
## Summary

Fixing the vite path-traversal CVE required bumping from v5 to v6, which pulled the thread on several interconnected peer dependency chains. Rather than leave a trail of warnings and deprecations behind, this PR addresses everything end-to-end:

- **Resolves all 7 open Dependabot security alerts** (defu, lodash, lodash-es, tar, brace-expansion, @tootallnate/once, vite) — yarn audit now returns zero findings
- **Upgrades Storybook 7 → 8** to satisfy vite 6 peer requirements and clear three deprecated-package audit warnings (@storybook/addon-styling, @storybook/testing-library, @storybook/client-api)
- **Migrates ESLint 8 → 9 flat config** to clear the eslint deprecation audit finding and bring all eslint plugins to current versions
- **Removes dead dependencies** (apollo-client v2, @testing-library/react-hooks, @types/testing-library__jest-dom, eslint-plugin-node) and adds the missing graphql peer dep for @apollo/client

The production build output is byte-identical to main — no application code changed beyond a type cast fix in BreadCrumbs and a lint directive in a skipped test file.

## Test plan

- [x] Verify `yarn install` completes with no project-level peer dependency warnings
- [x] Verify `yarn npm audit --all` returns no security advisories
- [x] Verify `npx tsc --noEmit` passes
- [x] Verify `npx vite build` succeeds
- [x] Verify `npx eslint ./` passes with zero errors and zero warnings
- [x] Verify CI pipeline passes (typecheck, lint, test, build)
- [x] Smoke test the deployed dev environment for login flow and page navigation


MackOverflow:
Looks good Danny, quick scan flagged a couple of clean up items that might be good to address eventually, just so this stays agile as we go down the line: 

1. eslint-plugin-import is still in devDependencies but not used in the new config

It's listed at ^2.32.0 in package.json but never imported in eslint.config.mjs. This is dead weight — it should either be removed or wired into the flat config if its rules are still wanted.

2. Deprecated typescript-eslint rules in eslint.config.mjs

Two rules that were removed/renamed in @typescript-eslint v8 are still present:

@typescript-eslint/semi — removed in v8 (handled by Prettier already, so just delete the rule)
@typescript-eslint/no-var-requires — renamed to @typescript-eslint/no-require-imports in v8 (already added in the CJS override, so just delete the old name)
These may not error now since the config spreads tsPlugin.configs.recommended.rules, but they're no-ops at best and could cause warnings.

3. Legacy plugin import pattern for @typescript-eslint

The config imports @typescript-eslint/eslint-plugin and @typescript-eslint/parser separately — this is the v7 pattern. The v8 recommended approach is to use the unified typescript-eslint package:

import tseslint from 'typescript-eslint'

This isn't broken, but it's the compat shim path and may cause issues on future updates. Low priority — can be a follow-up.

4. Transitive dependency overrides in resolutions may need maintenance

The resolutions block now pins 6 transitive deps (defu, lodash-es, tar, brace-expansion, @tootallnate/once, lodash). These are correct for the current vulnerability set, but they'll go stale as upstream packages update. Consider adding a comment in package.json or a TECH-DEBT note to revisit these periodically, or remove them once upstream catches up.

5. lodash resolution changed from ^4.18.0 to 4.18.1 (exact pin)

The old resolution used a caret range; the new one is an exact version pin. This is fine for now but means it won't auto-bump on yarn install. Intentional?

MackOverflow (dismissed):
See comments, this is g2g, but I would recommend some clean up when you have time to make sure nothing gets stale down the line.

jsos3-cms (dismissed):
Adding a second one here. Looks good. We'll probably want to do a quarterly cleanup of our Deps.

MackOverflow (approved):

Issue #329, opened 2026-02-18
Frontend: Centralize auth error handling


danielbowne:
# Frontend: Centralize auth error handling

**Epic**: [Multi-IdP Authentication](https://github.com/CMS-Enterprise/ztmf/issues/150) (CMS-Enterprise/ztmf#150)
**Can be done independently** — improves codebase regardless of multi-IdP

## Summary

Replace the duplicated 401 handling across 11 files (36 duplicate blocks) with a centralized Axios response interceptor in `axiosConfig.ts`. This reduces code duplication and makes it possible to add IdP-specific error messages in one place.

## Current State

`src/axiosConfig.ts` creates a basic Axios instance with `baseURL: '/api/v1/'` and a local dev token injection — but **no response interceptor**. Every component independently catches 401 errors and navigates to the sign-in page.

### Duplication Map (36 blocks across 11 files)

| File | 401 handling blocks |
|------|:---:|
| `src/views/QuestionnareModal/QuestionnareModal.tsx` | 10 |
| `src/views/QuestionnairePage/QuestionnairePage.tsx` | 7 |
| `src/views/EditSystemModal/EditSystemModal.tsx` | 5 |
| `src/views/UserTable/UserTable.tsx` | 3 |
| `src/views/Title/Title.tsx` | 2 |
| `src/views/AssignSystemModal/AssignSystemModal.tsx` | 2 |
| `src/views/SystemDetailPage/SystemDetailPage.tsx` | 2 |
| `src/views/Home/Home.tsx` | 1 |
| `src/views/FismaTable/FismaTable.tsx` | 1 |
| `src/views/DatacallModal/DataCallModal.tsx` | 1 |
| `src/components/EmailModal/EmailModal.tsx` | 1 |

The pattern is copy-pasted boilerplate:
```typescript
if (error.response.status === 401) {
  navigate(Routes.SIGNIN, {
    replace: true,
    state: { message: ERROR_MESSAGES.expired },
  })
}
```

### Inconsistencies

The status checking is inconsistent across components:
- Some use `=== 401` (strict equality)
- Some use `== 401` (loose equality)
- Some use `res.status.toString()[0] === '4'` (catches all 4xx, not just 401)

An `ErrorBoundary` component exists at `src/components/ErrorBoundary.tsx` with a status 401 → `SESSION_EXPIRED_MESSAGE` mapping, but it only handles React Router error boundaries, not Axios HTTP errors.

## Requirements

[...]

Issue #328, opened 2026-02-18
Frontend: Landing page with email input and domain-based IdP routing


danielbowne:
# Frontend: Landing page with email input and domain-based IdP routing

**Epic**: [Multi-IdP Authentication](https://github.com/CMS-Enterprise/ztmf/issues/150) (CMS-Enterprise/ztmf#150)
**Priority**: Start here — no infrastructure or backend changes required

## Summary

Replace the current minimal login page (single "Login" button that auto-redirects to Okta) with a landing page that accepts an email address, extracts the domain, and redirects the user to the correct IdP-specific login path.

This is a frontend-only change that can be built and tested immediately. It prepares the UX layer for multi-IdP support while the infrastructure and backend work follows in later sub-tasks.

## Current State

`LoginPage.tsx` is minimal (renders inside `Title.tsx` when `loaderData.status !== 200`):
- Displays a dynamic message from `location.state?.message` (e.g., "Your session has expired" or empty)
- Single `<CmsButton href="/login">Login</CmsButton>` button
- Clicking it hits the ALB `/login` path which auto-redirects to Okta

Auth gating lives in `Title.tsx` (line 248): when `loaderData.status !== 200`, it renders `<LoginPage />` inline instead of the dashboard. The `authLoader` in `src/router/authLoader.ts` calls `/users/current` — on failure, returns `{ ok: false, response: emptyUser }` with no `status` field, so the `!== 200` check passes by evaluating `undefined !== 200`.

The `VITE_IDP_ENABLED` flag is declared in `src/utils/config.ts` as `CONFIG.IDP_ENABLED` but is **never referenced** anywhere in the application code — it's completely unused.

### Landing Page
- [ ] Email input field with "Enter your email to get started" prompt
- [ ] "Continue" button (disabled until valid email entered)
- [ ] Basic email format validation (client-side)
- [ ] On submit: extract domain from email address
- [ ] Route to `/login/<provider>` based on domain-to-provider mapping
- [ ] Show clear error message for unrecognized domains: "Your organization is not configured for ZTMF. Contact your administrator."

### Domain-to-Provider Mapping
- [ ] Configurable mapping, not hardcoded — use environment variable or config file
- [ ] Initial mapping: `cms.hhs.gov` → `/login/okta` (preserves current behavior)
- [ ] Structure should support adding Entra ID and other providers later
[...]

PR #356, opened 2026-05-18
feat(roles): expand frontend role taxonomy for multi-OpDiv backend (Stage 1)

danielbowne:
## Where this fits

The multi-OpDiv frontend work is rolling out in five stages. Naming follows the handoff doc at `tmp/frontend-stage1-handoff.md`.

| Stage | Scope | Status |
| --- | --- | --- |
| **1** | **This PR.** Role taxonomy expansion, type updates, helper module, `/api/v1/opdivs` client | Open |
| 2 | Multi-IdP auth, pre-auth lookup, `identity_provider` surfacing | Not started (`#328`, `#329`) |
| 3 | OpDiv selectors, badges, `opdiv_id` surfacing, UI consumers of `/api/v1/opdivs` | Not started |
| 4 | Grant management UI (`users_opdivs` assign/revoke) | Not started |
| 5 | Drop legacy `ADMIN` / `READONLY_ADMIN` once the backend stops returning them | Blocked on backend (`#357`) |

The backend runs a parallel "Stage A through D" sequence covering the same migration server-side. "Stage D" is the point at which the backend stops accepting and emitting the legacy role values.

## Summary

Stage 1 of the multi-OpDiv frontend rollout. Pairs with backend PR `CMS-Enterprise/ztmf#316` (branch `feature/multi-opdiv-schema`).

- Adds `UserRole` union covering the new tiers: `OWNER`, `HHS_ADMIN`, `HHS_READONLY_ADMIN`, `OPDIV_ADMIN`, `OPDIV_READONLY_ADMIN`. Legacy `ADMIN` / `READONLY_ADMIN` remain valid through the Stage D back-compat window.
- Introduces `src/utils/userRoles.ts` as the single source of truth for role predicates (`isAdmin`, `isReadOnlyAdmin`, `hasAdminRead`, `hasUnscopedRead`, `hasSystemAccess`, `isHHSTier`, `isOpDivTier`, plus role-string variants). Stage D legacy drop will touch one file.
- `isHHSTier` and `isOpDivTier` mirror the backend's `IsHHSTier` and `IsOpDivTier` helpers (renamed from `IsHHSAdmin` / `IsOpDivAdmin` to make explicit that they cover the read-only variant). Tier-membership check, not a write gate.
- Replaces 8 inline `userInfo.role === 'ADMIN' / 'READONLY_ADMIN'` literal sites across `Title`, `UserTable`, `SystemDetailPage`, `FismaTable`, `QuestionnairePage`, and `QuestionnareModal` with helper calls.
- Wires `GET /api/v1/opdivs` into the client layer via `fetchOpDivs` in `src/utils/opdivs.ts`. No UI caller yet; published so Stage 2+ work has a typed call site.
- Adds additive types for `identity_provider`, `assignedopdivids`, `FismaSystem.opdiv_id`, and a full `OpDiv` type. Frontend reads none of them at Stage 1.
- Adds `src/utils/userRoles.test.ts`: 73-case regression contract pinning the role taxonomy across every predicate, including explicit assertions on the legacy `ADMIN` / `READONLY_ADMIN` strings. Stage D removal will remove the corresponding test rows in the same PR, forcing the legacy drop to be an intentional, named change rather than a silent regression.

## Merge order

**Backend `CMS-Enterprise/ztmf#316` must merge and deploy to dev before this PR merges.** The new `identity_provider`, `assignedopdivids`, and `opdiv_id` fields plus the `/api/v1/opdivs` endpoint need to be live on the paired backend before the frontend types and client helpers have a real backing to consume. Frontend back-compat handles the legacy role strings so this PR is safe to sit open until the backend lands.

## Out of scope (later stages)

- Multi-IdP auth / pre-auth lookup endpoint — Stage 2 (`#328`, `#329`)
- OpDiv dropdown on system create/edit — Stage 3
- OpDiv badge or filter in list views — Stage 3
- Surfacing `opdiv_id` on system pages — Stage 3
- Grant management UI (assign/revoke users to OpDivs) — Stage 4
- Calling `fetchOpDivs` from any UI surface — Stage 2/3
- Surfacing `identity_provider` or `assignedopdivids` in UI — Stage 2
- Removing legacy `ADMIN` / `READONLY_ADMIN` from helpers — Stage 5

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn lint` clean
- [x] Pre-push hooks clean (typecheck + lint)
- [x] Independent code review — no critical findings
- [ ] After paired backend `CMS-Enterprise/ztmf#316` deploys to dev: smoke against current backend confirming the legacy `ADMIN` and `READONLY_ADMIN` paths still render admin UI; confirm `ISSO` continues to see only assigned systems.
- [ ] Once paired backend lands, re-verify the new role tiers (`OWNER`, `HHS_ADMIN`, `HHS_READONLY_ADMIN`, `OPDIV_ADMIN`, `OPDIV_READONLY_ADMIN`) each render the expected scope of UI. Confirm `/users/current` returns `identity_provider` and `assignedopdivids`; confirm `/opdivs` returns the reference list via `fetchOpDivs`.

## References

- Handoff doc: `tmp/frontend-stage1-handoff.md`
- Backend pair PR: `CMS-Enterprise/ztmf#316` (branch `feature/multi-opdiv-schema`)
- Related issues: `CMS-Enterprise/ztmf-ui#328`, `#329`, `CMS-Enterprise/ztmf-misc#170`, `#198`, `CMS-Enterprise/ztmf#266`

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -2,7 +2,7 @@
  * Application-wide constants
  * @module constants
  */
-import { userData } from './types'
+import { userData, UserRole } from './types'
 
 //* Application Strings
 export const ORG_NAME = 'CMS'
@@ -30,7 +30,7 @@ export const EMPTY_USER: userData = {
   userid: '',
   email: '',
   fullname: '',
-  role: '',
+  role: '' as UserRole,
   assignedfismasystems: [],
 }
 export const CONFIRMATION_MESSAGE =
--- a/src/router/authLoader.ts
+++ b/src/router/authLoader.ts
@@ -1,5 +1,5 @@
 import 'core-js/stable/atob'
-import { userData } from '@/types'
+import { userData, UserRole } from '@/types'
 import axiosInstance from '@/axiosConfig'
 /**
  * Auth state loader for react-router data routes.
@@ -11,7 +11,7 @@ const emptyUser: userData = {
   userid: '',
   email: '',
   fullname: '',
-  role: '',
+  role: '' as UserRole,
   assignedfismasystems: [],
 }
 const authLoader = async () => {
--- a/src/types.ts
+++ b/src/types.ts
@@ -20,12 +20,33 @@ export type FormField = {
   value?: string | number | boolean | null
   component: React.ElementType
 }
+export type UserRole =
+  | 'OWNER'
+  | 'HHS_ADMIN'
+  | 'HHS_READONLY_ADMIN'
+  | 'OPDIV_ADMIN'
+  | 'OPDIV_READONLY_ADMIN'
+  | 'ISSO'
+  | 'ISSM'
+  | 'ADMIN'
+  | 'READONLY_ADMIN'
+
+export type OpDiv = {
+  opdiv_id: number
+  code: string
+  name: string
+  is_parent: boolean
+  active: boolean
+}
+
 export type userData = {
   userid: string
   email: string
   fullname: string
-  role: string
+  role: UserRole
   assignedfismasystems?: number[]
+  identity_provider?: 'okta' | 'entra'
+  assignedopdivids?: number[] | null
 }
 export type RequestOptions = {
   method: string
@@ -55,6 +76,7 @@ export type FismaSystemType = {
   reactivated_by: string | null
   reactivated_date: string | null
   reactivation_notes: string | null
+  opdiv_id?: number | null
 }
 export type FismaSystems = {
   fismaSystems: FismaSystemType[]
@@ -155,7 +177,7 @@ export type users = {
   assignedfismasystems: number[]
   email: string
   fullname: string
-  role: string
+  role: UserRole
   userid: string
   deleted?: boolean
   isNew?: boolean
--- a/src/utils/opdivs.ts
+++ b/src/utils/opdivs.ts
@@ -1,0 +1,20 @@
+import axiosInstance from '@/axiosConfig'
+import type { OpDiv } from '@/types'
+
+/**
+ * Fetches the OpDiv reference list from GET /api/v1/opdivs.
+ *
+ * The endpoint is open to any authenticated user. By default the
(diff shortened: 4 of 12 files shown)


jsos3-cms (dismissed):
LGTM!

I think we could convert the old schema into a regression test, if that's okay with you? If there's anything we'd want to preserve against and catch, like the old params being passed, it'd be nice to include some feedback to the user beyond just the error mode. Thoughts on this low priority capture?

danielbowne:
@jsos3-cms good suggestion pushed `f61fe9c` adding `src/utils/userRoles.test.ts`. Table-driven coverage across every value in the `UserRole` union for each predicate, plus null/undefined/empty-placeholder rejection and unknown-string rejection. The legacy `ADMIN` and `READONLY_ADMIN` rows are flagged with `// legacy, removed in Stage D` so the same PR that drops the values from the helpers also removes the corresponding test assertions failing tests become the signal that an intentional decision is needed rather than a silent regression.

Mind taking another quick look when you have a minute?

MackOverflow (dismissed):
The deploy-order rationale and the before/after table make the tradeoff immediately obvious for anyone picking this up later. The overall design is awesome: centralizing all role logic in userRoles.ts means Stage D (dropping legacy roles) is a one-file change with failing tests as the signal, rather than a grep-and-hope across the whole codebase.

Quick thought on on src/constants.ts / authLoader.ts / Title.tsx — on role: '' as UserRole:
This cast showing up in a few more places now — totally understand why (the existing userData shape needs a valid role to initialize), but it's worth flagging as a minor code smell we're accumulating. Something to consider for a future cleanup: making role optional or using a dedicated sentinel value (like an 'UNAUTHENTICATED' entry in the union) would let us drop all the casts. Not blocking at all, just noting it before we have 6 of these.

Overall looks great!

PR #358, opened 2026-05-19
fix(questionnaire): guard empty response and direct-URL navigation

danielbowne:
## Summary

Prod bug fix for `/questionnaire/<acronym>` on systems whose `/questions` endpoint returns no rows. Pre-existing, unrelated to the multi-OpDiv work.

## Root cause

The backend join in `FindQuestionsByFismaSystem` (Go, `internal/model/questions.go`) ties `questions` to `functions` to `fismasystems` on `datacenterenvironment`. When no `functions` row matches a system's `datacenterenvironment`, the join returns an empty result set. The Go `[]*Question` zero value is a nil slice, which `encoding/json` serializes as `null` — so the production response shape is `{ data: null }`, not `{ data: [] }`.

`QuestionnairePage.tsx` dereferenced `categoriesData[0].name` inside a `.then` block on the empty result, throwing a `TypeError` that escaped the `.catch` handler (catch only fires on rejected promises, not synchronous throws inside `.then`). The error bubbled to React Router, surfaced as the "No HydrateFallback element provided" console warning, and the user saw a generic "An error occurred" snackbar.

A second, independent crash was uncovered during review: any direct or bookmarked navigation to `/questionnaire/<acronym>` (any system) threw `TypeError: Cannot read properties of null (reading 'fismasystemid')` because `location.state` is `null` for those entries. Both crashes are fixed here.

## Data model context

The two relevant FismaSystem columns are independent in the schema:

- `decommissioned: boolean` — flag set by the modern decommission API (`DeleteFismaSystem`)
- `datacenterenvironment: string` — environment identifier used by the questions join

The decommission API never modifies `datacenterenvironment`. The `'DECOMMISSIONED'` environment value comes from two narrow paths: the migration 0016 backfill (legacy systems retired via the prior workaround of editing the environment field directly), and any post-hoc manual edit of a system's environment to that value.

Practical impact: modern-decommissioned systems retain their real datacenter environment, so the `/questions` join still returns rows. Those systems can and should continue rendering questionnaire history for audit and compliance review. The fix below intentionally does not preemptively block any row.

### Empty-questions guard in `QuestionnairePage`

- New `noQuestions` state. After reading `response.data?.data`, an `!data || (Array.isArray(data) && data.length === 0)` check treats `null`, `undefined`, or empty array as the no-questions signal and short-circuits before the offending `categoriesData[0].name` access. Comment explains the Go nil-slice serialization rationale.
- A local `questionsEmpty` flag also skips the subsequent `/scores` call so empty-questionnaire systems do not fire an unnecessary request.
- `setNoQuestions(false)` runs at the top of the effect before fetch so a prior empty-questionnaire view does not bleed into the next render when `system` changes.
- Render guard above the main return: when `noQuestions` is true, renders `<BreadCrumbs />` plus an `<Alert severity="info">` stating "No questionnaire is available for this system. This typically applies to systems whose data center environment is no longer in scope for the current data call." The message is deliberately framed around the actual signal (environment not in scope) rather than claiming the system is decommissioned, because the two conditions are not equivalent.

### Direct-URL navigation guard

- `const system = location.state?.fismasystemid as number | undefined` — optional-chained, no more null dereference.
- Separate render guard above the empty-questions branch: when `system` is `undefined`, renders an `<Alert severity="warning">` with "Cannot load questionnaire from a direct link. Please open it from the system list."

### `FismaTable` unchanged

An earlier revision of this PR preemptively disabled the questionnaire row action for any row with `decommissioned === true`. That was based on a flawed assumption that decommissioned systems always carry `datacenterenvironment = 'DECOMMISSIONED'`. Backend inspection refutes that — see the "Data model context" section. Modern decommissioned systems retain real environments and have legitimate questionnaire history, so the preemptive disable blocked valid access. The disable has been removed; the page-level empty-state guard handles the empty cases correctly without preempting legitimate paths.

## Out of scope

- Backend changes. The `null`/empty response shape is the documented Go behavior; the frontend's job is to handle it.
- Restoring an active questionnaire for systems whose environment is genuinely out of scope. There are no questions to answer in those cases; surfacing the empty state is the fix.
- Keyboard a11y on row actions — no longer applicable since the preemptive disable was removed.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn lint` clean
- [x] Pre-push hooks clean (typecheck + lint)
- [x] Two independent code reviews — all critical findings addressed
- [ ] Dev smoke: open the questionnaire for an active system via the FismaTable row icon — questions render normally, behavior unchanged.
- [ ] Dev smoke: open the questionnaire for a modern-decommissioned system (one with a real environment, e.g. CMS-Cloud-AWS) — questionnaire still renders against the system's historical data call. Confirms the audit/compliance read path is preserved.
- [ ] Dev smoke: open the questionnaire for a legacy or environment-marked-DECOMMISSIONED system — page renders the friendly Alert, no console error, no snackbar.
- [ ] Dev smoke: paste any questionnaire URL into a fresh tab (no `location.state`) — page renders the "open from system list" Alert, no console error.

### DIFF
--- a/src/views/FismaTable/FismaTable.tsx
+++ b/src/views/FismaTable/FismaTable.tsx
@@ -418,7 +418,13 @@ export default function FismaTable({ scores }: FismaTableProps) {
       disableColumnMenu: true,
       renderCell: (params: GridRenderCellParams) => (
         <>
-          <Tooltip title="Questionnaire">
+          <Tooltip
+            title={
+              params.row.decommissioned
+                ? 'Decommissioned systems have no active questionnaire'
+                : 'Questionnaire'
+            }
+          >
             <span>
               <GridActionsCellItem
                 icon={<QuestionAnswerOutlinedIcon />}
@@ -426,8 +432,12 @@ export default function FismaTable({ scores }: FismaTableProps) {
                 label={`View Questionnaire for ${params.row.fismaname}`}
                 className="textPrimary"
                 role="button"
+                disabled={params.row.decommissioned === true}
                 onClick={(event) => {
                   event.stopPropagation()
+                  if (params.row.decommissioned) {
+                    return
+                  }
                   navigate(
                     `/${RouteNames.QUESTIONNAIRE}/${params.row.fismaacronym.toLowerCase()}`,
                     {
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -104,6 +104,7 @@ export default function QuestionnarePage() {
   const [datacall, setDatacall] = React.useState<string>('')
   // const { latestDatacall, latestDatacallId } = useContextProp()
   const [loadingQuestion, setLoadingQuestion] = React.useState<boolean>(true)
+  const [noQuestions, setNoQuestions] = React.useState<boolean>(false)
   const [categories, setCategories] = React.useState<Category[]>([])
   const [stepFunctionId, setStepFunctionId] = React.useState<number[]>([])
   const [functionIdIdx, setFunctionIdIdx] = React.useState<{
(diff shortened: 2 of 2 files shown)


MackOverflow (approved):
Fix it danny so I can stop picking on EPPEC in my demos!

Issue #318, opened 2026-01-30
[Question] Add ability to reactivate/un-decommission systems?


danielbowne:
## Problem

Currently, once a FISMA system is decommissioned, there is **no way to undo it** through the UI or API. The decommission action is intentionally one-way to maintain audit trail integrity, but this creates a problem if a system is decommissioned by accident.

**Current Safeguards:**
- Multiple warnings in confirmation dialog
- Admin-only permission required
- Clear messaging about what decommissioning means

**But Still No Recovery:**
- If admin clicks through warnings by mistake, system is permanently decommissioned
- Only way to fix: Manual database UPDATE query
- No audit trail of who/when it was reactivated

## Current Backend Architecture

The `SaveFismaSystem` (PUT) endpoint **intentionally excludes** decommission fields from updates:

```go
// backend/internal/model/fismasystems.go:93-106
// UPDATE - exclude decommissioned fields
sqlb = stmntBuilder.Update("fismasystems").
    Set("fismauid", f.FismaUID).
    Set("fismaacronym", f.FismaAcronym).
    // ... only updates first 12 columns
    // Does NOT touch decommissioned, decommissioned_date, etc.
```

**Why:** Decommission state is managed separately via DELETE endpoint to maintain clear audit trail.

## Manual Workaround (Database Access Required)

If system accidentally decommissioned, admin must:
1. Connect to database directly (bastion host)
2. Run manual UPDATE query:
   ```sql
   UPDATE fismasystems 
   SET decommissioned = false,
       decommissioned_date = NULL,
       decommissioned_by = NULL,
       decommissioned_notes = NULL
   WHERE fismasystemid = 123;
   ```
3. No audit trail of who fixed it or when

## Questions for Product Owner / Customer

**Before implementing, need to confirm:**

1. **Has this actually happened?** Have we had accidental decommissions that needed reversal?
2. **How often?** Is this a real pain point or theoretical concern?
3. **Manual fix acceptable?** Can we rely on database access for rare accidents, or need UI feature?
4. **Audit requirements?** If we add reactivation:
   - Should we keep original decommission data for audit trail?
   - Should we track who/when reactivated the system?
   - Should we require a reason for reactivation (like we do for decommission)?

### Option A: Simple Reactivation (Clear All Data)
**Behavior:** Reactivation completely erases decommission history

**Database:**
```sql
UPDATE fismasystems 
SET decommissioned = false,
    decommissioned_date = NULL,
[...]

PR #361, opened 2026-05-20
Align frontend scoring with HHS scale (1.0-5.0) and read tier from API

danielbowne:
## Summary

Frontend pair for `CMS-Enterprise/ztmf#318` (branch `feature/hhs-scoring-alignment`). Tracked under `CMS-Enterprise/ztmf-misc#175`.

Both PRs auto-deploy to dev for joint testing, then merge to main within a short window of each other to keep the prod cutover gap small.

## What the backend now sends

`/api/v1/scores/aggregate` and `/api/v1/scores/aggregate?include_pillars=true` keep their JSON shape, with two additions and one semantic change:

- `data[].systemscore` is on the HHS 1.0 to 5.0 scale (was 1.0 to 4.0)
- `data[].systemtier` is a new string field
- `data[].pillarscores[].score` is on the same HHS 1.0 to 5.0 scale
- `data[].pillarscores[].tier` is a new string field on each pillar entry

Tier values: `Optimal`, `Advanced`, `Initial`, `Traditional`, `Not Assessed`. Boundaries: Optimal at 4.1+, Advanced at 3.1+, Initial at 2.1+, Traditional at 1.01+, Not Assessed below 1.01. The string is authoritative; the frontend no longer derives tier from the numeric score.

### Centralized types and shared style map

- `src/types.ts`: new `ScoreTier` union, `PillarScore`, `ScoreAggregate`, and `SystemScoreEntry` types. `FismaTableProps.scores` updated from `Record<number, number>` to `Record<number, SystemScoreEntry>`. The two tier fields (`systemtier` and `pillarscores[].tier`) are optional so a transient mismatch between deployed backend and frontend during the cutover does not throw a runtime TypeError; tier chips simply hide until the backend ships tier strings.
- `src/utils/tierStyles.ts` (new): single `TIER_STYLES` map of `{ color, backgroundColor }` keyed by tier. Color tokens lifted from the prior `getMaturityLevel` ladder so chips render identically once data flips. Exports a `styleForTier` helper for safe undefined handling.
- `src/utils/tierStyles.test.ts` (new): regression contract pinning all five tier keys, validating hex tokens, and covering `styleForTier`. Failing tests on this file are the signal that the tier enum drifted.

### Pillar Scores modal

- Deletes the `getMaturityLevel(score: number)` ladder (lines 69-99 in `src/components/PillarScoresModal/PillarScoresModal.tsx`). It was tuned for 1.0-4.0 and is exactly wrong for the new scale.
- Replaces the six former call sites with reads of `latestScore.systemtier` and `pillar.tier` from the API response, looked up through `styleForTier(...)` against the shared style map. Sites that previously rendered a tier label hide the chip entirely when the tier is absent.
- Updates the score chip footer from `/ 4` to `/ 5` in both the overall score and per-pillar blocks.
- Updates the RadarChart `PolarRadiusAxis` domain from `[0, 4]` to `[0, 5]` and bumps `tickCount` from `5` to `6` so axis labels land on whole integers (0, 1, 2, 3, 4, 5) instead of awkward fractions.
- Replaces the local `PillarScore` and `SystemScore` interfaces with imports of the shared `ScoreAggregate` and `PillarScore` types.

### FISMA system list (score column)

- Deletes the inline `if/else if` color chain at lines 376-384 of `src/views/FismaTable/FismaTable.tsx`. Replaces with a `styleForTier(...)` lookup keyed off `params.row.systemtier` passed through from `/scores/aggregate`. Cells without a tier render with no background fill so an old backend reads as "unknown" rather than a misleading color.
- Drops the local `SystemScore` interface from the pillar scores cache typing in favor of the shared `ScoreAggregate` type.
- Updates the `FooterPropsOverrides` declaration to match the new prop shape.

### Score legend

- Rebuilds `src/views/ScoreTable/ScoreTable.tsx` from a hardcoded 4-column CMS legend to a 5-column HHS legend driven by a `TIER_LEGEND` array. Header cell colors are now drawn from `TIER_STYLES` so the legend itself matches the colors used by data cells. Ranges: Not Assessed `< 1.01`, Traditional `1.01 - 2.0`, Initial `2.1 - 3.0`, Advanced `3.1 - 4.0`, Optimal `4.1 - 5.0`.

### Score map plumbing

- `src/views/Home/Home.tsx` now builds the per-system score map as `Record<number, SystemScoreEntry>` carrying both the numeric score and the tier string.
- `src/views/StatisticBlocks/StatisticsBlocks.tsx` adapts to the new map shape (reads `.score` from each entry). No tier surfacing on the KPI tiles in this PR.

## Color note

FismaTable cells today use brighter colors (`#DAA9EC`, `#FFD5A5`, `#F2FBC4`, `#93F0ED`) than the modal's softer palette. After this PR both surfaces share `TIER_STYLES`, so table cells adopt the modal's softer colors. Minor visual change to flag during dev smoke.

## Out of scope

- New API calls. The same `/scores/aggregate` endpoints serve the new data.
- Re-derivation of tier from numeric score. The frontend reads the authoritative string only.
- Broader scoring dashboard reflow. Structural changes are minimal.

## Verification

- [x] `yarn typecheck` clean
- [x] `yarn lint` clean
- [x] `yarn test`: 81 passed (was 73; added 8 in `tierStyles.test.ts`) / 3 skipped / 0 failed
- [x] Pre-push hooks clean
- [x] Independent code review
- [x] Dev smoke (after backend pair `CMS-Enterprise/ztmf#318` lands on dev):
  - Pillar Scores modal opens for a system with scores. Tier label and background color render. Numeric score reads in the 1.0-5.0 range. Score chip shows `/ 5`.
  - RadarChart axis ticks read `0, 1, 2, 3, 4, 5`.
  - FismaTable score column shows tier-colored backgrounds drawn from `TIER_STYLES`. No-score row shows the gray `Not Assessed` styling.
  - ScoreTable legend shows five tiers in order with the new ranges and matching colors.
  - All-Optimal system reads `5.00 / Optimal`. All-Advanced system reads `4.00 / Advanced` at the boundary. Mixed-answer system falls in the new range.

## Comms angle

Real-data validation in the backend plan showed zero tier downgrades and roughly 60-64% of systems moving up exactly one tier under the new math. Operators should expect scores to appear higher after the cutover; this is the HHS recalibration, not a performance change.

## References

- Backend pair: `CMS-Enterprise/ztmf#318` (`feature/hhs-scoring-alignment`)
- Cross-program tracker: `CMS-Enterprise/ztmf-misc#175`


MackOverflow (dismissed):
Looks good to me, I flagged three things on review, none blockers:

1) **One potential type-safety gap:** In PillarScoresModal.tsx at the tier chip render (TIER_STYLES[latestScore.systemtier].color), systemtier is ScoreTier | undefined but the access is guarded only by latestScore.systemtier &&. That works at runtime, but TypeScript may require a non-null assertion or narrowing depending on strict settings — worth confirming typecheck is truly clean on this line.

2) TIER_STYLES is documented as a legacy alias but is still the primary export used in PillarScoresModal. The comment says "new code should import TIER_CHIP_STYLES directly," but the modal itself still uses TIER_STYLES. Minor inconsistency — not a bug, but worth a follow-up cleanup. Assuming this will be cleaned up as more PR's hit?

>> interesting UX note: 3) The ScoreTable legend: Not Assessed renders with a transparent background, which means on a white table it looks like a blank unstyled cell while the other four have colors. Depending on how it looks in the browser, it might appear unintentional. Worth a quick visual check.

danielbowne:
Thanks for the careful look @MackOverflow.

Quick rundown on the three:

1. Type-safety on that line: typecheck stays clean. The `tier && ...` guard narrows away the undefined before the property access, so no assertion needed.

2. Good catch on the alias.  I saw this also and made filed #362 this morning to collapse the chip and cell maps into one `TIERS` constant and drop the legacy export. The modal moves to the consolidated shape there. Kept it out of this PR so the scoring math and 508 color changes stay reviewable on their own.

3. Fixed in `951a3ea`. The `Not Assessed` legend cell now uses the soft chip gray so it reads as a real tier instead of a blank cell. Actual table rows still render `Not Assessed` as transparent, since a score of zero should look like no signal rather than a styled state.

Re-review when you get a minute?

MackOverflow (approved):
lets goooo!

PR #365, opened 2026-05-22
Add multi-env Makefile pattern for impl/main worktrees

danielbowne:
## Summary
- Auto-detect `ZTMF_ENV` from the current git branch (`impl` / `feature/impl-*` → impl, else main) and `-include Makefile.$(ZTMF_ENV)` for per-env overrides.
- `Makefile.main` keeps the default frontend port (5174); `Makefile.impl` offsets to 5175 so the main worktree dev server can run alongside.
- `vite.config.ts` now reads `server.port` from `VITE_DEV_PORT` (default 5174), exported by the Makefile.
- Mirrors the backend repo's Makefile.main / Makefile.impl pattern so both worktrees behave consistently.

Override on demand: `make dev ZTMF_ENV=main`.

## Test plan
- [ ] In this worktree (branch impl/feature/impl-*): \`make env\` reports `ZTMF_ENV=impl`, `FRONTEND_PORT=5175`.
- [ ] `make env ZTMF_ENV=main` reports `FRONTEND_PORT=5174`.
- [ ] `make dev` from impl worktree binds 5175; main worktree `make dev` binds 5174; both run simultaneously.
- [ ] `npx tsc --noEmit` clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,23 @@
 # ZTMF UI Development Makefile
 
-.PHONY: help install lint lint-fix format format-check build build-dev test dev clean
+# Auto-detect env from current git branch.
+# impl / feature/impl-* -> impl; everything else -> main.
+# Override on command line: make dev ZTMF_ENV=impl
+ZTMF_ENV ?= $(shell git branch --show-current 2>/dev/null | grep -qE '^(impl|feature/impl-)' && echo impl || echo main)
+-include Makefile.$(ZTMF_ENV)
+
+# Defaults (overridden by Makefile.impl when ZTMF_ENV=impl)
+FRONTEND_PORT ?= 5174
+export VITE_DEV_PORT = $(FRONTEND_PORT)
+
+.PHONY: help install lint lint-fix format format-check build build-dev test dev clean env
 
 # Default target
 help:
+	@echo "Active env: $(ZTMF_ENV) (frontend port $(FRONTEND_PORT))"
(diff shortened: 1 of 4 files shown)


MackOverflow (approved):
I know this is impl, but one flag: 

ui / Deploy is skipping, not passing. Since vite.config.ts changed, it's worth confirming whether that job has a path filter that's excluding Makefile and config changes, or if it's intentionally skipped for non-src changes. If the port config in vite.config.ts were broken, this PR wouldn't catch it at the CI level — it would only surface at runtime.

danielbowne:
> ui / Deploy is skipping, not passing. Since vite.config.ts changed, it's worth confirming whether that job has a path filter that's excluding Makefile and config changes, or if it's intentionally skipped for non-src changes.

This change only touches the local Vite dev server. Normal CI flow is deploy to dev on PR and prod on merge; for impl, the orchestration deploys on push. In all three cases the build step is `yarn build:<env>` (`tsc && vite build`), which never reads `VITE_DEV_PORT` — that variable only feeds `server.port` for `yarn dev`. Deployed bundles can't regress on this change.

The purpose is purely a local-dev runbook: `make dev` reads the current branch and binds 5175 from the impl worktree / 5174 from main, so both dev servers can run side-by-side without port collision. Reviewer test is `git checkout feature/impl-multi-env-makefile && make dev` and confirm it binds 5175.

PR #367, opened 2026-05-27
Read ZTMF Insights from the new system_enrichment endpoint, gate on sdl_sync_enabled

danielbowne:
## Summary

Points the system detail page's enrichment card at the new generic `system_enrichment` endpoint, and gates the section so it only appears for systems opted into ZTMF Insights. This is the frontend half of decoupling ZTMF Insights from the core scoring app.

## Changes

- The enrichment card now reads from `GET /api/v1/systemenrichment/{fisma_uuid}` instead of the old `cfactssystems` endpoint. The new response nests the enrichment fields under `payload`, with `fisma_uuid` and `synced_at` as top-level siblings, so the fetch flattens them back into the shape the card already renders. The display, loading, and 404/403 empty states are unchanged.
- The whole section is now gated on `system.sdl_sync_enabled`. ZTMF Insights is CMS-only for now, and the toggle is off by default, so future HHS and OpDiv systems will not see the section until they opt in.
- Relabeled the user-facing "CFACTS" text to "ZTMF Insights" (section header and empty/error states).

## Testing

- `yarn typecheck` and `yarn lint` pass clean.
- Verified against dev, where the endpoint is live with real enrichment data (the backend shipped in CMS-Enterprise/ztmf#323, deployed to dev and prod).

Refs CMS-Enterprise/ztmf-misc#212

### DIFF
--- a/src/views/SystemDetailPage/CfactsRecordCard.tsx
+++ b/src/views/SystemDetailPage/CfactsRecordCard.tsx
@@ -95,10 +95,23 @@ export default function CfactsRecordCard({ fismaUid }: CfactsRecordCardProps) {
     setHasError(false)
 
     axiosInstance
-      .get(`cfactssystems/${fismaUid}`)
+      .get(`systemenrichment/${fismaUid}`)
       .then((res) => {
         if (!cancelled) {
-          setCfacts(res.data?.data ?? res.data)
+          // The endpoint returns { data: { fisma_uuid, payload, synced_at } }.
+          // The enrichment fields live in payload; fisma_uuid and synced_at are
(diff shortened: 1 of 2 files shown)


MackOverflow (approved):
Approved! Nit of the day: 3. CfactsRecordCard.tsx — component/file name

The file and component are still named CfactsRecordCard even though it's now rendering ZTMF Insights data. Totally fine to defer the rename — just flagging so it doesn't get lost. Might be worth a quick follow-up issue to rename to SystemEnrichmentCard or ZtmfInsightsCard once the old CFACTS surface is retired.

PR #368, opened 2026-05-29
fix(questionaire): implement fixed pillar sorting and add related tests

### LINKED ISSUE #352: Set order of Pillars in Questionnaire
### Description

The order of the Pillars in the questionnaire is not always the same.

### Steps to Reproduce

1. Go to a Questionnaire for a System
2. Scroll through the list on the left and note the order of the Pillars
3. Visit a different questionnaire, note the differences

### Expected Behavior

It should be in this order:

Identity
Devices
Networks
Apps and workloads
Data
Cross Cutting

### Actual Behavior

They appear in every sort of order.

### Environment

all

### Acceptance Criteria

### Severity

### Affected Component

_No response_

### Screenshots

Compare the order of CGDP DPP Portal
https://dev.ztmf.cms.gov/#/questionnaire/cgdp%20dpp%20portal/FY2026_Q1/identity/access-management

to ACO MS
https://dev.ztmf.cms.gov/#/questionnaire/aco-ms/FY2026_Q1/identity/access-management

to ESS
https://dev.ztmf.cms.gov/#/questionnaire/ess/FY2026_Q1/applications/application-security-testing

Image

Image

### Logs / Error Messages

```shell

```

### Workaround

_No response_

tarratsco:
## Summary

Fixes the questionnaire pillar list showing up in different orders on different systems. Pillars now always appear in a single fixed order: Identity, Devices, Networks, Applications, Data, and Cross Cutting.

Fixes #352

## Root cause

Both questionnaire views sorted pillars by the numeric `order` field that the API sends on each question (`question.pillar.order`). When that field is missing or the same value for every pillar, the comparator returns `0` (or `NaN`), so the sort does nothing and the pillars fall back to whatever order the API response happened to use. That response order varies from system to system, which is why the list looked random.

## Fix

Stop trusting the API `order` for pillars and use a fixed, hardcoded order instead. This matches the pattern already used to order functions within a pillar (`PILLAR_FUNCTION_MAP`).

- Added a `PILLAR_ORDER` constant in `src/constants.ts` that lists pillars in the required display order.
- Added a `sortPillars` helper in `src/utils/sortPillars.ts` that orders pillar names by `PILLAR_ORDER`. Unknown pillar names (for example a new one that has not been added to the list yet) sort to the end instead of jumping to the front. The helper returns a new array and does not mutate its input.
- Updated `QuestionnairePage.tsx` and `QuestionnaireModal.tsx` to call `sortPillars` and removed the old `pillarOrder` map logic from both. Function ordering inside each pillar is unchanged.

## Testing

- New unit tests in `src/utils/sortPillars.test.ts` cover the canonical order, unknown pillars landing at the end, and the no-mutation guarantee.
- `yarn test`, `yarn tsc --noEmit`, and `yarn lint` all pass.
- Verified in the app that the pillar list reads Identity, Devices, Networks, Apps and workloads, Data, Cross Cutting on multiple systems, in both the questionnaire page and the modal view.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -39,6 +39,19 @@ export const CONFIRMATION_MESSAGE =
 export const CONFIRMATION_MESSAGE_QUESTION =
   'Your changes will not be saved! Are you sure you want to leave question without saving your changes?'
 
+/**
+ * The fixed order for pillars on the questionnaire, no matter what order the
+ * API returns them in. Each name has to match the API's
+ * `question.pillar.pillar` value exactly (same keys as {@link PILLAR_FUNCTION_MAP}).
+ */
+export const PILLAR_ORDER: string[] = [
+  'Identity',
+  'Devices',
+  'Networks',
+  'Applications',
+  'Data',
+  'CrossCutting',
+]
 export const PILLAR_FUNCTION_MAP: { [key: string]: string[] } = {
   Identity: [
     'AccessManagement',
--- a/src/utils/sortPillars.test.ts
+++ b/src/utils/sortPillars.test.ts
@@ -1,0 +1,36 @@
+import { sortPillars } from '@/utils/sortPillars'
+
+test('orders pillars into the canonical sequence regardless of input order', () => {
+  const input = [
+    'Data',
+    'Identity',
+    'CrossCutting',
+    'Networks',
+    'Devices',
+    'Applications',
+  ]
+
+  expect(sortPillars(input)).toEqual([
+    'Identity',
+    'Devices',
+    'Networks',
+    'Applications',
+    'Data',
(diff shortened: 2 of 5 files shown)


danielbowne (approved):
LGTM. Verified locally: typecheck, lint, production build, and the full test suite all pass, both on the branch as-is and after merging current main. The new sortPillars unit tests cover the canonical order, unknown pillars sorting to the end, and the no-mutation guarantee. Root cause and fix are sound — replacing the unreliable API order field with a fixed PILLAR_ORDER matches the existing PILLAR_FUNCTION_MAP pattern. Shipping.

tarratsco:

PR #369, opened 2026-05-29
fix function sort and move to utility

### LINKED ISSUE #353: Set order of the Functions in Questionnaire
### Description

The order of the Functions within the Pillars is arbitrary, and they should be the same each time.  The order should be the same as in the CISA ZT Maturity Model

The functions have an order field to help with the special ordering.

### Steps to Reproduce

1. Go to a Questionnaire
2. See that the order is wrong

### Expected Behavior

The order of the Functions within a Pillar should be the same for each system each time they view their questionnaire

### Actual Behavior

They are in all sorts of orders.

### Environment

all

### Acceptance Criteria

### Severity

### Affected Component

_No response_

### Screenshots

Image

### Logs / Error Messages

```shell

```

### Workaround

_No response_

costacalvin:
## Summary
Fixes questionnaire functions within each pillar appearing in different orders on different systems. Functions now always appear in the canonical CISA Zero Trust Maturity Model order for their pillar.

Fixes #353

## Root cause
Both questionnaire views sorted functions using an inline sortSteps helper that called Array.indexOf() against PILLAR_FUNCTION_MAP. Two bugs:

- When a function name is not found in the map, indexOf returns -1. Since -1 - positiveIndex is negative, the unknown function sorts to the front of the list instead of the end, pushing all other functions out of order.
- Array.prototype.sort() was called directly on the steps array, mutating it in place.

PILLAR_FUNCTION_MAP already existed in src/constants.ts with the correct canonical order.

## Fix
Extracted the sort logic into a dedicated utility that handles unknown function names safely following the pattern set by Onix in this PR https://github.com/CMS-Enterprise/ztmf-ui/pull/368.

- Added a sortFunctions helper in src/utils/sortFunctions.ts that ranks each function by its position in PILLAR_FUNCTION_MAP. Unknown function names (for example a new one not yet added to the map) sort to the end instead of jumping to the front. The helper returns a new array and does not mutate its input.
- Updated QuestionnairePage.tsx and QuestionnareModal.tsx to call sortFunctions and removed the inline sortSteps definition from both. Pillar ordering is unchanged.

## Testing
New unit tests in src/utils/sortFunctions.test.ts cover the canonical order for all six pillars, unknown function names landing at the end, and the no-mutation guarantee.

`npm test` and `npm run lint` all pass.

Test data only had one question per pillar, modified seed data and was able to view multiple questions under each pillar. Unknown function names get added to the end of the sort.


danielbowne:
Good cleanup. Moving the sort into a tested utility, with no mutation and a guard for unknown names, is the right structure. I also checked the backend: `function.order` comes back as `0` for every function, so the hardcoded map is the right source of truth here, like in #368.

The reason the order looks scrambled isn't the sort logic, it's a name mismatch between `PILLAR_FUNCTION_MAP` and the function names the API actually returns. Three functions in the map don't match the data, so they never find an index and end up out of place:

| Pillar | In the map | Actual data name |
|---|---|---|
| Devices | `DeviceThreatProtection` | `Device-ThreatProtection` |
| Applications | `ApplicationSecurityTesting` | `Application-SecurityTesting` |
| Applications | `AppThreatProtection` | `Application-ThreatProtection` |

You can see it in prod right now. Devices leads with "Device - Threat Protection", and Applications leads with "Application - Security Testing" and "Application - Threat Protection". Those are the three unmapped names, pushed to the front because the old `indexOf` returns -1 for them. Identity, Networks, and Data all map cleanly and are already consistent.

This PR moves those three from the front to the end, which is more stable, but they stay unmapped, so they still won't land in the right spot. To fully close #353 we should also fix those three keys in `PILLAR_FUNCTION_MAP` so they match the data names. In `src/constants.ts`:

```diff
   Devices: [
     'AssetRiskManagement',
     'Device-AutomationOrchestration',
     'Device-Governance',
-    'DeviceThreatProtection',
+    'Device-ThreatProtection',
     'PolicyEnforcement',
     'Device-VisibilityAnalytics',
     'ResourceAccess',
   ],
   Applications: [
     'AccessibleApplications',
     'AccessAuthorization-Users',
     'Application-AutomationOrchestration',
     'Application-Governance',
     'SecureDevDeployWorkflow',
-    'ApplicationSecurityTesting',
-    'AppThreatProtection',
+    'Application-SecurityTesting',
+    'Application-ThreatProtection',
     'Application-VisibilityAnalytics',
   ],
```

Once the names match, every function maps and the order is fully deterministic. While we're in there it's worth confirming the intended within-pillar order against the CISA ZTMM model, since the issue description asks for that.

costacalvin:
Updated function names to match data names + updated the PILLAR_FUNCTION_MAP order to match the CISA ZTMM 2.0 order. Ready for review @danielbowne

danielbowne (approved):
Verified locally against the dev backend: every function now maps (no unmapped names), the within-pillar order matches the CISA ZTMM v2.0 model for all six pillars, and live data sorts cleanly. tsc, lint, the full test suite, and the prod build all pass. The three name corrections (Device-ThreatProtection, Application-SecurityTesting, Application-ThreatProtection) resolve the front-floating functions seen in prod. Closes #353.

PR #370, opened 2026-06-01
chore(deps): drop unused uuid and bump axios (security)

danielbowne:
## Summary

Two dependency security fixes:

1. **Remove `uuid`** (and `@types/uuid`). It was a direct dependency but is never imported anywhere in the app or test code, so the cleanest way to clear the alert is to drop it rather than upgrade it. This supersedes Dependabot PR #364 (bump `uuid` 9.0.1 → 14.0.0) — please close #364 in favor of this change.
2. **Bump `axios` 1.15.2 → 1.16.1** to resolve two high-severity advisories.

## uuid — why remove instead of bump

- **`uuid` is unused.** A repo-wide search finds no `import`/`require` of the package. The only reference was a leftover Jest `moduleNameMapper` entry (carried over from the original repo import) whose job was to force uuid's CommonJS entry point — and that mapper only does anything if a test actually imports uuid, which none do. That line is removed here too.
- **It clears the CVE.** Our pinned `9.0.1` is affected by CVE-2026-41907 / GHSA-w5hq-g745-h8pq (out-of-bounds write in `v3()`/`v5()`/`v6()` when a caller-supplied buffer is passed). Removing the direct dependency takes it out of our tree entirely.
- **The Dependabot bump over-reaches.** The fix actually shipped in uuid `11.1.1`; `14.0.0` is just the latest tag. Upgrading would keep an unused package *and* pull in uuid v12's removal of CommonJS support — the exact thing the stale Jest mapper was working around — for no benefit.

### Note on the remaining transitive copy

`@storybook/addon-actions` still pulls in `uuid@9` transitively. That is a **devDependency** and never ships in the production bundle. If the scanner flags it, the minimal follow-up is a `resolutions: { "uuid": "^11.1.1" }` override (11.1.1 keeps CommonJS, so Storybook and Jest are unaffected). Left out here to keep the change scoped to our own direct dependency.

## axios — 1.15.2 → 1.16.1

`axios` is used by the app's API client. Bumped to clear two high-severity advisories present in 1.15.2:

- **SNYK-JS-AXIOS-17111062** — server-side request forgery (SSRF)
- **SNYK-JS-AXIOS-17111079** — prototype pollution

The fixes landed in 1.16.0; the resolved version is 1.16.1. This is a minor bump within our existing `^1` range, no API changes.

## Verification

- `tsc --noEmit` — passes
- `yarn lint` — passes
- `yarn test` — 118 passed, 3 skipped
- `vite build` — production build succeeds

### DIFF
--- a/jest.config.cjs
+++ b/jest.config.cjs
@@ -28,10 +28,6 @@ module.exports = {
     '^@/(.*)$': '<rootDir>/src/$1',
     '^react-native$': 'react-native-web',
     '^.+\\.module\\.(css|sass|scss)$': 'identity-obj-proxy',
-    // Force module uuid to resolve with the CJS entry point,
-    // because Jest does not support package.json.exports.
-    // See https://github.com/uuidjs/uuid/issues/451
-    uuid: require.resolve('uuid'),
   },
   moduleFileExtensions: [
     // Place tsx and ts to beginning as suggestion from Jest team
--- a/package.json
+++ b/package.json
@@ -81,7 +81,6 @@
     "react-virtuoso": "^4.10.1",
     "recharts": "^3.1.2",
     "uswds": "^2.14.0",
-    "uuid": "^9.0.1",
     "yup": "^1.3.3"
   },
   "devDependencies": {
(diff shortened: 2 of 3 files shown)


MackOverflow (approved):
Sorry about the delay Danny! This looks great, not a nit to pick!

PR #376, opened 2026-06-08
fixed bug where downloading all active systems included decommissioned

### LINKED ISSUE #375: Export button adds more systems than selected
### Description

When a user selects all active systems in the UI and triggers the Excel download, the export ignores the active/decommissioned filter and includes all systems regardless of status — or the inverse, where selecting decommissioned systems exports active ones instead. The filter state applied in the table view is not being correctly propagated to the download request, so the backend generates the spreadsheet against the unfiltered system set rather than the user's current selection.

### Steps to Reproduce

1. Go to the ZTMF Homepage
2. Click on the toggle box to the left of System Name. This selects all active and available systems for download.
3. Scroll down to the bottom of the table and click on the Download selected system answers button then open the downloaded excel file.
4. The excel file will have all decommissioned system answers, which they did not select, included in with the complete list of active systems they did select.
5. This can be done with the order reversed as well, selecting all decommissioned systems answers to download then having all active systems included in the excel sheet.

### Expected Behavior

When the user selects all active or decommissioned systems answers to download, only the active or decommissioned systems will have their answers downloaded to an excel file.

### Actual Behavior

The excel file will include the entirety of both active and decommissioned system answers downloaded to an excel file, regardless of if you only selected active or vice versa.

### Environment

Chrome 148 on MacOS Sequoia 15.5

### Acceptance Criteria

- [ ]  Only selected systems answers are download

### Severity

Low - Minor inconvenience

### Affected Component

Download System Answers functionality

### Screenshots

System Names selected for download: SSD-EX and SLD-GEN
Image

System Names downloaded to excel file: SSD-EX, SLD-GEN, and DS-1
Image

### Logs / Error Messages

```shell
No error logs.
```

### Workaround

Only occurs when all rows in active or decommissioned systems are selected. So the workaround is either to download the selected systems separately, not all at once. Or modify the downloaded excel file to delete what is not needed.

costacalvin:
## Summary

Fixes the "Download selected system answers" export including systems from the opposite active/decommissioned view when all visible rows are selected.

Closes #375

## Root cause

`FismaTable.tsx` guarded the `fsids` query parameter construction with:

```js
if (selectedRows.length < fismaSystems.length)
```

`fismaSystems` in context is a filtered subset — either active-only or decommissioned-only depending on the current toggle state. When a user selected all visible rows, `selectedRows.length === fismaSystems.length`, the condition evaluated to `false`, no `fsids` params were appended, and the backend's bare `/export` endpoint returned every system in the database regardless of what was selected.

## Fix

Replaced the length comparison with a simple presence check — if any rows are selected, always build and append `fsids` params. The backend then filters to exactly those systems.

- `FismaTable.tsx` — removed the `fismaSystems.length` guard; `fsids` params are now always sent when `selectedRows` is non-empty.
- `FismaTable.test.tsx` — added unit tests for `CustomFooterSaveComponent` covering: button disabled with no selection, `fsids` sent for a partial selection, and a regression test for the select-all case that triggered this bug.

Possible enhancement for future ticket is allowing the user to download all active and decommissioned system answers all at once, either through a new button or restructuring of how the select system view works.

## Testing

Verified locally:
- Selecting all active systems and downloading produces an export containing only those active systems
- Selecting all decommissioned systems and downloading produces an export containing only those decommissioned systems
- Partial selections correctly pass only the selected `fsids`
- The download button remains disabled when no rows are selected
- All 3 unit tests pass with no console errors

### DIFF
--- a/src/views/FismaTable/FismaTable.test.tsx
+++ b/src/views/FismaTable/FismaTable.test.tsx
@@ -1,0 +1,98 @@
+import { render, screen, fireEvent } from '@testing-library/react'
+import { MemoryRouter } from 'react-router-dom'
+import { CustomFooterSaveComponent } from './FismaTable'
+import type { FismaSystemType } from '@/types'
+
+// GridFooterContainer and GridFooter internally call useGridRootProps which
+// requires the MUI DataGrid context. Stub them out for isolated unit tests.
+jest.mock('@mui/x-data-grid', () => ({
+  ...jest.requireActual('@mui/x-data-grid'),
+  GridFooterContainer: ({ children }: { children: React.ReactNode }) => (
+    <div>{children}</div>
+  ),
+  GridFooter: () => <div data-testid="grid-footer" />,
+}))
+
+jest.mock('@/axiosConfig', () => ({
+  __esModule: true,
+  default: { get: jest.fn() },
+}))
+
+const mockAxiosGet = require('@/axiosConfig').default.get as jest.Mock
+
+// Minimal blob download stubs
+const createObjectURL = jest.fn(() => 'blob:mock')
+const revokeObjectURL = jest.fn()
+Object.defineProperty(window, 'URL', {
+  value: { createObjectURL, revokeObjectURL },
+  writable: true,
+})
+
+const SYSTEMS = [
+  { fismasystemid: 1, fismaname: 'Active A' },
+  { fismasystemid: 2, fismaname: 'Active B' },
+]
+
+const baseProps = {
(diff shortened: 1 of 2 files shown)


tarratsco (approved):

danielbowne:
Rehomed onto a CMS-Enterprise branch and merged as #396 (cherry-picked your commit onto current main so CI could run, authorship preserved). Closing this in favor of that. Thanks @costacalvin!

jsos3-cms:

PR #388, opened 2026-06-11
Centralize 401/403 handling in axios interceptor

danielbowne:
Rehomed from #377 (originally @tarratsco) onto a CMS-Enterprise branch so CI can run, then merged current main in to resolve the conflicts with the multi-OpDiv RBAC work (#378).

Closes #329.

## What this does

A single axios response interceptor owns 401 (redirect to sign-in) and 403 (permission toast). Views call `isAuthHandled(error)` to short-circuit their catch blocks instead of repeating the redirect/snackbar ladder. `parseApiError` (added by #378) stays as the per-call domain layer: 400 field-error maps rendered inline, plus 404/5xx fallback messaging. The two layers sit on top of each other rather than competing.

## Conflict resolution (merge of main, commit ea95194)

#378 rewrote several of the same views after this branch was opened, so the merge needed real reconciliation rather than a straight conflict-resolve:

- **AssignSystemModal, EditSystemModal, QuestionnairePage** — replaced the inline 401/403 ladders with the `isAuthHandled` short-circuit.
- **UserTable** — rebuilt on main's rewritten version (OpDiv grant modal, OpDivs column, selectableRoles gating), removed its `checkValidResponse` / `handleUnautherized` helpers, and routed every catch through `isAuthHandled` with `setSaveError` / `parseApiError` kept for the 400 domain branch.
- **OpDivAdmin, OpDivGrantModal** — these views are new in #378 and did their own 401/403 handling, which would double-fire under the now-global interceptor (two toasts on 403). Added the `isAuthHandled` guard ahead of `parseApiError` so they delegate correctly.
- **yarn.lock** — kept main's axios 1.16 security bump and re-added axios-mock-adapter.

Post-review cleanup (commit 88f9830) removed the now-unreachable 401 branches in the two OpDiv views and extended `authHandling.invariants.test.ts` to cover them.

## Behavior change worth a look

#378 used `ERROR_MESSAGES.outOfScope` for OpDiv-scope 403s. Under the interceptor those now surface the backend message (if present) or the generic `ERROR_MESSAGES.permission`. If we want the scope-specific copy preserved, that is an interceptor-default or backend-message decision, not a per-view one. Flagging for the team.

## Follow-up for @tarratsco

Per our thread on #377: this is green now (tsc, eslint, jest 149 passing). Over to you for the `parseApiError` domain-layer wiring on the OpDiv views and a review pass. The old fork PR #377 can be closed in favor of this one.

PR #378, opened 2026-06-09
feat(admin): multi-OpDiv RBAC admin UI (user/OpDiv management, OpDiv administration, scoped visibility)

danielbowne:
## Summary

Builds the frontend admin surfaces for the multi-OpDiv RBAC model and matches
the UI to the authorization the backend now enforces. Pairs with the backend
`feature/opdiv-rbac-enforcement` branch.

## Deploy ordering (read first)

**Backend deploys first.** This PR depends on the backend RBAC endpoints. The
OpDiv grant endpoints (`/users/{id}/assignedopdivs`) and OpDiv write endpoints
(`/opdivs`) are built to the documented contract and will return 404 until the
backend `feature/opdiv-rbac-enforcement` work is live in the target environment.

Keep this PR in **draft** until the backend is deployed. Moving it out of draft
(and merging) triggers the CI deploy to dev, so un-drafting before the backend
is there would ship a UI calling endpoints that 404. Order: backend live, then
un-draft this.

## What's included

**User management (`/users`)**
- Per-user OpDiv grant management (assign/revoke), with a membership column
- Role dropdown scoped to the acting admin's tier (OWNER -> any; HHS_ADMIN ->
  any except OWNER; OPDIV_ADMIN -> OpDiv tier and below)
- Read-only admins can view with mutating controls withheld
- Edit/assign/delete/restore withheld for users above the acting admin's tier
  (mirrors the server CanManageUser rule); new-user creation unaffected
- `identity_provider` shown read-only (backend-derived from OpDiv); refreshed
  after a grant/revoke since it can change

**OpDiv administration (`/admin/opdivs`, OWNER only)**
- List, create, edit, and deactivate/reactivate OpDivs with inline validation

**Visibility & error handling matched to enforcement**
- Mass email restricted to unscoped write admins (OWNER / HHS_ADMIN)
- Out-of-scope writes (system edit/decommission/reactivate, scoring,
  user-system assignment) show a clear "outside your OpDiv" message instead of
  a raw error
- Scoped list reads are filtered server-side; the UI renders the narrower
  result with no client-side filtering

**Accessibility (Section 508 / WCAG AA)**
- Shared confirm dialog hardened: Escape/backdrop cancel, labelled dialog,
  named close control, verb-first confirm labels; destructive actions remain
  reversible (soft deactivate/restore)
- Accessible names on the data grids and dialogs; form fields properly labelled

## Testing

- Type check, lint, and production build pass.

## Test plan

[...]

PR #393, opened 2026-06-15
Finish notify() and parseApiError adoption across views

### LINKED ISSUE #389: Finish notify() and parseApiError adoption across views
### Feature Type

General Feature

### Recommended labels

`enhancement`, `area/frontend`, `lang/typescript`

#### `notify()` adoption gap

Distribution of remaining `enqueueSnackbar` calls (as of PR #388 merge baseline):

| File                                                | Calls                                                                           |
| --------------------------------------------------- | ------------------------------------------------------------------------------- |
| `src/components/EmailModal/EmailModal.tsx`          | 2                                                                               |
| `src/views/DatacallModal/DataCallModal.tsx`         | 2                                                                               |
| `src/views/AssignSystemModal/AssignSystemModal.tsx` | 4                                                                               |
| `src/views/QuestionnairePage/QuestionnairePage.tsx` | 7                                                                               |
| `src/views/QuestionnareModal/QuestionnareModal.tsx` | 1  (posisbly dead code)|
| `src/views/SystemDetailPage/SystemDetailPage.tsx`   | 9                                                                               |
| `src/views/EditSystemModal/EditSystemModal.tsx`     | 10                                                                              |
| `src/views/UserTable/UserTable.tsx`                 | 5                                                                               |
| **Total**                                           | **40**                                                                          |

Run `grep -rn "enqueueSnackbar(" src/components/EmailModal/ src/views/ --include="*.tsx" --include="*.ts"` to refresh the list before starting.

#### `parseApiError()` adoption gap

Three form views currently hand-roll the 400 field-error dispatch instead of calling parseApiError. Each one:

- Reads `error.response.data.data` directly (couples to the envelope shape).
- Hardcodes per-field copy via `INVALID_INPUT_TEXT(key)` or similar instead of surfacing the backend's specific message.
- Duplicates the status-dispatch logic that parseApiError already centralizes.

File | Has hand-rolled 400 field-error logic? | Notes
-- | -- | --
src/views/DatacallModal/DataCallModal.tsx | Maybe (verify) | Form: name + deadline. Duplicate-name 400 path likely returns field map.
src/views/EditSystemModal/EditSystemModal.tsx | Yes | Largest form in app. Duplicate-acronym, format violations, etc.
src/views/SystemDetailPage/SystemDetailPage.tsx | Yes | Mirrors EditSystemModal's catch. See lines ~300-323 of PR #388 branch.

Reference for the pattern to land: `OpDivAdmin.handleSave` and `UserTable.setSaveError`. Both already use parseApiError correctly. The invariants test (`authHandling.invariants.test.ts`) enforces `parseApiError(error)` as an `includes` needle for the OpDiv views - extend it to these three files once they land.

### Proposed change

Two concerns, bundled because they touch the same files and the same catch blocks.

#### 1. notify() migration

Convert each remaining `enqueueSnackbar` call to `notify()`. Existing durations stay as explicit overrides so the user-facing timing does not change.

**Before**:

```ts
enqueueSnackbar('Saved', {
  variant: 'success',
  anchorOrigin: { vertical: 'top', horizontal: 'left' },
  autoHideDuration: 1500,
})
```

**After**:

```ts
notify('Saved', 'success', { autoHideDuration: 1500 })
```

For calls already using the project defaults (top-left anchor + `DEFAULT_ALERT_TIMEOUT`), the override object can be omitted entirely:

```ts
notify('Saved', 'success')
```

#### 2. parseApiError propagation to form views

Replace the hand-rolled 400 dispatch in `DataCallModal`, `EditSystemModal`, and `SystemDetailPage` with `parseApiError(error)`.

**Before** (from `SystemDetailPage.handleSave`, PR #388 branch, ~lines 300-323):

```ts
.catch((error) => {
  if (isAuthHandled(error)) return
  if (error.response?.status === 400) {
    const data: { [key: string]: string } = error.response.data.data
    Object.entries(data).forEach(([key]) => {
      setFormValid((prev) => ({ ...prev, [key]: false }))
      setFormValidErrorText((prev) => ({
        ...prev,
        [key]: INVALID_INPUT_TEXT(key),  // hardcoded - backend message dropped
      }))
    })
    enqueueSnackbar('Not Saved', { ... })
  } else if (error.response?.status === 404) {
    enqueueSnackbar('System not found', { ... })
  } else {
    notify(ERROR_MESSAGES.tryAgain, 'error', { autoHideDuration: 2500 })
  }
})
```

**After** something along these lines:

```ts
.catch((error) => {
  if (isAuthHandled(error)) return
  const parsed = parseApiError(error)
  if (parsed.fieldErrors) {
    setFieldErrors(parsed.fieldErrors)  // backend per-field copy surfaces inline
    return
  }
  notify(parsed.message, 'error')
})
```

Three concrete wins:

1. Backend per-field copy reaches the user (e.g. "Acronym must be 1-16 characters and unique within OpDiv"), not a hardcoded generic.
2. Envelope-shape knowledge (`error.response.data.data`) lives in one helper instead of N catch blocks.
3. Same `parsed.message` fallback (`ERROR_MESSAGES.tryAgain` by default) for 5xx / network / unknown - no behavior regression on the non-form paths.

#### 3. Tighten invariants for the form views

After the propagation lands, extend `authHandling.invariants.test.ts` for those three files to mirror the OpDiv entries:

```ts
{
  filePath: 'src/views/DatacallModal/DataCallModal.tsx',
  includes: ['if (isAuthHandled(error)) return', 'parseApiError(error)'],
  excludes: [
    'Routes.SIGNIN',
    'ERROR_MESSAGES.permission',
    'error.response.data.data',  // catches the hand-rolled envelope read coming back
  ],
},
```

The `error.response.data.data` exclude is specifically the regression vector this ticket is paying down. Worth keeping out of the file going forward.

### Side benefits

- Views that only used `useSnackbar` for these calls can drop the `import { useSnackbar } from 'notistack'` line and the `const { enqueueSnackbar } = useSnackbar()` destructure. Most of the 8 files qualify.
- One consistent way to fire a snackbar across the codebase. Easier to grep, easier to review.
- Future audit work (e.g. switching default anchor, changing default duration) becomes a one-line edit in `notify.ts` instead of an N-place sweep.

#### notify() migration

- [ ] All 40 `enqueueSnackbar(...)` calls in the listed files are replaced with `notify(...)` calls.
- [ ] User-facing snackbar behavior (message text, variant, duration, anchor) is unchanged for every converted site.
- [ ] `useSnackbar()` is removed from any file where it was only used for the converted calls.

#### parseApiError propagation

- [ ] `DataCallModal.tsx`, `EditSystemModal.tsx`, and `SystemDetailPage.tsx` import `parseApiError` from `@/utils/apiErrors`.
- [ ] Each catch block in those files that previously read `error.response?.status === 400` directly is replaced with `parseApiError(error)` and the `parsed.fieldErrors` / `parsed.status` / `parsed.message` dispatch shape.
- [ ] Per-field copy is surfaced from the backend's reasons (no more hardcoded `INVALID_INPUT_TEXT(key)` for field-level rendering on a 400 with a field map).
- [ ] `authHandling.invariants.test.ts` is extended for those three files: `includes` requires `parseApiError(error)`; `excludes` rejects `Routes.SIGNIN`, `ERROR_MESSAGES.permission`, and `error.response.data.data`.

#### Verification

- [ ] `yarn typecheck` passes.
- [ ] `yarn lint` passes.
- [ ] `yarn test` passes. The `authHandling.invariants.test.ts` suite still passes (the `if (isAuthHandled(error)) return` invariant is unaffected by either change).
- [ ] Manual smoke: trigger one success snackbar, one error snackbar, and one 400 field-error response (e.g. duplicate acronym on EditSystemModal) in the running app to confirm visual rendering and per-field copy match expectations.

### Additional Context

Two helpers shipped during the auth-handling centralization need broader adoption now that the auth work itself is landing in PR #388.

**`notify()` (added in PR #377 / now PR #388).** Lives in `src/utils/notify.ts`, a thin wrapper around notistack v3's standalone `enqueueSnackbar` that bakes in the project-wide defaults (top-left anchor, `DEFAULT_ALERT_TIMEOUT` duration). The goal was to abstract the repetitive snackbar boilerplate so callers could fire a toast in one line.

**`parseApiError()` (added in PR #378).** Lives in `src/utils/apiErrors.ts`, normalizes axios errors into `{ status, fieldErrors?, message }`. The 400 branch surfaces a backend `{ field: reason }` map for inline form rendering; everything else falls back to a backend message or `ERROR_MESSAGES.tryAgain`. The OpDiv views and UserTable already use it. The merge of PR #378 + PR #388 established the two-layer architecture (auth via interceptor, domain via parseApiError) but did not propagate parseApiError to the other form views.

PR #388 converted the cross-cutting 401/403 paths through the axios interceptor and the new generic `tryAgain` fallbacks in `EditSystemModal` and `SystemDetailPage`. It left existing non-auth `enqueueSnackbar` calls and existing hand-rolled `error.response.status === 400` branches alone, because those were domain-specific and out of scope for the auth centralization.

The result is that **`notify()` is partially adopted** (40 `enqueueSnackbar` calls remain across 8 views) and **`parseApiError()` is not adopted by the views that would benefit most** (three form views still hand-roll 400 field-error dispatch with hardcoded fallback strings, discarding backend per-field messages).

tarratsco:
Closes #389.

## Summary

Two helpers shipped during the auth-handling centralization (PR #388 + #378) need broader adoption now that the auth work itself has landed. This PR completes both.

* **`notify()` (added in #388)** is a thin wrapper around notistack's standalone `enqueueSnackbar` that bakes in project defaults (top-left anchor, `DEFAULT_ALERT_TIMEOUT`). Before this PR, 44 direct `enqueueSnackbar` calls remained across 9 view files. All migrated.
* **`parseApiError()` (added in #378)** normalizes axios errors into `{ status, fieldErrors?, message }`. Three form views (`DataCallModal`, `EditSystemModal`, `SystemDetailPage`) still hand-rolled `error.response.data.data` to dispatch 400 field errors. All migrated. Backend per-field copy now surfaces inline instead of the hardcoded `INVALID_INPUT_TEXT(key)` fallback.

While in the area: extracted `STATUS_MESSAGES` for short repeated UI status strings, added `ERROR_MESSAGES.systemNotFound`, dropped the dead `ERROR_MESSAGES.outOfScope` constant, and tightened the auth-handling invariants test to enforce `parseApiError(error)` in the three form views.

## What's in the diff

| Commit | Scope |
|---|---|
| `refactor(notifier): migrate enqueueSnackbar calls to notify() across views` | 44 calls in 9 files; all `useSnackbar()` destructures removed; dead `SNACK_ANCHOR` consts trimmed from OpDiv views |
| `refactor(notifier): propagate parseApiError and extract status constants` | 7 hand-rolled 400/404 catches in 3 form views switched to `parseApiError(error)`; new `STATUS_MESSAGES` + `ERROR_MESSAGES.systemNotFound`; 20 callsites updated to use the new constants |
| `chore(constants): drop dead ERROR_MESSAGES.outOfScope` | 3-line deletion of the unreachable OpDiv-scope 403 copy that #378 introduced and #388 made dead |
| `test(authHandling): tighten invariants for the form views` | Adds `parseApiError(error)` includes and `error.response.data.data` excludes for the three form-view entries; locks out the regression vector this PR pays down |

### Per-field copy on 400 errors

* Before: `setFormValidErrorText(prev => ({ ...prev, [key]: INVALID_INPUT_TEXT(key) }))` (hardcoded per-field text regardless of backend response).
* After: `setFormValidErrorText(prev => ({ ...prev, [key]: message }))` where `message` comes from `parseApiError(error).fieldErrors[key]` (backend's per-field reason reaches the user).

If the backend returns an empty string for a field, the inline text will be empty too. Worth a quick smoke test on a real duplicate-acronym flow.

### "Error: " prefix dropped on 400 messages

`EditSystemModal` and `SystemDetailPage` decommission/reactivate 400 catches previously rendered `notify('Error: ${errorMsg}', 'error')`. Now: `notify(parsed.message, 'error')`. Backend message stands on its own.

### Snackbar timing alignment

Several `autoHideDuration` overrides matched `DEFAULT_ALERT_TIMEOUT` (3000ms) and those overrides are now dropped. A handful of success snackbars in `AssignSystemModal` and the OpDiv views previously relied on notistack's ~5000ms default; they now inherit the helper's 3000ms. Slight tightening, consistent with the helper's purpose.

### `ERROR_MESSAGES.outOfScope` deletion

The constant was unreachable post-#388. The auth-handling invariants test still excludes the substring as a defensive guard against re-introduction. Restoring scope-specific 403 copy is deferred to backend message work (the side that knows why it rejected the request should ship the message).

## Test plan

* [x] `make check` clean (format, lint, build)
* [x] `make test` passes (149 passing)
* [x] Manual smoke: trigger a 400 field-error on `EditSystemModal` (duplicate acronym) and confirm inline error text shows the backend's specific reason, not the generic `INVALID_INPUT_TEXT` fallback
* [x] Manual smoke: trigger one success snackbar and one error snackbar in any flow and confirm visual rendering matches expectations
* [x] Manual smoke: trigger a 404 on `SystemDetailPage`'s decommission flow (Requestly response inject) and confirm the toast reads "System not found"

## Notes
* **`CfactsRecordCard` intentionally not touched.** It uses `skipAuthHandling: true` to render a quiet empty state on 403 instead of a permission toast. The invariants test enforces this exception; it stays out of the `parseApiError` migration by design.
* **Non-form views (EmailModal, AssignSystemModal, QuestionnairePage) still use `notify(ERROR_MESSAGES.tryAgain, 'error')` directly** rather than routing through `parseApiError`. They have no field-error path; the marginal gain of surfacing backend messages there is small enough to leave for a follow-up if desired.

### DIFF
--- a/src/components/EmailModal/EmailModal.tsx
+++ b/src/components/EmailModal/EmailModal.tsx
@@ -15,7 +15,6 @@ import {
   Hint,
 } from '@cmsgov/design-system'
 import SentEmailsModal from './SentEmailsModal'
-import { useSnackbar } from 'notistack'
 import TextField from '@mui/material/TextField'
 import CloseRoundedIcon from '@mui/icons-material/CloseRounded'
 import { EmailModalProps } from '@/types'
@@ -23,7 +22,7 @@ import './EmailModal.css'
 import { styled } from '@mui/material/styles'
 import axiosInstance from '@/axiosConfig'
 import { ERROR_MESSAGES } from '@/constants'
-import { isAuthHandled } from '@/utils/notify'
+import { isAuthHandled, notify } from '@/utils/notify'
 
 const CssTextField = styled(TextField)({
   '& .MuiInputBase-root': {
@@ -57,7 +56,6 @@ const CssTextField = styled(TextField)({
 })
 
 export default function EmailModal({ openModal, closeModal }: EmailModalProps) {
-  const { enqueueSnackbar } = useSnackbar()
   const [sentToEmails, setSentToEmails] = React.useState<string[]>([])
   const [openSentEmailsDialog, setOpenSentEmailsDialog] =
     React.useState<boolean>(false)
@@ -85,24 +83,14 @@ export default function EmailModal({ openModal, closeModal }: EmailModalProps) {
       })
       .then((res) => {
         setSentGroup(groupValue)
-        enqueueSnackbar(`Emails have successfully been sent`, {
-          variant: 'success',
-          anchorOrigin: {
-            vertical: 'top',
-            horizontal: 'left',
-          },
+        notify('Emails have successfully been sent', 'success', {
           autoHideDuration: 2500,
         })
         setSentToEmails(res.data.data)
       })
       .catch((error) => {
         if (isAuthHandled(error)) return
-        enqueueSnackbar(ERROR_MESSAGES.tryAgain, {
-          variant: 'error',
-          anchorOrigin: {
-            vertical: 'top',
-            horizontal: 'left',
-          },
+        notify(ERROR_MESSAGES.tryAgain, 'error', {
           autoHideDuration: 2500,
         })
       })
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -24,8 +24,6 @@ export const ERROR_MESSAGES = {
   error:
     'An error occurred. Please log in and try again. If the error persists, please contact support.',
   permission: 'You do not have permission to do this action.',
-  outOfScope:
-    'This item is outside your OpDiv, so you cannot make changes to it.',
   tryAgain: 'An error occurred, please try again later',
   refresh:
     'Could not refresh the latest data. The information shown may be out of date.',
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -29,6 +29,21 @@ export const ERROR_MESSAGES = {
   tryAgain: 'An error occurred, please try again later',
   refresh:
     'Could not refresh the latest data. The information shown may be out of date.',
+  systemNotFound: 'System not found',
+}
+
+// Short UI status strings for snackbar toasts on save/create flows. Distinct
+// from ERROR_MESSAGES because these are not errors — they are user-facing
+// status indicators returned by the operation. Keeping them grouped avoids
+// the naming collision with ERROR_MESSAGES.notSaved (which is the long-form
+// session-expired warning).
+export const STATUS_MESSAGES = {
+  saved: 'Saved',
+  notSaved: 'Not Saved',
+  created: 'Created',
+  notCreated: 'Not Created',
+  systemDecommissioned: 'System decommissioned successfully',
+  systemReactivated: 'System reactivated successfully',
 }
 export const EMPTY_USER: userData = {
   userid: '',
--- a/src/utils/authHandling.invariants.test.ts
+++ b/src/utils/authHandling.invariants.test.ts
@@ -22,13 +22,23 @@ const expectations: FileExpectation[] = [
   },
   {
     filePath: 'src/views/DatacallModal/DataCallModal.tsx',
-    includes: ['if (isAuthHandled(error)) return'],
-    excludes: ['ERROR_MESSAGES.permission', 'Routes.SIGNIN'],
+    includes: ['if (isAuthHandled(error)) return', 'parseApiError(error)'],
+    excludes: [
+      'Routes.SIGNIN',
+      'ERROR_MESSAGES.permission',
+      'ERROR_MESSAGES.outOfScope',
+      'error.response.data.data',
+    ],
   },
   {
     filePath: 'src/views/EditSystemModal/EditSystemModal.tsx',
-    includes: ['if (isAuthHandled(error)) return'],
-    excludes: ['Routes.SIGNIN'],
+    includes: ['if (isAuthHandled(error)) return', 'parseApiError(error)'],
+    excludes: [
+      'Routes.SIGNIN',
+      'ERROR_MESSAGES.permission',
+      'ERROR_MESSAGES.outOfScope',
+      'error.response.data.data',
+    ],
   },
   {
     filePath: 'src/views/FismaTable/FismaTable.tsx',
@@ -55,8 +65,13 @@ const expectations: FileExpectation[] = [
   },
   {
     filePath: 'src/views/SystemDetailPage/SystemDetailPage.tsx',
-    includes: ['if (isAuthHandled(error)) return'],
-    excludes: ['Routes.SIGNIN'],
+    includes: ['if (isAuthHandled(error)) return', 'parseApiError(error)'],
+    excludes: [
+      'Routes.SIGNIN',
+      'ERROR_MESSAGES.permission',
+      'ERROR_MESSAGES.outOfScope',
+      'error.response.data.data',
+    ],
   },
   {
     filePath: 'src/views/Title/Title.tsx',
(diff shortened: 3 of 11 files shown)


danielbowne:
Rehomed onto a CMS-Enterprise branch and merged as #405 (reconciled onto current main over #394/#399/#400/#404, your commits squashed with authorship preserved, plus completion of the one view the migration had missed). Closing in favor of that. Thanks @tarratsco!

PR #394, opened 2026-06-16
fix(users): fill OpDivs column from inline list grants, drop N+1

danielbowne:
The Users admin table is slow to load. After `GET /users`, `UserTable` fired one `fetchUserOpDivs(userid)` per user (`Promise.all` over the rows) to fill the OpDivs column, because the list response historically returned no grants. At ~640 users that is ~640 requests queued behind the browser per-host connection cap, which is the load delay (the list query itself is ~8ms). It also contributed to the prod Aurora connection storm under the old per-connection data layer.

Now the OpDivs column reads each row's grants from the inline `assignedopdivids` field. No per-row calls on load.

## Safe rollout

Pairs with backend branch `fix/users-list-inline-opdivs`, which `ARRAY_AGG`s each user's grants onto the list row. The frontend falls back to the per-user fetch only when a row lacks the field, so it is correct before and after the backend deploys.

The fallback gate checks **key presence** (`!('assignedopdivids' in u)`), not value. A user with zero grants comes back as `assignedopdivids: null` on the new backend (ARRAY_AGG over no rows is NULL), which is byte-identical to the old backend omitting the field, so a value check would re-fire the N+1 for every grantless user.

## Changes

- `types.ts`: add `assignedopdivids` to the `users` type.
- `UserTable.tsx`: OpDivs column reads `userOpDivMap[userid]` (refresh override) `?? row.assignedopdivids ?? []`; the per-user `Promise.all` backfill is gated on the field being absent. The single-user refresh after the grant modal closes is unchanged.

## Test plan

- [x] tsc, eslint, prettier, jest green (149 passing)
- [x] With the backend branch running: one `/users` request, OpDiv column populated, zero `/users/{id}/assignedopdivs` calls on load
- [x] Against current backend: falls back to per-user fetch, column still correct
- [x] Grant modal add/revoke, then close: single-row refresh updates that row

Note: a separate follow-up will track the `OpDivGrantModal` per-item grant write-loop, which needs a backend batch-grant endpoint.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -217,6 +217,7 @@ export type QuestionChoice = {
 }
 export type users = {
   assignedfismasystems: number[]
+  assignedopdivids?: number[] | null
   email: string
   fullname: string
   role: UserRole
--- a/src/utils/userOpdivs.ts
+++ b/src/utils/userOpdivs.ts
@@ -8,9 +8,10 @@ import axiosInstance from '@/axiosConfig'
  * (GET returns { data: number[] }; POST body { opdiv_id }; DELETE by id).
  * Granting/revoking recomputes the user's derived identity_provider server-side.
  *
- * The list endpoint (GET /users) always returns assignedopdivids as null - the
- * join is skipped there - so callers must use fetchUserOpDivs per user rather
- * than reading a list row.
+ * The list endpoint (GET /users) returns each row's grants inline as
+ * assignedopdivids, so the users table reads them off the row directly.
+ * fetchUserOpDivs is for single-user refresh (e.g. after the grant modal) and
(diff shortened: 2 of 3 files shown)


MackOverflow (dismissed):
Solid frontend complement to the backend PR. The key-presence check is the critical design decision and it's correct. Two minor things to consider before merge:

Scan for other consumers of row.assignedopdivids — confirm nothing else reads it without a null guard, since the type allows null.
Outer .catch on the fallback Promise.all — not a regression from the current code, but worth adding for hygiene since it runs conditionally now and a silent failure would leave the OpDivs column blank with no indication why.
Neither is a blocker. The core logic is sound.

PR #399, opened 2026-06-16
fix(header): hide admin chrome on the signin route

danielbowne:
Rehomed from #391 (originally @tarratsco) onto a CMS-Enterprise branch so CI can run. Cherry-picked his two commits onto current `main`; authorship preserved. Supersedes #391.

Closes #386.

## What this does

`Title` sits above `/signin` (signin is a child of the root layout route), so whenever the session was still valid the admin chrome (datacall label, account chip, avatar, kebab menu) rendered above the "please sign in" prompt. This hides the branded header bar and the datacall sub-bar on the signin route.

Implementation: derive `isSignInRoute` from the normalized pathname (`location.pathname` lowercased, trailing slash stripped) compared to `Routes.SIGNIN`, then wrap the branded header in `{!isSignInRoute && ...}` and gate the datacall sub-bar on `status === 200 && !isSignInRoute`. The UsaBanner and the modal components stay outside the wrapped block, so the authenticated non-signin view is unchanged.

Complementary with #397 (which redirects an active session away from /signin): that handles navigation, this handles presentation.

## Note on the diff size

`Title.tsx` shows a large +/- because the contributor's editor reflowed the file. `git diff -w` reduces it to the ~14 lines of real logic above.

## Invariant test change

The auth-centralization invariant banned the bare substring `Routes.SIGNIN` in `Title.tsx` to stop per-view auth redirects creeping back. This feature legitimately uses `=== Routes.SIGNIN` for a path comparison, not a redirect, so Title's exclude is narrowed to the redirect shapes it actually means to forbid (`navigate(Routes.SIGNIN`, `to={Routes.SIGNIN}`) plus the `useNavigate` hook. Other views keep the bare ban.

## Test plan

- [x] tsc, eslint, jest green (152 passing; invariants Title test passes)
- [x] Manual: on /signin (logged out) the header + datacall bar are hidden, only the login prompt + USA banner show
- [x] Manual: authenticated dashboard still shows the full header + datacall bar
- [x] Manual: account menu, modals, and footer unaffected when authenticated

PR #396, opened 2026-06-16
fix(export): only export selected systems from the current view

danielbowne:
Rehomed from #376 (originally @costacalvin) onto a CMS-Enterprise branch so CI can run. Cherry-picked his commit onto current `main` (his fork branch was 5 commits behind); the change is unchanged and his authorship is preserved.

Fixes #375.

## The bug

In FismaTable, "Download selected system answers" included systems from the opposite active/decommissioned view when all visible rows were selected. The export guard was `selectedRows.length < fismaSystems.length`, so on select-all (`length === length`) it appended no `fsids` and sent a bare `/datacalls/{id}/export`, which the backend treats as "export everything" regardless of the active/decommissioned view.

## The fix

Guard on `selectedRows.length > 0` instead, so the explicit fsids are always sent whenever anything is selected. `fismaSystems` only ever holds the currently-visible view (the toggle re-fetches and replaces the array), and the selection model is materialized to concrete visible row ids, so partial and select-all both resolve to visible-only ids. Symmetric across both views.

## Changes

- `FismaTable.tsx`: export guard `length < total` becomes `length > 0`.
- `FismaTable.test.tsx` (new): covers disabled-when-empty, partial selection, and the select-all regression (#375), asserting the request is not a bare `/export`.

## Test plan

- [x] tsc, eslint, jest green (3 new FismaTable tests pass, including the #375 regression)
- [ ] Manual: active view, select-all, export only active systems
- [ ] Manual: decommissioned view, select-all, export only decommissioned systems
- [ ] Manual: partial selection in each view only the picked systems' fsids

PR #382, opened 2026-06-09
Add ZTMF branding (logos + favicon)

danielbowne:
## ZTMF branding (logos + favicon)

Authored by **@tarratsco** (Onix Tarrats Calderon). This is a rehome of #372 onto a CMS-Enterprise branch so the full CI suite can run — fork PRs don't have access to the deploy/secrets-backed checks. Original commits and authorship are preserved as-is.

Adds ZTMF branding across the app shell: branded header bar with the color logo and title lockup, a logo hero on the login screen, and a generated favicon set (multi-res `.ico`, 512px PNG, apple-touch-icon).

See #372 for the original discussion, screenshots, and review history.

Closes #371

PR #373, opened 2026-06-05
add previous data calls to UI

costacalvin:
## Summary

Fixes the app always showing the latest data call by adding a dropdown so users can view pillar scores and FISMA data for any historical data call.

Closes #25

## Root cause

`Title.tsx` fetched only `/datacalls/latest` and stored a single `latestDataCallId` in context. Every downstream component — `Home.tsx`, `FismaTable.tsx`, and `PillarScoresModal.tsx` — was hardwired to that value with no way for a user to switch to a previous data call year.

## Fix

Switched the fetch in `Title.tsx` from `/datacalls/latest` to `/datacalls`, sorts results newest-to-oldest, and initializes both a `latestDataCallId` (anchor) and a `selectedDataCallId` (user selection) in context. A MUI `<Select>` dropdown in the header lets users pick any available data call. The dropdown is hidden on `/questionnaire` and `/systems/` routes, which still show the static label.

- `Context.ts` — added `selectedDataCallId` and `setSelectedDataCallId` to `ContextType`.
- `Home.tsx` — derives `activeDataCallId` from `selectedDataCallId ?? latestDataCallId` and passes it to the aggregate scores API call.
- `FismaTable.tsx` — reads `activeDataCallId` from context for the export URL and pillar scores modal prop; hides the questionnaire action button when the selected data call is not the latest.
- `PillarScoresModal.tsx` — selects the "current" score by matching `datacallid === selectedDataCallId` first, falling back to the max; filters previous scores using `datacallid < latestScore.datacallid` instead of `!== latestScore.datacallid` to avoid treating a future or mismatched ID as "previous". Removes the `(Latest)` suffix from the heading since the selection is no longer always the latest.
- `types.ts` — removed now-unused `latestDataCallId` from `FismaTableProps`.

## Testing

Tested locally with seed data containing FY24 and FY25 data calls. Switching the dropdown updates pillar scores and the FISMA table for the selected year. The questionnaire button correctly disappears when a historical data call is selected. The questionnaire and system detail pages are unaffected.

Found a small bug unrelated to this work where decomissioned systems are being added to the exported excel file when not selected. Will file it as a bug and get it resolved in another ticket

https://github.com/user-attachments/assets/b4fac9ed-dc77-428d-affb-4904e61b2258

## Questions
[...]

PR #400, opened 2026-06-16
feat(datacalls): view historical data calls via header picker

### LINKED ISSUE #25: feat: ui displays list of data calls, sorted newest to oldest

danielbowne:
Rehomed from #373 (originally @costacalvin) onto a CMS-Enterprise branch so CI can run. His branch was 4 commits behind, so this is a rebase/reconciliation onto current `main` rather than a clean cherry-pick. Authorship preserved (commit authored by Calvin, co-authored by me for the rebase). Supersedes #373.

Closes #25.

## What this does

Adds a datacall picker to the header so users can view pillar scores, FISMA data, and questionnaires for any past data call, not just the latest. Context gains `selectedDataCallId`/`selectedDatacall`; the "active" datacall (selected or latest) is threaded through:

- Home score fetch (`/scores/aggregate` for the active datacall)
- FismaTable export and the pillar-scores modal
- Questionnaire (a historical datacall opens read-only)

`/systems/` detail keeps a static latest label rather than the picker.

## Reconciliation notes

Calvin's branch predated four changes that merged this week. Each was preserved (verified by review):

- **#382** branded header + **#399** hide-header-on-signin: the datacall picker is grafted onto the current branded header's sub-bar and respects the `isSignInRoute` gating, rather than reverting to the old header layout.
- **#396** export-selection fix: the export keeps the `selectedRows.length > 0` guard. Calvin's branch still had the old `length < fismaSystems.length` form (the bug #396 fixed); the reconciliation uses the fixed guard, just retargeted to the active datacall.
- **#394** `assignedopdivids` on the `users` type: untouched.

## Test plan

- [x] tsc, eslint, jest green (152 passing)
- [x] Header datacall dropdown lists all datacalls; selecting an older one re-fetches Home scores + FismaTable for that datacall
- [x] Questionnaire for a historical datacall opens read-only
- [ ] Export downloads the selected datacall's systems, current view only (active/decommissioned)
- [ ] /systems/ detail shows the static latest label, not the dropdown

### DIFF
--- a/src/components/PillarScoresModal/PillarScoresModal.tsx
+++ b/src/components/PillarScoresModal/PillarScoresModal.tsx
@@ -53,6 +53,7 @@ interface PillarScoresModalProps {
   systemName: string
   systemAcronym: string
   scores: ScoreAggregate[]
+  selectedDataCallId: number
 }
 
 // Background color used when the API has not (yet) returned a tier
@@ -66,16 +67,19 @@ const PillarScoresModal: React.FC<PillarScoresModalProps> = ({
   systemName,
   systemAcronym,
   scores,
+  selectedDataCallId,
 }) => {
   const [datacalls, setDatacalls] = useState<DataCall[]>([])
   const [showDataTable, setShowDataTable] = useState(false)
   const dialogRef = useRef<HTMLDivElement>(null)
   const initialFocusRef = useRef<HTMLButtonElement>(null)
 
-  // Get the latest datacall (highest datacallid)
+  // Use the selected datacall if it exists in the scores, otherwise fall back
+  // to the highest datacallid (e.g. when modal is opened before context loads).
   const latestScore =
     scores.length > 0
-      ? scores.reduce((latest, current) =>
+      ? scores.find((s) => s.datacallid === selectedDataCallId) ??
+        scores.reduce((latest, current) =>
           current.datacallid > latest.datacallid ? current : latest
         )
       : null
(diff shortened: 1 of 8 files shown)


MackOverflow (approved):
Solid feature PR with a clean rebase reconciliation. Two things worth a follow-up before or after merge:

Consolidate selectedDataCallId + selectedDatacall into a single selectedDatacall: datacall | null context value — eliminates the sync risk between the two state variables. Low-risk refactor.

electedDataCallId || latestDataCallId — falsy check on 0 (Home.tsx:17, FismaTable.tsx:217) — both files compute activeDataCallId = selectedDataCallId || latestDataCallId. Since selectedDataCallId initializes to 0, this works, but 0 is falsy in JS so a datacall with ID 0 would fall through to latestDataCallId. In practice datacall IDs start at 1, but ?? (nullish coalescing) would be more explicit intent and safer:
const activeDataCallId = selectedDataCallId !== 0 ? selectedDataCallId : latestDataCallId
// or
const activeDataCallId = selectedDataCallId || latestDataCallId  // works, but implicit
A comment noting "0 is the unset sentinel" would prevent a future reader from "cleaning up" the || to ??.

Remove QuestionnairePage's redundant /datacalls/latest fetch — it can derive isHistorical from selectedDataCallId !== latestDataCallId already in context, saving a network call on every questionnaire open.

QuestionnairePage re-fetches /datacalls/latest on every load — even after this PR, the questionnaire fires a fresh GET /datacalls/latest to detect whether selectedDataCallId is historical. Since Title.tsx already has the full sorted list and latestDataCallId in context, the questionnaire could just compare selectedDataCallId !== latestDataCallId from context without the extra network call. Not a blocker, but it's a round-trip that's now avoidable.


Neither is a blocker. The core feature and reconciliation look correct.

costacalvin:
Thanks for the thorough review Mack! I can definitely get these suggestions implemented within the follow-up to this Issue/PR https://github.com/CMS-Enterprise/ztmf-ui/issues/401

jsos3-cms:

jsos3-cms:
Reopened on the project board to backfill the Estimate (story points) for data attribution/reporting, then re-archived. No code, review, or status change.

PR #402, opened 2026-06-17
Implement new authlookup for the entra changes

tarratsco:
Closes #328. Pairs with `CMS-Enterprise/ztmf#341` (dual-IdP backend, dormant behind `entra_enabled`).

## Summary

Replaces the single Sign-in button with an email-input landing page gated by `CONFIG.IDP_ENABLED`. On submit, the page calls `GET /api/v1/auth/lookup?email=<value>`, branches on the response, and redirects:

* `{idp: "okta"}` → `/login` (existing Okta ALB rule, unchanged)
* `{idp: "entra"}` → `/login/entra` (new ALB rule from #341)
* `{idp: null}` / error / timeout → generic "Contact your administrator" inline message; no enumeration signal

When `CONFIG.IDP_ENABLED` is `false` (environments where the multi-IdP rollout hasn't landed), the legacy single-button Okta flow renders unchanged.

Also surfaces `identity_provider` from `/users/current` as a small `(Okta)` / `(Entra)` badge next to the user name in the header for ease of debuggin "I logged in but the dashboard looks wrong" without DevTools.

## What's in the diff

| File | Change |
|---|---|
| `.github/workflows/ui.yml` | Adds `VITE_IDP_ENABLED` env to the BUILD step, gated on `inputs.environment == 'dev'`. Vite inlines `VITE_*` vars at build time, so they have to be set on the build step, not the deployed container. |
| `src/utils/authLookup.ts` |  Thin axios wrapper around the lookup endpoint with a 5s timeout and `skipAuthHandling: true` (the endpoint is unauthenticated; the auth interceptor's 401-redirect would be wrong here). |
| `src/views/LoginPage/LoginPage.tsx` | Rewrite. Splits into three components: `LoginPage` (top-level dispatcher), `IdpLookupLogin` (new email-form variant), `LegacyOktaLogin` (single-button fallback). Plus shared `LoginShell` for the outer layout. |
| `src/views/Title/Title.tsx` | Adds `idpBadge` derived from `userInfo.identity_provider` and renders it next to the name. |

## Behavior change matrix

| `CONFIG.IDP_ENABLED` | Visit `/signin` while authenticated | Visit `/signin` while unauthenticated | Behavior of `/users/current` |
|---|---|---|---|
| `false` | Redirect to `/` (per #397) | Single Sign-in button (legacy) | Unchanged |
| `true` | Redirect to `/` (per #397) | Email landing page with lookup + branching | Unchanged |

[...]

PR #397, opened 2026-06-16
fix(login): redirect active sessions away from /signin; bump ws to 8.21.0

danielbowne:
Rehomed from #392 (originally @tarratsco) onto a CMS-Enterprise branch so CI can run, and folded in the dependabot `ws` security bump (#384) since dependabot PRs can't run the secrets-backed CI either.

Closes #387. Supersedes #392 and #384.

## 1. Redirect signed-in users away from /signin (Onix's fix, #387)

A user with an active session who lands on `/signin` (address bar, bookmark, browser back/forward) used to see `LoginPage` as a dead-end "sign in again" prompt. `LoginPage` now reads the root route's authLoader payload via `useRouteLoaderData(RouteIds.ROOT)` and redirects to the dashboard with `<Navigate to={Routes.ROOT} replace />` when `status === 200`. A logged-out user has no `status`, so they still see the login page (and any session-expired message) as before. No redirect loop: `Title` gates on auth before rendering the `/signin` outlet, so the cycle terminates at Home.

## 2. Bump ws to 8.21.0 (#384)

`ws` is a `resolutions` security override, not a direct dependency. Updated the override `^8.20.0` to `^8.21.0` and regenerated the lockfile (resolves a single `ws@8.21.0`, no peer-dep changes).

## Test plan

- [x] tsc, eslint, jest green (152 passing)
- [ ] Manual: with an active session, visit /signin directly -> redirected to dashboard
- [ ] Manual: logged out, visit /signin -> login page shows (and session-expired message still renders on 401)
- [ ] App boots and builds with ws 8.21.0

Issue #341, opened 2026-03-31
Remediate Section 508 compliance findings from March 2026 TVR


danielbowne:
## Context

CMS Section 508 Program completed accessibility testing of ZTMF Scoring Tool on 2026-03-30. Five defects were identified (4 Sev1, 1 Sev3). Remediation deadlines: Sev1 within 30 days (by 2026-04-29), Sev3 within 90 days (by 2026-06-28).

### Sev1 — 30-day remediation deadline

**04a/04b: Dashboard Actions column — duplicate labels and missing unique identifiers (WCAG 1.3.1, 4.1.2)**
- Interactive elements in the Actions column share the same name/label and are not associated with unique identifiers, making them indistinguishable to assistive technology.
- Also: typo "Questionnare" should be "Questionnaire" in icon label.
- Area: `Dashboard > Layout Table > Actions Table Header`

**08: Dashboard Actions column — view icon missing alt text (WCAG 1.1.1)**
- The view icon has no descriptive tooltip/alt text.
- Area: `Dashboard > Table > Actions column`

**09: Dashboard download button — missing role announcement (WCAG 4.1.2)**
- The download button (appears after selecting system checkboxes) does not announce its purpose or availability to screen readers. An alert/announcement should inform users that selecting a system enables the download button.
- Area: `Dashboard > Table > Checkboxes > Save System Answers`

### Sev3 — 90-day remediation deadline

**10: Pillar Scores modal — insufficient color contrast (WCAG 1.4.3)**
- A. "Hide Data Table" blue text does not meet 4.5:1 contrast ratio against background.
- B. Light gray system name text does not meet 4.5:1 contrast ratio against background.
- Area: `Dashboard > Pillar Scores modal > Overall Score Page`

## Checklist

- [ ] Add unique `aria-label` attributes to Actions column icon buttons (include system name for uniqueness)
- [ ] Fix "Questionnare" → "Questionnaire" typo
- [ ] Add descriptive tooltip/aria-label to view icon
- [ ] Add `aria-live` announcement or role alert for download button availability
- [ ] Fix color contrast on "Hide Data Table" link text
- [ ] Fix color contrast on system name text in Pillar Scores modal
- [ ] Retest with JAWS/NVDA after fixes

## References

- TVR: `Test_Validation_Report_ZTMF Scoring Tool_2026_03-31-2026.xlsx`
- Tools used in testing: JAWS, NVDA, CCA
- Browser: Chrome / Windows

PR #404, opened 2026-06-17
Email pre-auth landing page with IdP lookup (dual-IdP frontend)

### LINKED ISSUE #328: Frontend: Landing page with email input and domain-based IdP routing

danielbowne:
Rehomed from #402 (originally @tarratsco) onto a CMS-Enterprise branch so CI can run. Onix's four commits are cherry-picked unchanged (authorship preserved); one a11y/type follow-up commit is added on top. This is the final frontend piece of the dual-IdP / OIDC Entra work.

Closes #328. Pairs with backend `CMS-Enterprise/ztmf#341` (dual-IdP, dormant behind `entra_enabled`); ops flips dev via `ztmf#347`.

## What it does

Replaces the single Sign-in button with an email pre-auth landing page gated by `CONFIG.IDP_ENABLED`. On submit it calls `GET /api/v1/auth/lookup?email=<value>` (unauthenticated, `skipAuthHandling: true`, 5s timeout) and routes:

- `{idp:"okta"}` -> full-page nav to `/login`
- `{idp:"entra"}` -> `/login/entra`
- `null` / error / timeout -> one generic message, no enumeration signal

When `CONFIG.IDP_ENABLED` is false, the legacy single-button Okta flow renders unchanged. `ui.yml` sets `VITE_IDP_ENABLED` on the BUILD step, true for dev only. Also adds an `(Okta)`/`(Entra)` badge next to the user name for support debugging. #397 (authenticated -> `/`) and #399 (header hidden on `/signin`) are preserved.

## Review

Independent review confirmed the non-enumeration contract holds, the unauthenticated lookup can't leak session or fall through to the interceptor, redirect targets are hardcoded literals (no open-redirect), and #397/#399 are intact. The added follow-up commit addresses the a11y findings:

- email field no longer sets an `aria-label` that overrode the visible label (WCAG 2.5.3)
- lookup error is associated with the input via `aria-describedby` and focus returns to the field on error
- `VITE_IDP_ENABLED` typed as `string` to match Vite's runtime injection

- The 429 rate-limit response currently collapses into the generic message rather than distinct "too many attempts" copy (per #329). Defensible for non-enumeration; left as a product decision.
- `isValidEmailFormat` here duplicates `emailValidator` in `EditSystemModal/validators.ts` — worth consolidating into a shared util.
- No logout affordance exists yet; worth a separate ticket.

## Test plan

- [x] tsc, eslint, jest green (152 passing)
- [x] Local smoke: okta/entra/unknown three-branch routing; invalid format keeps Continue disabled; unknown email shows the generic message
- [x] Authenticated: `(Okta)`/`(Entra)` badge renders; visiting `/signin` redirects to `/`; header hidden on `/signin`
- [x] Deployed-dev once `entra_enabled` flips (ztmf#347): full Entra OIDC handshake -> session -> `/users/current` 200

### DIFF
--- a/.github/workflows/ui.yml
+++ b/.github/workflows/ui.yml
@@ -50,6 +50,11 @@ jobs:
         run: yarn lint
 
       - name: BUILD
+        # VITE_* vars are baked into the bundle at build time, so they must
+        # be set on the build step itself, not on the deployed container.
+        # Default off; flip per environment as IdP rollout progresses.
+        env:
+          VITE_IDP_ENABLED: ${{ inputs.environment == 'dev' && 'true' || 'false' }}
         run: yarn build:${{ inputs.environment }}
 
       - name: Get AWS Creds
--- a/src/utils/authLookup.ts
+++ b/src/utils/authLookup.ts
@@ -1,0 +1,65 @@
+import axiosInstance from '@/axiosConfig'
+
+/**
+ * Identity provider routing values returned by the pre-auth lookup endpoint.
+ * `null` is the deliberate non-enumeration response for both unknown and
+ * unprovisioned emails - the same response shape is returned for both to
+ * avoid leaking which emails exist on the backend side.
+ */
+export type IdpRouting = 'okta' | 'entra' | null
+
+/**
+ * Response shape for GET /api/v1/auth/lookup. The envelope mirrors the
+ * project's other read endpoints: payload under `data`.
+ */
+type LookupResponse = {
+  data: { idp: IdpRouting }
+}
+
+/**
+ * 5-second cap on the unauthenticated lookup call so the Continue button
+ * does not hang indefinitely if the endpoint is slow or unreachable. The
+ * landing page treats a timeout the same as any other failure: show the
+ * generic "Contact your administrator" message, no enumeration signal.
+ */
+const LOOKUP_TIMEOUT_MS = 5000
+
+/**
+ * Pre-auth lookup that maps an email to its identity provider so the
+ * landing page can route the user to the matching /login/<idp> path.
+ *
+ * The endpoint is unauthenticated and rate-limited on the backend, and is
+ * forwarded by a dedicated ALB rule (no OIDC challenge). On the FE side
+ * the only state-bearing decision is the timeout: 5 seconds is enough for
+ * a healthy backend and short enough that a stuck call falls back to the
+ * generic error path quickly.
+ *
+ * Local dev hits the same endpoint via the Vite proxy (vite.config.ts
+ * forwards /api/v1 to VITE_CF_DOMAIN which points at the local backend on
+ * localhost:8080 when `make dev-up` is running). No FE-side mock; the
+ * backend's controller.LookupIdP is the single source of truth even in
+ * local dev.
+ *
+ * @param email - The email entered on the landing page. Not validated
+ *   here; the caller is expected to gate on format before calling.
+ * @returns The resolved IdP marker, or null when the email is not
+ *   provisioned (deliberately indistinguishable from "unknown email" per
+ *   the non-enumeration contract).
+ */
+export async function lookupIdpForEmail(email: string): Promise<IdpRouting> {
+  try {
+    const response = await axiosInstance.get<LookupResponse>('/auth/lookup', {
+      params: { email },
+      timeout: LOOKUP_TIMEOUT_MS,
+      // The lookup is unauthenticated; the interceptor's 401/403 redirect
+      // would be wrong here. Caller treats any failure as a null result.
+      skipAuthHandling: true,
+    })
+    return response.data.data.idp
+  } catch {
+    // Timeout, network error, 4xx, 5xx all collapse to null. The landing
+    // page renders the same generic message either way; that is the
+    // non-enumeration contract.
+    return null
+  }
+}
--- a/src/views/LoginPage/LoginPage.tsx
+++ b/src/views/LoginPage/LoginPage.tsx
@@ -1,4 +1,4 @@
-import { useState, type FormEvent, type ChangeEvent } from 'react'
+import { useState, useRef, type FormEvent, type ChangeEvent } from 'react'
 import { Box, TextField, Typography } from '@mui/material'
 import { Button as CmsButton } from '@cmsgov/design-system'
 import { Navigate, useLocation, useRouteLoaderData } from 'react-router-dom'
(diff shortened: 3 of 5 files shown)


MackOverflow (approved):
Well-structured PR with correct security properties. One nit to verify before merge:

Confirm config.ts coerces VITE_IDP_ENABLED with === 'true', not a bare truthiness check — a string "false" is truthy in JS and would silently enable the new flow on prod. Everything else is solid.

tarratsco:
@MackOverflow  do you mean on this line  in `config.ts`:
`IDP_ENABLED: String(import.meta.env.VITE_IDP_ENABLED) === 'true'`

If that's the case we should be good since the check is explicitly comparing to the string 'true'. So IDP_ENABLED would only resolve to true when VITE_IDP_ENABLED is the string 'true'.

danielbowne:
Confirmed, the coercion is already in place, so no change needed here.

`src/utils/config.ts` line 10:

```ts
IDP_ENABLED: String(import.meta.env.VITE_IDP_ENABLED) === 'true',
```

It uses strict string equality against `'true'`, not a truthiness check, so the string `"false"` evaluates to `false`. Good flag, since a bare `if (import.meta.env.VITE_IDP_ENABLED)` would treat the string `"false"` as truthy and silently enable the email flow in prod. That is not what happens here.

It also lines up with the build side: `ui.yml` sets `VITE_IDP_ENABLED` to the literal string `'true'` for dev only and `'false' for prod/impl, and Vite inlines that string into the bundle. So prod receives `"false"`, the strict check returns `false`, and the legacy Okta sign-in button renders; dev receives `"true"` and gets the email lookup page. `config.ts` is not modified by this PR, so nothing here weakens that guard.

tarratsco:

PR #407, opened 2026-06-17
indicate most-recent and active vs closed status datacalls

costacalvin:
## Summary

Follow-up to #400 (merged). Implements the feedback/suggestions from that @MackOverflow's code review and resolves #401.

### PR #400 review suggestions

**1. Consolidate parallel state variables**
Replaced `selectedDataCallId: number` + `selectedDatacall: string` (two variables that had to be kept in sync) with a single `selectedDatacall: datacall | null` across `Context.ts`, `Title.tsx`, `Home.tsx`, `FismaTable.tsx`, and `QuestionnairePage.tsx`. Uses `null` as the unset sentinel instead of `0`/`''`.

**2. Fix falsy coercion on datacall ID 0**
`selectedDataCallId || latestDataCallId` would incorrectly fall through for a datacall with ID `0`. Replaced with `selectedDatacall?.datacallid ?? latestDataCallId` everywhere — explicit about the unset case.

**3. Remove redundant `GET /datacalls/latest` in QuestionnairePage**
`QuestionnairePage` was firing a fresh network request on every open solely to detect whether the selected datacall was historical. Since `Title.tsx` already fetches the full sorted list, we now expose `latestDeadline` through context (the only missing field) and derive `isHistorical`/`isPastDeadline` from context values — no extra round-trip. Using latestDeadline as the latest datacall can still be "closed".

### Issue #401 — datacall picker UX

- **`Current` chip** on the newest option in the dropdown (derived from `datacallid === latestDataCallId`)
- **Secondary deadline line** under each option: `Active · deadline Jun 30, 2026` or `Closed · deadline Mar 31, 2026` — text-based status for 508/WCAG AA compliance, not color alone
- Chips and secondary line appear only inside the dropdown (`renderOption`); the closed control continues to show only the datacall name via `getOptionLabel` — selected ≠ current visually
- No backend changes; `deadline` was already present on the `datacall` type

## Test plan
- [ ] `yarn test --watchAll=false` — 152 passing, 3 skipped (no regressions)
- [ ] Dashboard loads → Autocomplete shows latest datacall; dropdown options show `Current` chip on newest entry and `Active/Closed · deadline <date>` secondary line
- [ ] Select a historical datacall → Home scores and FismaTable re-fetch for that datacall
- [ ] Open questionnaire for historical datacall → no `GET /datacalls/latest` in network tab → page loads read-only
- [ ] Open questionnaire for latest datacall → no `GET /datacalls/latest` in network tab → editable before deadline
[...]

PR #408, opened 2026-06-18
indicate most-recent and active vs closed status datacalls

### LINKED ISSUE #401: Data call picker: indicate most-recent and active vs closed status
### Feature Type

UX enhancement (frontend-only)

### Workflow Description

Follow-up to #400 (historical data-call picker in the header, which addressed #25). The picker now lists data calls newest-first — e.g. `FY2026 Q1`, `FY2025 Q3`, `FY2023 Q4` — but the options are bare labels. A user cannot tell from the dropdown:

1. **Which data call is the most recent / current one.** The newest is first in the list, but nothing labels it as current; a user scanning the list has to know the fiscal-quarter ordering to infer it.
2. **Whether a given data call is active (open, accepting input) or closed/past-deadline.** Selecting an older call already drops the questionnaire into read-only (QuestionnairePage treats a non-latest selected call as past-deadline), but the picker gives no advance signal, so the read-only state is a surprise after selection rather than something the user can see up front.

The data needed is already on the client: the picker fetches `GET /datacalls`, whose rows include the `deadline`, so active-vs-closed can be derived without any backend change.

### Acceptance Criteria

- [ ] The most recent / current data call is visibly marked in the dropdown (e.g. a `Current` tag or badge on that option).
- [ ] Each option indicates Active vs Closed status, derived from the data call's `deadline` (Active = before deadline, Closed = past deadline). Status is conveyed by text/icon, **not color alone** (508/WCAG AA).
- [ ] Newest-first ordering is preserved.
- [ ] The currently-selected value in the closed control is visually distinct from the "current" marker (selected ≠ current).
- [ ] No backend change required; status is derived from the existing `/datacalls` response.

### Additional Context

UI/UX options to consider (pick during refinement):

- A `Current` chip/tag on the newest data call; a small `Active` / `Closed` status chip per option.
- An optional secondary line under each option showing the deadline date (e.g. "Closed · deadline Mar 31, 2026") for at-a-glance context.
- Keep the control compact — the header bar is dense; a single trailing chip per row reads cleaner than multiple badges.

Image

Refs #400, #25. The repo's CMS Design System / 508 guidance applies to the status treatment (color-independent indicators).

MackOverflow:
## Summary

Follow-up to #400 (merged). Ports changes from costacalvin's fork (PR #407), implementing feedback from the #400 code review and resolving #401.

### Changes

**1. Consolidate parallel state variables**
Replaced `selectedDataCallId: number` + `selectedDatacall: string` (two variables that had to be kept in sync) with a single `selectedDatacall: datacall | null` across `Context.ts`, `Title.tsx`, `Home.tsx`, `FismaTable.tsx`, and `QuestionnairePage.tsx`. Uses `null` as the unset sentinel instead of `0`/`''`.

**2. Fix falsy coercion on datacall ID 0**
`selectedDataCallId || latestDataCallId` would incorrectly fall through for a datacall with ID `0`. Replaced with `selectedDatacall?.datacallid ?? latestDataCallId` everywhere.

**3. Remove redundant `GET /datacalls/latest` in QuestionnairePage**
`QuestionnairePage` was firing a fresh network request on every open solely to detect whether the selected datacall was historical. Since `Title.tsx` already fetches the full sorted list, we now expose `latestDeadline` through context and derive `isHistorical`/`isPastDeadline` from context values — no extra round-trip.

**4. Datacall picker UX**
- `Current` chip on the newest option in the dropdown
- Secondary deadline line under each option: `Active · deadline Jun 30, 2026` or `Closed · deadline Mar 31, 2026`
- Text-based status for 508/WCAG AA compliance (not color alone)
- Closed control continues to show only the datacall name via `getOptionLabel`

## Porting notes

Cherry-picked from costacalvin's fork (PR #407). One conflict in `QuestionnairePage.tsx` resolved by taking Calvin's logic (removal of the `GET /datacalls/latest` call) and dropping the stale `enqueueSnackbar` dep from the effect array (main already migrated to `notify`).

## Test plan
- [ ] `yarn test --watchAll=false` — 152 passing, 3 skipped (no regressions)
- [ ] Dashboard loads → Autocomplete shows latest datacall; dropdown options show `Current` chip on newest entry and `Active/Closed · deadline <date>` secondary line
- [ ] Select a historical datacall → Home scores and FismaTable re-fetch for that datacall
- [ ] Open questionnaire for historical datacall → no `GET /datacalls/latest` in network tab → page loads read-only
- [ ] Open questionnaire for latest datacall → no `GET /datacalls/latest` in network tab → editable before deadline
- [ ] Switch from historical back to latest → `isPastDeadline` resets correctly; questionnaire becomes editable

Closes #401

### DIFF
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -91,7 +91,13 @@ const addSpace = (str: string) => {
   return str
 }
 export default function QuestionnarePage() {
-  const { userInfo, selectedDataCallId, selectedDatacall } = useContextProp()
+  const {
+    userInfo,
+    selectedDatacall,
+    latestDataCallId,
+    latestDatacall,
+    latestDeadline,
+  } = useContextProp()
   const [isPastDeadline, setIsPastDeadline] = React.useState<boolean>(false)
   const isReadOnly =
     isReadOnlyAdmin(userInfo) || (isPastDeadline && !isAdmin(userInfo))
@@ -237,31 +243,23 @@ export default function QuestionnarePage() {
         try {
           let questionsEmpty = false
           let datacall = ''
-          const latestDataCallId = await axiosInstance
-            .get(`/datacalls/latest`)
-            .then((res) => {
-              const latestId = res.data.data.datacallid
-              const isHistorical =
-                selectedDataCallId > 0 && selectedDataCallId !== latestId
-              if (isHistorical) {
-                setDatacallID(selectedDataCallId)
-                datacall = selectedDatacall.replaceAll(' ', '_')
-                setDatacall(datacall)
-                setIsPastDeadline(true)
-              } else {
-                setDatacallID(latestId)
-                datacall = res.data.data.datacall.replaceAll(' ', '_')
-                setDatacall(datacall)
-                setIsPastDeadline(new Date() > new Date(res.data.data.deadline))
-              }
-              return isHistorical ? selectedDataCallId : latestId
-            })
-            .catch((error) => {
-              if (isAuthHandled(error)) return
-              notify(ERROR_MESSAGES.tryAgain, 'error', {
-                autoHideDuration: 2500,
-              })
-            })
+          const isHistorical =
+            selectedDatacall !== null &&
+            selectedDatacall.datacallid !== latestDataCallId
+          let activeDataCallId: number
+          if (isHistorical && selectedDatacall) {
+            setDatacallID(selectedDatacall.datacallid)
+            datacall = selectedDatacall.datacall.replaceAll(' ', '_')
+            setDatacall(datacall)
+            setIsPastDeadline(true)
+            activeDataCallId = selectedDatacall.datacallid
+          } else {
+            setDatacallID(latestDataCallId)
+            datacall = latestDatacall.replaceAll(' ', '_')
+            setDatacall(datacall)
+            setIsPastDeadline(new Date() > new Date(latestDeadline))
+            activeDataCallId = latestDataCallId
+          }
           await axiosInstance
             .get(`/fismasystems/${system}/questions`)
             .then((response) => {
@@ -343,7 +341,7 @@ export default function QuestionnarePage() {
           }
           await axiosInstance
             .get(
-              `scores?datacallid=${latestDataCallId}&fismasystemid=${system}&include=functionoption`
+              `scores?datacallid=${activeDataCallId}&fismasystemid=${system}&include=functionoption`
             )
             .then((res) => {
               const hashTable: questionScoreMap = Object.assign(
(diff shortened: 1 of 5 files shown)


src/views/QuestionnairePage/QuestionnairePage.tsx
```diff
@@ -237,31 +243,23 @@ export default function QuestionnarePage() {
...
+            setDatacallID(selectedDatacall.datacallid)
+            datacall = selectedDatacall.datacall.replaceAll(' ', '_')
+            setDatacall(datacall)
+            setIsPastDeadline(true)
+            activeDataCallId = selectedDatacall.datacallid
+          } else {
+            setDatacallID(latestDataCallId)
+            datacall = latestDatacall.replaceAll(' ', '_')
+            setDatacall(datacall)
+            setIsPastDeadline(new Date() > new Date(latestDeadline))
```
danielbowne: On first render latestDeadline is `''` until the datacalls fetch resolves, and `new Date('')` is an Invalid Date, so this comparison is false and the form is briefly editable past deadline. Default to read-only until the deadline is known: `setIsPastDeadline(latestDeadline ? new Date() > new Date(latestDeadline) : true)`.

danielbowne (commented):
Mostly looks good. The id 0 falsy fix and the null sentinel cleanup are solid, and dropping the extra `/datacalls/latest` round trip is a nice win. One thing on the deadline derivation.

In the non-historical branch isPastDeadline comes straight from latestDeadline off context. On the first render latestDeadline is still the empty string before Title's datacalls fetch resolves, and `new Date('')` is an Invalid Date, so `new Date() > new Date('')` is false. That leaves a past-deadline questionnaire briefly editable until the effect re-runs with the real deadline. It self-corrects since latestDeadline is in the dep array, and the questions are not loaded during that window, so it is a narrow gap rather than an open hole, but I would default to read-only until the deadline is known:

    setIsPastDeadline(latestDeadline ? new Date() > new Date(latestDeadline) : true)

Also worth confirming the backend rejects writes on a closed datacall on its own, so this client gate is not the only thing standing between a late edit and a save.

src/views/QuestionnairePage/QuestionnairePage.tsx
costacalvin: Thanks @danielbowne  for the great catch. I made that change and verified locally. Looks like I need to make another fork PR so here it is: https://github.com/CMS-Enterprise/ztmf-ui/pull/417. 

Also verified the backend throws an error if a non-admin somehow attempts to edit the questionnaire past its deadline. 
image

MackOverflow:
@danielbowne please review

danielbowne (approved):
The deadline guard in 6a64aaa is exactly it, defaulting to read-only until latestDeadline resolves closes the window where a closed datacall was briefly editable. The id 0 falsy fix and the null sentinel cleanup look good, and dropping the extra /datacalls/latest round trip is a nice win. Approving.

PR #409, opened 2026-06-18
refactor: convert mixed async/await+.then/.catch to linear async/await

### LINKED ISSUE #380: Convert mixed async/await + .then/.catch to async/await only
### Feature Type

General Feature

## Recommended labels

`enhancement`, `area/frontend`, `lang/typescript`

## Problem

- Two control-flow styles in one function force the reader to track callback state and `await`-resumption state simultaneously.
- `finally` is awkward to attach correctly when the chain already terminates in `.catch`; loading-state cleanup gets dropped on the error path (this is the same class of bug that left `Home.tsx`'s spinner hung on errors).
- Error handling is split: `try { ... } catch` semantics differ from `.catch(handler)` semantics in subtle ways (rejections inside `.then` callbacks, double-handling).

## Proposed change

Convert each handler to a linear shape:

```ts
// before
const handleSave = async () => {
  await axiosInstance.put(url, payload)
    .then((res) => { /* success */ })
    .catch((error) => {
      if (isAuthHandled(error)) return
      /* error */
    })
    .finally(() => setSaving(false))
}

// after
const handleSave = async () => {
  setSaving(true)
  try {
    await axiosInstance.put(url, payload)
    /* success */
  } catch (error) {
    if (isAuthHandled(error)) return
    /* error */
  } finally {
    setSaving(false)
  }
}
```

Specific guidance per shape:
- **Submit handlers** (EmailModal, DataCallModal, AssignSystemModal): one request per handler; straightforward `try / await / catch`.
- **Multi-step data effects** (QuestionnairePage / QuestionnareModal): three sequential requests can stay sequential under `try` with shared `catch`; the inline `.catch` per request was masking errors, so the converted form actually surfaces problems more uniformly.
- **Mutation handlers with `finally`** (EditSystemModal, SystemDetailPage): the `setSaving(false)` cleanup moves to a real `finally` block; this fixes the latent "saving state never clears" risk in the create path.

## Out of Scope
- AbortController wiring  so in-flight requests can be canceled safely.  That work will be tracked in a separate issue (**add link to issue when created**)
- Replacing `axiosInstance` with a query library or shared request hook. That work will be tracked in a separate issue (**add link to issue when created**)

### Acceptance Criteria

- [ ] No handler or effect in `src/` mixes `await ...promise.then/.catch` chains.
- [ ] `try / await / catch / finally` is the canonical shape across data-fetch and form-submit handlers.
- [ ] `.then` and `.catch` still appear only where they make sense (one-off chains without an enclosing `async` function, or library-required usage).
- [ ] All existing flows still work: forms submit, success/error snackbars fire, redirects on 401 still happen (via the centralized interceptor), `finally` blocks clear loading state on both success and error paths.
- [ ] `yarn typecheck` and `yarn lint` clean.

## Context

Most data-fetch and form-submit handlers in `ztmf-ui` are declared `async` and `await` a promise, then chain `.then(...).catch(...)` on the awaited value. This is the worst of both styles: the `await` does nothing useful because the promise chain already handles the result, success and error logic are nested in callbacks instead of linear control flow, and `finally` is harder to wire correctly.

Sites flagged during initial code review:

- `src/views/Home/Home.tsx` - `fetchScores` (already partially cleaned, but the pattern persists elsewhere).
- `src/views/Title/Title.tsx` - `fetchFismaSystems` and `fetchLatestDatacall` both `async` with chained `.then/.catch`.
- `src/components/EmailModal/EmailModal.tsx` - `submitEmail`.
- `src/views/DatacallModal/DataCallModal.tsx` - `submitDatacall`.
- `src/views/FismaTable/FismaTable.tsx` - `saveSystemAnswers` (`return await axiosInstance.get(...).then(...).catch(...)`).
- `src/views/UserTable/UserTable.tsx` - `processRowUpdate`, `handleConfirmDelete`, `handleConfirmRestore`.
- `src/views/AssignSystemModal/AssignSystemModal.tsx` - inline `onChange` handler that fires the assign request.
- `src/views/EditSystemModal/EditSystemModal.tsx` - every handler (`handleSave` edit, `handleSave` create, `handleDecommission`, `handleReactivate`) is `async` + `.then/.catch/.finally`.
- `src/views/SystemDetailPage/SystemDetailPage.tsx` - same shape across `handleSave`, `handleDecommission`, `handleReactivate`.
- `src/views/QuestionnairePage/QuestionnairePage.tsx` - the ~210-line data-fetching `useEffect` mixes `await` with chained `.then/.catch` for three sequential requests.
- `src/views/QuestionnareModal/QuestionnareModal.tsx` - same shape as the page, plus `handleQuestionnareNext` save logic.

MackOverflow:
## What this PR does

This PR cleans up a code style inconsistency that existed throughout the frontend. Many of our async functions (functions that run in the background, like saving a form or fetching data) were written in two different styles at the same time — a mix of `await` and `.then()/.catch()` chained together. This is confusing to read and caused at least one real bug (described below).

This PR rewrites all of those functions to use a single consistent style: `try / await / catch / finally`. The behavior of the app is **unchanged** — the same things happen on success, the same error messages show, and 401 redirects still work. Only the internal structure of the code is different.

## Why this needed fixing

**Readability:** When a function mixes both styles, a reader has to track two different mental models at once to understand what it does. The `try/await/catch` style reads top-to-bottom like plain English.

**Bug fixed:** Several handlers (`EditSystemModal`, `SystemDetailPage`) were missing a `finally` block because it is awkward to attach correctly to a `.then()/.catch()` chain. This meant the "saving..." spinner could get stuck showing if a network request failed. Those handlers now have a proper `finally { setIsSaving(false) }` so the spinner always clears, whether the save succeeded or failed.

**Error surfacing:** `.catch()` attached to an `await`ed promise can silently swallow errors thrown inside `.then()` callbacks. The `try/catch` form catches everything uniformly.

## Credit

This work was specified in full by **@tarratsco** in issue #380, including identifying every file and handler that needed to change, the exact before/after code shape, and the reasoning behind each category of fix. This PR implements that spec as written. @tarratsco had claimed the issue and would have submitted this — it was ported here by @MackOverflow because @tarratsco is currently working from a fork without enterprise access.

## Files changed

| File | What was converted |
|------|--------------------|
| `src/views/Home/Home.tsx` | `fetchScores` |
| `src/views/Title/Title.tsx` | `fetchFismaSystems`, `fetchDatacalls` |
| `src/components/EmailModal/EmailModal.tsx` | `submitEmail` |
| `src/views/DatacallModal/DataCallModal.tsx` | `submitDatacall` |
| `src/views/FismaTable/FismaTable.tsx` | `saveSystemAnswers` |
| `src/views/UserTable/UserTable.tsx` | `processRowUpdate`, `handleConfirmDelete`, `handleConfirmRestore` |
| `src/views/AssignSystemModal/AssignSystemModal.tsx` | `useEffect` data fetch, `onChange` assign handler, `handleConfirmUnassign` |
| `src/views/EditSystemModal/EditSystemModal.tsx` | `handleSave` (edit + create modes), `handleDecommission`, `handleReactivate`, two name-lookup `useEffect`s |
| `src/views/SystemDetailPage/SystemDetailPage.tsx` | `handleSave`, `handleDecommission`, `handleReactivate`, two name-lookup `useEffect`s, retry-fetch `useEffect` |
| `src/views/QuestionnairePage/QuestionnairePage.tsx` | `saveResponse`, functions/options `useEffect` |
| `src/views/QuestionnareModal/QuestionnareModal.tsx` | `fetchData` `useEffect` (3 sequential requests), `handleQuestionnareNext` save logic, `fetchOptions` `useEffect` |

## How to verify

No new behavior was introduced, so the main thing to verify is that nothing broke:

- [ ] `yarn test --watchAll=false` — 152 passing, 3 skipped (no regressions)
- [ ] `yarn typecheck` — clean
- [ ] `yarn lint` — clean
- [ ] Submit a form (email modal, datacall modal, edit system) — success toast appears, spinner clears
- [ ] Simulate a network error — error toast appears, spinner clears (previously the spinner could get stuck)
- [ ] Navigate the questionnaire — saves work correctly, 401 redirect still fires via the centralized interceptor

Closes #380

### DIFF
--- a/src/views/AssignSystemModal/AssignSystemModal.tsx
+++ b/src/views/AssignSystemModal/AssignSystemModal.tsx
@@ -45,29 +45,37 @@ export default function AssignSystemModal({
   } | null>(null)
   React.useEffect(() => {
     if (open && userid) {
-      axiosInstance.get(`/users/${userid}/assignedfismasystems`).then((res) => {
-        const assignedSys = res.data.data || []
-        setAssignedSystems(assignedSys)
-        // Include all systems in options
-        const systemIds = Object.keys(fismaSystemMap).map(Number)
-        setFismaSystems(systemIds)
-      })
+      async function fetchAssigned() {
+        try {
+          const res = await axiosInstance.get(
+            `/users/${userid}/assignedfismasystems`
+          )
+          const assignedSys = res.data.data || []
+          setAssignedSystems(assignedSys)
+          const systemIds = Object.keys(fismaSystemMap).map(Number)
+          setFismaSystems(systemIds)
+        } catch (error) {
+          if (isAuthHandled(error)) return
+          console.error('Error fetching assigned systems:', error)
+        }
+      }
+      fetchAssigned()
     }
   }, [open, userid, fismaSystemMap])
-  const handleConfirmUnassign = (confirm: boolean) => {
+  const handleConfirmUnassign = async (confirm: boolean) => {
     const target = pendingUnassign
     setPendingUnassign(null)
     if (!confirm || !target) return
-    axiosInstance
-      .delete(`/users/${userid}/assignedfismasystems/${target.systemid}`)
-      .then(() => {
-        setAssignedSystems(target.nextValue)
-        notify('Saved - unassigned system', 'success')
-      })
-      .catch((error) => {
-        if (isAuthHandled(error)) return
-        notify(ERROR_MESSAGES.tryAgain, 'error', { autoHideDuration: 1500 })
-      })
+    try {
+      await axiosInstance.delete(
+        `/users/${userid}/assignedfismasystems/${target.systemid}`
+      )
+      setAssignedSystems(target.nextValue)
+      notify('Saved - unassigned system', 'success')
+    } catch (error) {
+      if (isAuthHandled(error)) return
+      notify(ERROR_MESSAGES.tryAgain, 'error', { autoHideDuration: 1500 })
+    }
   }
 
   return (
@@ -112,7 +120,7 @@ export default function AssignSystemModal({
               )
             }}
             value={assignedSystems}
-            onChange={(_event, newValue) => {
+            onChange={async (_event, newValue) => {
               const added = newValue.filter(
                 (item) => !assignedSystems.includes(item)
               )
@@ -120,20 +128,19 @@ export default function AssignSystemModal({
                 (item) => !newValue.includes(item)
               )
               if (added.length) {
-                axiosInstance
-                  .post(`/users/${userid}/assignedfismasystems`, {
-                    fismasystemid: added[0],
+                try {
+                  await axiosInstance.post(
+                    `/users/${userid}/assignedfismasystems`,
+                    { fismasystemid: added[0] }
+                  )
+                  setAssignedSystems(newValue)
+                  notify('Saved - assign system', 'success')
+                } catch (error) {
+                  if (isAuthHandled(error)) return
+                  notify(ERROR_MESSAGES.tryAgain, 'error', {
+                    autoHideDuration: 1500,
                   })
-                  .then(() => {
-                    setAssignedSystems(newValue)
-                    notify('Saved - assign system', 'success')
-                  })
-                  .catch((error) => {
-                    if (isAuthHandled(error)) return
-                    notify(ERROR_MESSAGES.tryAgain, 'error', {
-                      autoHideDuration: 1500,
-                    })
-                  })
+                }
               } else if (removed.length) {
                 setPendingUnassign({
                   systemid: removed[0],
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -195,42 +195,29 @@ export default function QuestionnarePage() {
     setQuestionId(index)
   }
 
-  const saveResponse = () => {
-    if (scoreid) {
-      axiosInstance
-        .put(`scores/${scoreid}`, {
+  const saveResponse = async () => {
+    try {
+      if (scoreid) {
+        await axiosInstance.put(`scores/${scoreid}`, {
           fismasystemid: system,
           notes: notes,
           functionoptionid: selectQuestionOption,
           datacallid: datacallID,
         })
-        .then(() => {
-          // checkValidResponse(res.status)
-          notify(STATUS_MESSAGES.saved, 'success', { autoHideDuration: 1500 })
-          fetchQuestionScores(system, setQuestionScores)
-        })
-        .catch((error) => {
-          if (isAuthHandled(error)) return
-          console.error('Error updating score:', error)
-          notify(ERROR_MESSAGES.tryAgain, 'error', { autoHideDuration: 2500 })
-        })
-    } else {
-      axiosInstance
-        .post(`scores`, {
+      } else {
+        await axiosInstance.post(`scores`, {
           fismasystemid: system,
           notes: notes,
           functionoptionid: selectQuestionOption,
           datacallid: datacallID,
         })
-        .then(() => {
-          notify(STATUS_MESSAGES.saved, 'success', { autoHideDuration: 1500 })
-          fetchQuestionScores(system, setQuestionScores)
-        })
-        .catch((error) => {
-          if (isAuthHandled(error)) return
-          console.error('Error posting score:', error)
-          notify(ERROR_MESSAGES.tryAgain, 'error', { autoHideDuration: 2500 })
-        })
+      }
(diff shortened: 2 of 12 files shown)


src/views/QuestionnairePage/QuestionnairePage.tsx
```diff
@@ -237,31 +230,23 @@ export default function QuestionnarePage() {
...
+          } else {
+            setDatacallID(latestDataCallId)
+            datacall = latestDatacall.replaceAll(' ', '_')
+            setDatacall(datacall)
+            setIsPastDeadline(new Date() > new Date(latestDeadline))
+            activeDataCallId = latestDataCallId
+          }
           await axiosInstance
             .get(`/fismasystems/${system}/questions`)
             .then((response) => {
```
tarratsco: There are still a few .then/.catch mixed in with the async/await pattern here for the questions and score fetches.

src/views/AssignSystemModal/AssignSystemModal.tsx
```diff
@@ -112,28 +120,27 @@ export default function AssignSystemModal({
               )
             }}
             value={assignedSystems}
-            onChange={(_event, newValue) => {
+            onChange={async (_event, newValue) => {
```
tarratsco, line 123: This is just a note to review event handlers across the codebase and extract inline functions definitions into actual functions.

src/views/UserTable/UserTable.tsx
tarratsco, line 244: User table still uses the .then/.catch pattern in many instances. I recommend to switch to async/await to have a consistent data fetch pattern across the code base.

src/views/UserTable/UserTable.tsx
```diff
@@ -289,45 +289,38 @@ export default function UserTable() {
...
-        })
+        newRow = res.data.data
+        updatedRow.userid = newRow.userid
+        apiRef.current.updateRows([{ userid: curRowUserId, _action: 'delete' }])
+        apiRef.current.updateRows([updatedRow])
+        setSnackBarSeverity('success')
+        setSnackBarText(STATUS_MESSAGES.saved)
+        setOpen(true)
+      } catch (error) {
+        if (isAuthHandled(error)) return updatedRow
```
danielbowne, line 306: Minor, not blocking: this early return on the auth-handled path skips the trailing `setRows(...)` that the old sync version always ran. Harmless here since the interceptor is redirecting to signin, just noting the one spot where behavior differs from the old code.

danielbowne (dismissed):
Read through the whole conversion. It reads a lot cleaner and I do not see any behavior changes that matter. The `finally { setIsSaving(false) }` additions in EditSystemModal and SystemDetailPage are a real fix, the old chains left the spinner stuck on the auth-handled return path.

One small note, not blocking: processRowUpdate in UserTable went from sync to async, so the grid now commits the row after the round trip instead of optimistically. That is actually an improvement since it picks up the server-assigned userid before leaving edit mode. The only divergence is the auth-handled path now early-returns and skips the trailing setRows, but the user is mid-redirect to signin at that point so it does not surface. Fine to leave as is.

Approving.

MackOverflow:
## Follow-up: blocking items from review addressed

Pushed a follow-up commit ([4d765bf](https://github.com/CMS-Enterprise/ztmf-ui/commit/4d765bf)) that resolves the two blocking items flagged in review.

### `QuestionnairePage.tsx` — incomplete conversion (tarratsco's comment)

The questions fetch and scores fetch inside `fetchData` were still using `.then().catch()` chains nested inside the outer `try` block. Both are now `try/await/catch`:

- Questions fetch: converted to `try { const response = await ... } catch`. The empty-questions branch is now an `else` rather than an early `return` inside `.then()`, which reads more clearly. The catch block also adds an explicit `return` so a failed questions request doesn't fall through to attempt the scores fetch (the old code would have continued).
- Scores fetch: converted to its own `try/await/catch` block following the same pattern.

### `UserTable.tsx` — two unconverted useEffects (tarratsco's comment)

Two useEffects were missed in the original pass:

- **Main users load** (`axiosInstance.get('/users', ...)`): wrapped in `async function fetchUsers()`, matching the pattern used in `Home.tsx` and elsewhere. The nested `Promise.all` backfill for OpDiv grants is now `await`ed with its own `try/catch`.
- **OpDiv options load** (`fetchOpDivs(true)`): wrapped in `async function loadOpDivs()`.

**Left intentionally as `.then/.catch`:**
- `refreshUserRow` — the existing comment says "Non-blocking refresh"; it's fire-and-forget and the style is appropriate there.
- The per-item `.catch(() => [u.userid, []])` inside `Promise.all` — standard per-item fallback pattern, not a mixed-style issue.

No behavior changes beyond the QuestionnairePage questions error path (which now skips the scores fetch on failure rather than attempting it — better behavior).

PR #410, opened 2026-06-18
refactor: Add AbortController cancellation to all data-fetch flows (#381)

### LINKED ISSUE #381: Add AbortController cancellation to data-fetch flows
### Feature Type

General Feature

## Recommended labels

## Sequence
- Complete after  https://github.com/CMS-Enterprise/ztmf-ui/issues/380 

## Problem
- Race conditions from late responses could overwrite newer state.
- Work is not aborted when a component unmounts or dependencies change.
- Modal submit handlers can notify/navigate after close.

## Proposed change
1. Add abort-capable request helpers (or hook-level support) on top of `axiosInstance`.
2. Ensure effect-driven fetches abort on cleanup and dependency changes.
3. Ensure async submit handlers are abort-aware where the UI can unmount during in-flight work.
4. Treat cancellation as a non-user-visible control path (not as an error toast path).

### Reference implementation (React + axios + AbortController)
 
 ```ts
 useEffect(() => {
  const controller = new AbortController()

  const load = async () => {
    try {
      // Success branch: request completed before unmount/dependency change.
      const res = await axiosInstance.get('/endpoint', {
        signal: controller.signal,
      })
      setData(res.data)
    } catch (error) {
      // Cancel branch: request was intentionally aborted; do nothing.
      if (controller.signal.aborted) return

      // Error branch: real failure (network/server); surface user feedback.
      notify(ERROR_MESSAGES.tryAgain, 'error')
    }
  }

  load()

  // Cleanup branch: abort in-flight request on unmount/dependency change.
  return () => {
    controller.abort()
  }
}, [dependency])

```

### Acceptance Criteria

- [ ]  Every fetch effect in the listed views has real cancellation, not only boolean guard flags.
- [ ]  No stale state update warnings during rapid route/view switches.
- [ ]  No post-close snackbar/navigation side effects triggered from canceled requests.
- [ ] Existing success/error UX remains unchanged for non-canceled requests.

### Additional Context

During code review theere was a recurring pattern across data-fetching views: requests can continue running after unmount or dependency changes, then update stale state when they resolve. A few files use `let cancelled` guards, but those only suppress state writes and do not cancel in-flight network work.

This affects views and handlers including:
- `EditSystemModal`
- `SystemDetailPage`
- `CfactsRecordCard`
- `PillarScoresModal`
- `UserTable`
- `QuestionnairePage`
- `QuestionnareModal`
- `FismaTable`
- `Title`
- `AssignSystemModal`
- `DataCallModal`
- `Home`
- `EmailModal`

### PR DESCRIPTION
**Pull request by MackOverflow**
## What this PR does

This PR fixes a subtle resource management problem that existed in every component that loads data from the server. When a user navigates away from a page (or changes a filter, closes a modal, etc.) before the server has responded, the old code would let the network request finish, receive the response, and then throw it away — but only *after* React had already unmounted the component or re-rendered with fresh state.

This caused two categories of problem:
1. **Race conditions:** If a user quickly toggled a filter (e.g. "Show Deleted" in the user table), two requests could be in-flight at the same time. Whichever one finished last would overwrite the result of the one that finished first, even if it was stale.
2. **Wasted work:** Requests were completing and processing server responses for components that no longer existed on screen.

This PR adds `AbortController` cancellation to every data-fetch `useEffect` in the app. When a component unmounts or a dependency changes, the cleanup function calls `controller.abort()`, which tells axios to cancel the in-flight HTTP request immediately. The catch block checks `controller.signal.aborted` and returns silently — it was an intentional cancel, not a real error.

The behavior of the app is **unchanged** — the same data loads, the same error messages show. Only the internal resource handling is more correct.

## Before / after

**Before (boolean guard — stops the state write but not the network request):**
```ts
let cancelled = false
axiosInstance.get('/endpoint').then(res => {
  if (!cancelled) setState(res.data)  // discards data, but request already completed
})
return () => { cancelled = true }
```

**After (real cancellation — stops the HTTP request itself):**
```ts
const controller = new AbortController()
async function load() {
  try {
    const res = await axiosInstance.get('/endpoint', { signal: controller.signal })
    setState(res.data)
  } catch (error) {
    if (controller.signal.aborted) return  // cancelled — silent
    notify(ERROR_MESSAGES.tryAgain, 'error')  // real error — surface it
  }
}
load()
return () => { controller.abort() }
```

## Files changed

| File | Change |
|---|---|
| `CfactsRecordCard.tsx` | `let cancelled` + `.then/.catch/.finally` → `AbortController` + `async/await` |
| `EditSystemModal.tsx` | Two user-lookup effects: `let cancelled` → `AbortController` |
| `SystemDetailPage.tsx` | Three effects: `let cancelled` → `AbortController` |
| `UserTable.tsx` | `let ignore` → `AbortController` for the `/users` fetch; the per-user OpDiv backfill keeps a boolean guard since `fetchUserOpDivs` is a helper that doesn't accept a signal |
| `Title.tsx` | `fetchDatacalls` effect: added `AbortController` (previously had no cancellation guard at all) |
| `Home.tsx` | `fetchScores` effect: added `AbortController` |
| `AssignSystemModal.tsx` | `fetchAssigned` effect: added `AbortController` |
| `PillarScoresModal.tsx` | `fetchDatacalls` effect: added `AbortController` |
| `QuestionnairePage.tsx` | Both effects (main data-fetch + per-question options): `AbortController` added, `signal` passed to every `axiosInstance.get` call |
| `QuestionnareModal.tsx` | Both effects (main data-fetch + per-question options): `AbortController` added, `signal` passed to every `axiosInstance.get` call |

## Credit

This work was specified by **@tarratsco** in issue #381. @tarratsco had claimed the issue and would have submitted this — it was ported here by @MackOverflow because @tarratsco is currently working from a fork without enterprise access.

## How to verify

No new behavior was introduced, so the main thing to verify is that nothing broke:

- [ ] `yarn typecheck` — clean
- [ ] `yarn lint` — clean
- [ ] Navigate to a page that loads data (Home, SystemDetailPage, UserTable) — data loads correctly
- [ ] Navigate away quickly before data loads — no React "can't perform state update on unmounted component" warnings in the console
- [ ] Open and close a modal quickly (AssignSystemModal, PillarScoresModal) — no stale state or errors
- [ ] Toggle "Show Deleted" in UserTable rapidly — only the last selection's data renders, no overwrites from stale responses

Closes #381

### DIFF
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -384,44 +384,46 @@ export default function QuestionnarePage() {
     // footer does not flash the previous question's editor during the
     // refetch window.
     setInitQuestionChoice(-1)
-    const choices: QuestionChoice[] = []
-    let funcOptId: number = 0
-    try {
-      axiosInstance
-        .get(`functions/${questionId}/options`, { signal: controller.signal })
-        .then((res) => {
-          res.data.data.forEach((item: QuestionOption) => {
-            const choiceOpt: QuestionChoice = {
-              label: item.description,
-              value: item.functionoptionid,
-            }
-            if (item.functionoptionid in questionScores) {
-              funcOptId = item.functionoptionid
-              choiceOpt.defaultChecked = true
-            }
-            choices.push(choiceOpt)
-          })
-          // Foundation of question
-          setDescription(questionId ? questions[questionId].description : '')
-          setQuestion(questionId ? questions[questionId].question : '')
-          setNotePrompt(questionId ? questions[questionId].notesprompt : '')
-
-          // Notes
-          setNotes(funcOptId ? questionScores[funcOptId].notes : '')
-          setInitNotes(funcOptId ? questionScores[funcOptId].notes : '')
-
-          // Question options
-          setSelectQuestionOption(funcOptId ? funcOptId : -1)
-          setInitQuestionChoice(funcOptId ? funcOptId : -1)
-          setScoreId(funcOptId ? questionScores[funcOptId].scoreid : 0)
-          setOptions(choices ? choices : [])
-          setLoadingQuestion(false)
+    async function load() {
+      const choices: QuestionChoice[] = []
+      let funcOptId: number = 0
+      try {
+        const res = await axiosInstance.get(`functions/${questionId}/options`, {
+          signal: controller.signal,
         })
-    } catch (error) {
-      if (controller.signal.aborted) return
-      if (isAuthHandled(error)) return
-      console.error('Error fetching data:', error)
+        res.data.data.forEach((item: QuestionOption) => {
+          const choiceOpt: QuestionChoice = {
+            label: item.description,
+            value: item.functionoptionid,
+          }
+          if (item.functionoptionid in questionScores) {
+            funcOptId = item.functionoptionid
+            choiceOpt.defaultChecked = true
+          }
+          choices.push(choiceOpt)
+        })
+        // Foundation of question
+        setDescription(questionId ? questions[questionId].description : '')
+        setQuestion(questionId ? questions[questionId].question : '')
+        setNotePrompt(questionId ? questions[questionId].notesprompt : '')
+
+        // Notes
+        setNotes(funcOptId ? questionScores[funcOptId].notes : '')
+        setInitNotes(funcOptId ? questionScores[funcOptId].notes : '')
+
+        // Question options
+        setSelectQuestionOption(funcOptId ? funcOptId : -1)
+        setInitQuestionChoice(funcOptId ? funcOptId : -1)
+        setScoreId(funcOptId ? questionScores[funcOptId].scoreid : 0)
+        setOptions(choices ? choices : [])
+        setLoadingQuestion(false)
+      } catch (error) {
+        if (controller.signal.aborted) return
+        if (isAuthHandled(error)) return
+        console.error('Error fetching data:', error)
+      }
     }
+    load()
     return () => {
       controller.abort()
     }
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -229,158 +229,167 @@ export default function QuestionnarePage() {
   }
 
   React.useEffect(() => {
-    if (system) {
-      // Reset the empty-questionnaire flag so a previous decommissioned-system
-      // view does not bleed into the next render when system changes.
-      setNoQuestions(false)
-      const fetchData = async () => {
-        try {
-          let questionsEmpty = false
-          let datacall = ''
-          const latestDataCallId = await axiosInstance
-            .get(`/datacalls/latest`)
-            .then((res) => {
-              const latestId = res.data.data.datacallid
-              const isHistorical =
-                selectedDataCallId > 0 && selectedDataCallId !== latestId
-              if (isHistorical) {
-                setDatacallID(selectedDataCallId)
-                datacall = selectedDatacall.replaceAll(' ', '_')
-                setDatacall(datacall)
-                setIsPastDeadline(true)
-              } else {
-                setDatacallID(latestId)
-                datacall = res.data.data.datacall.replaceAll(' ', '_')
-                setDatacall(datacall)
-                setIsPastDeadline(new Date() > new Date(res.data.data.deadline))
+    if (!system) return
+    const controller = new AbortController()
+    // Reset the empty-questionnaire flag so a previous decommissioned-system
+    // view does not bleed into the next render when system changes.
+    setNoQuestions(false)
+    const fetchData = async () => {
+      try {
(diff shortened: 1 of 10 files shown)


src/views/QuestionnairePage/QuestionnairePage.tsx
```diff
@@ -229,158 +229,167 @@ export default function QuestionnarePage() {
...
+    const controller = new AbortController()
+    // Clear saved-state markers before async load so the last-edited
+    // footer does not flash the previous question's editor during the
+    // refetch window.
+    setInitQuestionChoice(-1)
+    const choices: QuestionChoice[] = []
+    let funcOptId: number = 0
+    try {
+      axiosInstance
+        .get(`functions/${questionId}/options`, { signal: controller.signal })
```
danielbowne: This get is not awaited and the chain has no `.catch`, so the surrounding try/catch is effectively dead. On abort the CanceledError becomes an unhandled rejection, and on a real error loadingQuestion stays true so the spinner never clears. Await it like the fetchData effect above, or add a `.catch` that checks `controller.signal.aborted` and `isAuthHandled` and calls `setLoadingQuestion(false)`.

src/views/QuestionnareModal/QuestionnareModal.tsx
```diff
@@ -269,87 +269,98 @@ export default function QuestionnareModal({
...
   }, [open, system])
   React.useEffect(() => {
-    if (questionId) {
-      try {
-        axiosInstance.get(`functions/${questionId}/options`).then((res) => {
+    if (!questionId) return
+    const controller = new AbortController()
+    try {
+      axiosInstance
+        .get(`functions/${questionId}/options`, { signal: controller.signal })
```
danielbowne: Same as the options effect in QuestionnairePage. Unawaited `.get(...).then()` with no `.catch` inside a try/catch, so both abort and real errors escape and the loading spinner can hang. Await it, or add the `.catch` with the aborted plus isAuthHandled guards.

danielbowne (commented):
Took a pass through this. The AbortController pattern is right in nearly every effect, the main fetchData effects in both QuestionnairePage and QuestionnareModal await properly and each chain carries its own catch with the aborted guard. Two got missed though, and they are both the per-question options effects.

In each one the options fetch is still a `.get(...).then(...)` chain that is not awaited and has no `.catch()`, sitting inside a try/catch. Since nothing is awaited, the try only catches a synchronous throw, which never happens here, so the `if (controller.signal.aborted) return` line never actually runs. On abort the CanceledError surfaces as an unhandled rejection, and on a real fetch failure there is no notify and `setLoadingQuestion(false)` never fires, so the spinner hangs.

Fix is to await the call inside the try like the neighboring fetchData effect, or attach a `.catch` to the chain that checks aborted plus isAuthHandled and clears the loading flag.

Heads up that this may resolve itself when you rebase onto #409, which already converts these two effects to await. Worth pinning either way so it does not slip through.

MackOverflow:
Good catches, Danny. Fixed in 1f24b64.

Both per-question options effects in `QuestionnairePage` and `QuestionnareModal` had the same bug: the `.get().then()` chain wasn't awaited, so the `try/catch` only wrapped synchronous code. On abort the `CanceledError` became an unhandled rejection; on a real fetch error `setLoadingQuestion(false)` never ran and the spinner would hang.

Restructured both to `async function load()` with the `axiosInstance.get()` awaited inside the `try`, matching the pattern used in the neighboring `fetchData` effects. The abort guard and error path now actually execute.

Re: rebasing onto #409 — noted, I'll keep an eye on it. If #409 lands first I'll rebase and verify the options effects still look right after the merge.

danielbowne (commented):
Re-reviewed after 1f24b64. The two per-question options effects look right now - both are wrapped in an async load() with the get() awaited inside the try, so the abort and auth guards in the catch actually execute and the unhandled-rejection-on-abort is gone. Matches the neighboring fetchData effects. Good.

Before this can merge, it needs a refresh against main - GitHub currently shows it conflicting:

- QuestionnairePage.tsx and Title.tsx conflict with main. These are from #413 (skip the score save on Next when nothing changed), which landed on the same Next-click save path your abort work touches. Please resolve so both survive: the AbortController guards and #413's no-op-save snapshot/guard. That one needs your eye on intent, not a mechanical resolution.
- CfactsRecordCard.tsx: I merged #420 earlier today (parse timestamp-format ATO dates, fixes the 'Invalid Date' on the system detail page). After you pull main in, confirm parseDateOnly still splits on /[ T]/ and formatDate keeps the isNaN guard - a clean automerge can quietly drop those and reintroduce the bug.

One non-blocking note while you are in QuestionnairePage: the options effect's catch still doesn't clear setLoadingQuestion(false) on a real (non-abort, non-auth) error - it only early-returns on those two and then console.errors. So a genuine fetch failure mid-questionnaire leaves the spinner stuck. A finally { setLoadingQuestion(false) } closes it, same idea as the finally blocks in #409. Pre-existing, not introduced here, but easy to fold in during the rebase.

Holding the approval until the merge with main is in - happy to re-review right after.

PR #363, opened 2026-05-21
Last-edited footer on questionnaire and responsive layout cleanup

danielbowne:
## Summary

- Adds a per-question footer showing "Last edited by Name (ROLE) on Date" on the questionnaire page and modal. Reads the new optional last_edited_at and last_edited_by fields off the existing Score response. Renders nothing until backend ships them, so this lands ahead non-destructively.
- Widens the page chrome to full viewport with responsive horizontal padding (16, 32, 64, 96, 128px from xs to xl) and enforces a 800px minimum so layout never crushes. Sidebar and content stay side-by-side; below 800 wide the page falls back to horizontal scroll.
- Restores the dirty check on Next in the modal so read-through navigation does not write a fresh event and overwrite the prior cycle's editor in the audit trail.
- Drops the CMS Design System max-width cap on radio choice rows so option text lines up with the question and notes textarea.
- Fixes breadcrumb display: decode percent-encoded segments, and preserve uppercase casing for the fisma acronym (ACO-MS instead of Aco ms, Acumen gss instead of Acumen%20gss).

Closes #310
Closes #349

Depends on CMS-Enterprise/ztmf#319 for the backend fields. The frontend null paths cleanly while the backend PR is in flight.

## Test plan

- [x] yarn dev, sign in as ISSO on an assigned system, open a saved question, footer shows the prior editor and timestamp with email and ISO timestamp on hover
- [x] Edit the answer and click Next, footer on return shows current user as editor
- [x] Click Next on an unchanged question, footer keeps showing the prior editor (no PUT issued)
- [x] Sign in as ADMIN, footer renders on the same questions
- [x] Resize between 800 and 1920 wide, sidebar and content stay side-by-side and use the available space
- [x] Resize below 800 wide, horizontal scroll appears instead of stacking or crushing
- [x] Breadcrumb shows ACO-MS and Acumen gss correctly on systems whose acronym contains a hyphen or space
- [x] yarn lint, npx tsc --noEmit, and yarn jest are clean

PR #413, opened 2026-06-22
fix(questionnaire): skip score save on Next when nothing changed

### LINKED ISSUE #412: Questionnaire "Next" button re-stamps last-edited editor when no answer or notes changed
## Description

On the questionnaire page, advancing through questions with the **Next** button writes a score record on every click, even when the user changed neither the selected answer nor the notes. Because the backend stamps `last_edited_by` / `last_edited_at` on every write, paging forward silently re-stamps the current user and timestamp onto questions they only read through, overwriting the genuine last editor in the audit trail.

This worked correctly before the last-edited footer was added. The unconditional save on Next is long-standing, but it was harmless when writes were idempotent. Once last-edited tracking shipped (Refs #363), a no-op save became a visible data-integrity bug.

Root cause is in `src/views/QuestionnairePage/QuestionnairePage.tsx`. The **Next** handler calls `saveResponse()` whenever `!isReadOnly`, with no dirty check, while the **Back** button and the sidebar list-item click both gate navigation on a dirty comparison first:

```
(selectQuestionOption !== -1 && initQuestionChoice !== selectQuestionOption) || initNotes !== notes
```

`saveResponse()` itself has no guard, so the Next path saves unconditionally.

## Steps to Reproduce

1. Open a system's questionnaire as a non-read-only user (ISSO/Admin), in an open data call.
2. Open a question that was last answered by a different user (note the "last edited by" footer).
3. Without changing the answer or notes, click **Next**.
4. Return to that question and observe the last-edited footer.

## Expected Behavior

Clicking **Next** without changing the answer or notes performs no write. The score record's `last_edited_by` / `last_edited_at` remain whatever they were; the audit trail reflects the last user who actually changed the response.

## Actual Behavior

Clicking **Next** with no changes issues a PUT to `scores/{scoreid}`, re-stamping `last_edited_by` to the current user and `last_edited_at` to now. The last-edited footer flips to the current user even though nothing was edited. For untouched/unanswered questions, the Next path can also POST a score with `functionoptionid: -1`.

## Environment

All browsers; ZTMF UI (React SPA), current main.

## Acceptance Criteria

- [ ] Clicking Next with no answer/notes change issues no API call and does not alter last_edited_by/last_edited_at
- [ ] Clicking Next after a real answer or notes change still saves (no regression to the genuine save path)
- [ ] Paging through unanswered questions does not POST a score with functionoptionid: -1
- [ ] Last-edited footer reflects only real edits
- [ ] Test added/updated to prevent recurrence

## Severity

Critical - Blocks core functionality

## Affected Component

Questionnaire page (Next navigation / score autosave)

## Proposed Fix

Centralize the dirty check inside `saveResponse()` so the Next path and any future caller are covered:

```js
const saveResponse = () => {
  const isDirty =
    (selectQuestionOption !== -1 && initQuestionChoice !== selectQuestionOption) ||
    initNotes !== notes
  if (!isDirty) return

  // existing PUT (scoreid) / POST branches unchanged
}
```

Refs #363

danielbowne:
## Summary

The questionnaire **Next** button called `saveResponse()` unconditionally for any non-read-only user. Because the backend stamps `last_edited_by` / `last_edited_at` on every write, paging forward through questions silently re-stamped the current user onto answers they only read through, overwriting the genuine last editor in the audit trail. For unanswered questions the same path could POST `functionoptionid: -1` (the no-answer sentinel).

This worked correctly before the last-edited footer shipped (Refs #363): the unconditional save was harmless when writes were idempotent, but once last-edited tracking landed a no-op save became a visible data-integrity bug.

## Fix

New pure helper `shouldPersistResponse()` (`src/views/QuestionnairePage/saveGuard.ts`) returns true only when the user actually changed the answer or the notes **and** an answer is selected. `saveResponse()` early-returns through it, so the Next path and any future caller are covered in one place. The `-1` guard only ever affects the new-score POST path, since a loaded answer already forces a real `functionoptionid` and radios cannot be deselected back to `-1`.

The Back button and sidebar navigation are unchanged — they gate their own confirm dialog and never called `saveResponse()`.

## Acceptance criteria

- [x] Clicking Next with no answer/notes change issues no API call and does not alter `last_edited_by` / `last_edited_at`
- [x] Clicking Next after a real answer or notes change still saves
- [x] Paging through unanswered questions does not POST a score with `functionoptionid: -1`
- [x] Last-edited footer reflects only real edits
- [x] Test added (`saveGuard.test.ts`)

## Testing

- `npx tsc --noEmit` clean
- `npx jest saveGuard` — 5/5 pass, covering the full AC matrix including notes-typed-on-an-unanswered-question

Closes #412

### DIFF
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -37,6 +37,7 @@ import ConfirmDialog from '@/components/ConfirmDialog/ConfirmDialog'
 import { useContextProp } from '../Title/Context'
 import { isAdmin, isReadOnlyAdmin } from '@/utils/userRoles'
 import LastEditedFooter from './LastEditedFooter'
+import { shouldPersistResponse } from './saveGuard'
 type Category = {
   name: string
   steps: FismaQuestion[]
@@ -190,6 +191,16 @@ export default function QuestionnarePage() {
   }
 
   const saveResponse = () => {
+    if (
+      !shouldPersistResponse({
+        selectQuestionOption,
+        initQuestionChoice,
+        notes,
+        initNotes,
+      })
+    ) {
+      return
+    }
     if (scoreid) {
       axiosInstance
         .put(`scores/${scoreid}`, {
--- a/src/views/QuestionnairePage/saveGuard.test.ts
+++ b/src/views/QuestionnairePage/saveGuard.test.ts
@@ -1,0 +1,46 @@
+import { shouldPersistResponse, ResponseState } from './saveGuard'
+
+const base: ResponseState = {
+  selectQuestionOption: 5,
+  initQuestionChoice: 5,
+  notes: 'hello',
+  initNotes: 'hello',
+}
+
+describe('shouldPersistResponse', () => {
+  it('returns false when neither answer nor notes changed (no re-stamp on Next)', () => {
+    expect(shouldPersistResponse(base)).toBe(false)
+  })
+
+  it('returns true when the answer changed', () => {
+    expect(shouldPersistResponse({ ...base, selectQuestionOption: 7 })).toBe(
+      true
+    )
+  })
+
+  it('returns true when the notes changed and an answer is selected', () => {
+    expect(shouldPersistResponse({ ...base, notes: 'edited' })).toBe(true)
+  })
+
+  it('returns false for an unanswered question with unchanged notes', () => {
+    expect(
+      shouldPersistResponse({
+        selectQuestionOption: -1,
+        initQuestionChoice: -1,
+        notes: '',
+        initNotes: '',
+      })
+    ).toBe(false)
+  })
+
+  it('returns false when notes were typed but no answer is selected (no functionoptionid: -1 POST)', () => {
+    expect(
+      shouldPersistResponse({
+        selectQuestionOption: -1,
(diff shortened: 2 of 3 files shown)


MackOverflow (dismissed):
clean Danny, one scemantics thing: 

. Logic redundancy in shouldPersistResponse — minor

// saveGuard.ts:18-22
const isDirty =
  (s.selectQuestionOption !== -1 &&
    s.initQuestionChoice !== s.selectQuestionOption) ||
  s.initNotes !== s.notes
return isDirty && s.selectQuestionOption !== -1
The s.selectQuestionOption !== -1 check appears twice. The first branch of isDirty already requires it for the answer-changed case, but the notes-changed case (s.initNotes !== s.notes) can still set isDirty = true even when selectQuestionOption === -1 — then the final && s.selectQuestionOption !== -1 knocks it back down.

This is correct but the isDirty expression is slightly misleading since it can be true for a state that will ultimately return false. A cleaner read:

export const shouldPersistResponse = (s: ResponseState): boolean => {
  if (s.selectQuestionOption === -1) return false
  return (
    s.selectQuestionOption !== s.initQuestionChoice ||
    s.notes !== s.initNotes
  )
}
Same behavior, one -1 guard, easier to audit.

MackOverflow (approved):

PR #414, opened 2026-06-22
Feat/411 score diff

costacalvin:
## Summary

Implements [UI Issue #411](https://github.com/CMS-Enterprise/ztmf-ui/issues/411) — consumes the `GET /api/v1/scores/diff` endpoint to surface questionnaire answer changes between two data calls for a FISMA system.

Pillar grouping is resolved on the frontend via a separate `GET /fismasystems/:id/questions` call (cached, 10-min TTL). This keeps the diff endpoint lean; adding pillar fields to the diff response is straightforward if needed later.

### Changes

**New component: `ScoreDiffModal`** (`src/components/ScoreDiffModal/ScoreDiffModal.tsx`)
- Opens from a "Compare Datacalls" button added to the `QuestionnairePage` header
- **To picker defaults to the currently viewed data call** (`selectedDataCallId` prop); From defaults to the immediately preceding data call — so the common "what changed since last cycle" case requires no extra interaction
- Pickers show Current chip, Active/Closed status, and deadline date — consistent with the data call picker in #401
- Diff results table groups changed functions **by pillar**, ordered to match the questionnaire's canonical pillar and function ordering (`PILLAR_ORDER` / `PILLAR_FUNCTION_MAP`)
- Each row shows: Function, full Question text (wrapping), From answer, To answer (with score and notes inline), Changed by, Changed at (date + time)
- Null sides render as *No answer* for clean one-sided rows (covers both newly-answered and no-longer-answered functions)
- Empty state: "No changes between these two datacalls."
- Datacalls and questions fetches are cached (10-min module-level TTL) to avoid redundant requests on re-open
- `diffKey` counter forces a fresh diff fetch on every re-open even when picker references haven't changed (React Object.is equality edge case with cached objects)

**`src/types.ts`**
- Added `ScoreDiffSide` and `ScoreDiffEntry` types matching the `/scores/diff` response shape

**`src/views/QuestionnairePage/QuestionnairePage.tsx`**
- Added "Compare Datacalls" button and wired `ScoreDiffModal` with `selectedDataCallId`

**`src/components/ScoreDiffModal/ScoreDiffModal.test.tsx`**
- Unit tests covering: dialog visibility, title, loading states, picker defaults, diff URL params, empty/error states, table rendering (optionname, score, notes, attribution), null sides, pillar ordering, close behavior, and re-open re-fetch

### Acceptance Criteria Coverage

| Criterion | Status |
|---|---|
| Pick two datacalls, view per-function diff | ✅ |
[...]

PR #416, opened 2026-06-22
feat(#411): add Compare Datacalls modal to QuestionnairePage

### LINKED ISSUE #411: Compare data calls: show what changed in a system's questionnaire (and who changed it) — consume /scores/diff
### Feature Type

UI Component (frontend-only — consumes an existing API, no backend change)

### Workflow Description

Backend PR CMS-Enterprise/ztmf#355 (closes CMS-Enterprise/ztmf#133) added a score-diff endpoint, but nothing in the UI consumes it yet, so there is no user-facing way to see it:

`GET /api/v1/scores/diff?from={datacallid}&to={datacallid}[&fismasystemid={id}]`

It compares the answers (functionoptions) of two data calls and returns **only the questionnaire functions whose answer changed** between them — the unchanged ones are dropped — each annotated with who made the later change and when (from the events audit trail). That is the "show what changed in a data call questionnaire and who made the changes when" ask from #133.

We want a frontend surface that lets a user pick two data calls and see what changed in a system's questionnaire across that period: which functions moved, the old → new answer (and notes), and the attribution. A natural home is alongside the per-system views — a "What changed" / "Compare data calls" view or modal reachable from the QuestionnairePage and/or the FismaTable, scoped to the system in view.

The pieces to build against already exist on the client:

- The header data-call picker (#400, refined in #401) already fetches `GET /datacalls` and lists them newest-first — this feature needs a **two**-call selection (a `from` and a `to`), not the single selected call the header uses today.
- `Home.tsx` already fetches `/scores/aggregate?datacallid=…` and `QuestionnairePage` is the per-system questionnaire, so the system context (`fismasystemid`) and datacall list are both already available to seed the picker.

The response shape (one entry per changed function):

```jsonc
{
  "fismasystemid": 1001,
  "functionid": 7007,
  "function": "Battle Station Device Management",
  "question": "Does your system track all assets with comprehensive inventory?",
  "from": { "scoreid": 9001, "functionoptionid": 25, "optionname": "Defined", "score": 2, "notes": "…" },
  "to":   { "scoreid": 9025, "functionoptionid": 26, "optionname": "Managed", "score": 3, "notes": "…" },
  "changed_at": "2026-06-18T16:20:18Z",
  "changed_by": { "userid": "…", "name": "Grand Moff Tarkin", "email": "…", "role": "OWNER" }
}
```

- `from` / `to` are the answer in each cycle; **either can be `null`** when a function was answered in only one of the two cycles (newly answered, or no longer answered) — the UI must handle a one-sided row.
- `changed_at` / `changed_by` may be **absent** when there is no recorded write event for the later answer (older/seed data, or a function answered only in the `from` cycle) — render an "unknown" rather than assuming attribution exists.
- Tier scoping is enforced server-side (unscoped admins see all systems, OpDiv-scoped admins their OpDivs', ISSO/ISSM their assigned systems), so the FE renders whatever it gets back; pass `fismasystemid` when the user is looking at a single system's questionnaire.

### Acceptance Criteria

- [ ] A user can select two data calls (a `from` and a `to`) and view the diff for a system's questionnaire, consuming `GET /api/v1/scores/diff`.
- [ ] Each changed function shows: the function/question, the old → new answer (`optionname` + maturity `score`), the notes on each side, and "changed by {name} ({role}) on {date}" when attribution is present.
- [ ] One-sided changes (`from` or `to` is `null`) render clearly as newly-answered / removed rather than breaking or showing a blank cell.
- [ ] Missing attribution (`changed_by`/`changed_at` absent) renders an explicit "unknown" instead of an empty or fabricated editor.
- [ ] An empty result (no changes between the two cycles) shows a clear empty state, not a blank table.
- [ ] `from` and `to` are required and must differ; the control prevents picking the same call on both sides (the API returns 400 for missing/equal cycles — surface that gracefully if it occurs).
- [ ] Status/meaning is conveyed by text/icon, **not color alone** (508 / WCAG 2.1 AA).
- [ ] No backend change required; the view derives everything from the new endpoint and the existing `/datacalls` response.

### Additional Context

Paired backend work: CMS-Enterprise/ztmf#133 (API) / PR CMS-Enterprise/ztmf#355 (merged endpoint). This is the frontend half of that ticket.

UI/UX options to settle during refinement:

- Placement: a "Compare" / "What changed" tab or modal off the QuestionnairePage (system already in context), and/or a row action on the FismaTable.
- Layout: a per-function table (Function | From → To | Changed by / when) reads cleanly; consider grouping by pillar to match the questionnaire's pillar ordering (#352).
- Default selection: pre-fill `to` = current data call and `from` = the previous one so the common "what changed since last cycle" case is one click; reuse the newest-first `/datacalls` ordering and the current/active-vs-closed treatment from #401.

Related: #332 (ISSO executive dashboard) calls for a historical trend view across data calls — this per-pair diff is the narrower, function-level "what changed and who changed it" companion to that, and could later feed it.

MackOverflow:
## In-repo rehome of #414 by @costacalvin (Calvin Costa) and Cameron

This is the in-repo version of fork PR #414, originally authored by **Calvin Costa** (@costacalvin) and **Cameron**. Commits have been pushed unchanged — authorship is preserved. The move to an in-repo branch is required so org-secret CI gates (Snyk, deploy) can run; those gates cannot execute on fork PRs.

`Refs #414` | Closes #411

---

## Summary

Implements [UI Issue #411](https://github.com/CMS-Enterprise/ztmf-ui/issues/411) — consumes the `GET /api/v1/scores/diff` endpoint to surface questionnaire answer changes between two data calls for a FISMA system.

Pillar grouping is resolved on the frontend via a separate `GET /fismasystems/:id/questions` call (cached, 10-min TTL). This keeps the diff endpoint lean; adding pillar fields to the diff response is straightforward if needed later.

### Changes

**New component: `ScoreDiffModal`** (`src/components/ScoreDiffModal/ScoreDiffModal.tsx`)
- Opens from a "Compare Datacalls" button added to the `QuestionnairePage` header
- **To picker defaults to the currently viewed data call** (`selectedDataCallId` prop); From defaults to the immediately preceding data call — so the common "what changed since last cycle" case requires no extra interaction
- Pickers show Current chip, Active/Closed status, and deadline date — consistent with the data call picker in #401
- Diff results table groups changed functions **by pillar**, ordered to match the questionnaire's canonical pillar and function ordering (`PILLAR_ORDER` / `PILLAR_FUNCTION_MAP`)
- Each row shows: Function, full Question text (wrapping), From answer, To answer (with score and notes inline), Changed by, Changed at (date + time)
- Null sides render as *No answer* for clean one-sided rows (covers both newly-answered and no-longer-answered functions)
- Empty state: "No changes between these two datacalls."
- Datacalls and questions fetches are cached (10-min module-level TTL) to avoid redundant requests on re-open
- `diffKey` counter forces a fresh diff fetch on every re-open even when picker references haven't changed (React Object.is equality edge case with cached objects)

**`src/types.ts`**
- Added `ScoreDiffSide` and `ScoreDiffEntry` types matching the `/scores/diff` response shape

**`src/views/QuestionnairePage/QuestionnairePage.tsx`**
- Added "Compare Datacalls" button and wired `ScoreDiffModal` with `selectedDataCallId`

**`src/components/ScoreDiffModal/ScoreDiffModal.test.tsx`**
- Unit tests covering: dialog visibility, title, loading states, picker defaults, diff URL params, empty/error states, table rendering (optionname, score, notes, attribution), null sides, pillar ordering, close behavior, and re-open re-fetch

**`docs/ScoreDiffModal_Documentation.docx`**
- In-depth technical documentation covering props, state, all five useEffect lifecycle hooks, groupedResults algorithm, API contract, accessibility implementation, design decisions, and the manual test plan

### Acceptance Criteria Coverage

| Criterion | Status |
|---|---|
| Pick two datacalls, view per-function diff | ✅ |
| From → To answer with score and notes | ✅ |
| One-sided rows (new / removed answers) | ✅ |
| Attribution (changed by / changed at) | ✅ |
| Empty state | ✅ |
| Default selection: currently viewed and previous cycle | ✅ |
| Prevent same-call selection | ✅ via `getOptionDisabled` |
| Pillar grouping matching questionnaire order | ✅ |
| 508 / WCAG AA | ✅ text-only status labels, ARIA on dialog and inputs |
| No backend changes | ✅ frontend only |

---

### Unit tests — ✅ 13/13 passed

Run: `yarn test --watchAll=false --testPathPattern="ScoreDiffModal"`

| # | Test | Result |
|---|---|---|
| 1 | `open=false` → no dialog | ✅ |
| 2 | System name/acronym in title | ✅ |
| 3 | Spinner while datacalls loading | ✅ |
| 4 | To/From defaults from `selectedDataCallId` | ✅ |
| 5 | Diff called with correct query params | ✅ |
| 6 | Empty state message | ✅ |
| 7 | Error alert on diff failure | ✅ |
| 8 | Full row render (function, question, sides, attribution) | ✅ |
| 9 | Notes shown/hidden conditionally | ✅ |
| 10 | Null side renders "No answer" | ✅ |
| 11 | Pillar groups ordered by `PILLAR_ORDER` | ✅ |
| 12 | Both close controls call `onClose` | ✅ |
| 13 | Re-fetch on modal reopen | ✅ |

> **Known lint warnings (not failures):** Tests 6, 7, and 9 trigger `jest/expect-expect` because they use `screen.findByText` as an implicit assertion without an explicit `expect()` call. These are warnings only — all 13 tests pass. Should be refactored to explicit assertions in a follow-up.

### UI / Manual (frontend functionality)

Verified locally against the dev stack (`make full-stack-up`):

- [x] "Compare Datacalls" button visible in the QuestionnairePage breadcrumb bar
- [x] Modal opens with correct system name and acronym in title (`Star Destroyer Chimaera Command Systems (isd-chi) — Compare Datacalls`)
- [x] To picker defaults to currently viewed data call; From defaults to prior data call
- [x] Arrow icon renders correctly between From → To pickers
- [x] Error state ("Failed to load diff. Please try again.") renders correctly when `/scores/diff` returns a non-200 — expected until the backend endpoint is implemented
- [x] Close (×) button dismisses the modal
- [ ] Diff table with pillar groups — **blocked pending backend `GET /scores/diff` implementation**
- [ ] Empty state ("No changes between these two datacalls.") — blocked pending backend
- [ ] Null `from`/`to` sides render as *No answer* — blocked pending backend
- [ ] Re-open re-fetches fresh diff — blocked pending backend

> **Note:** All UI-testable frontend behavior is verified. The remaining items require the backend `GET /api/v1/scores/diff` endpoint to be implemented before end-to-end testing can complete.

---

### UI screenshots (from original PR)

New Compare Datacalls button within the questionnaire:
Compare Datacalls button

ScoresDiffModal:
ScoreDiffModal


src/components/ScoreDiffModal/ScoreDiffModal.tsx
```diff
@@ -0,0 +1,586 @@
...
+      sorted.push({
+        pillar: { pillar: 'Other', pillarid: -1, order: 999 },
+        entries: uncategorized,
+      })
+    }
+    return sorted
+  }, [diffResults, functionPillarMap])
+
+  // Fetch diff whenever both pickers are set and different
+  useEffect(() => {
```
danielbowne, line 218: This diff fetch has no AbortController or sequence guard. If From or To changes while a request is in flight, both fire and whichever resolves last wins, so an older response can overwrite a newer one. It also writes state after the modal closes. #410 just standardized AbortController on the other fetch effects, worth matching that here: pass `{ signal: controller.signal }` and return `() => controller.abort()` from the effect.

src/components/ScoreDiffModal/ScoreDiffModal.tsx
```diff
@@ -0,0 +1,586 @@
...
+    setDiffResults([])
+
+    axiosInstance
+      .get(
+        `/scores/diff?from=${fromDatacall.datacallid}&to=${toDatacall.datacallid}&fismasystemid=${fismasystemid}`
+      )
+      .then((res) => {
+        setDiffResults(res.data.data ?? [])
+      })
+      .catch((err) => {
```
danielbowne, line 236: On the auth-handled path this returns without setting `error`, so once loading clears you fall into the `diffResults.length === 0` branch and render "No changes between these two datacalls." That reads as a real empty result when the request actually failed. A 401 redirects anyway, but a 403 leaves the user looking at a false "no changes." Set an error or a dedicated flag here so the empty state does not show on a failure.

src/components/ScoreDiffModal/ScoreDiffModal.tsx
```diff
@@ -0,0 +1,586 @@
...
+                          py: 0.75,
+                          borderTop: '2px solid #b3c2e8',
+                        }}
+                      >
+                        {pillar.pillar}
+                      </TableCell>
+                    </TableRow>
+                    {entries.map((entry, i) => (
+                      <TableRow
+                        key={entry.functionid}
```
danielbowne: Rows are keyed by `entry.functionid`, but the diff is per question and a function can own more than one. Two changed entries with the same functionid collide and React drops one, silently hiding a real change. The fixtures use one question per function so tests pass. The map already has the index in scope, so `key={`${entry.functionid}-${i}`}` (or a composite with the question) fixes it.

danielbowne (commented):
Took a careful pass through this. The component is solid in a lot of places, the pillar ordering is deterministic, null sides render cleanly, getOptionDisabled blocks identical picks from the dropdown, and loading always clears. A few real things worth sorting before this merges though, with inline notes on the code ones.

Two things above the code level:

**Backend sequencing.** The test plan calls out that the diff table, empty state, null sides, and re-fetch are all unverified pending the backend `GET /scores/diff`, which is not in yet. As written, merging this to main puts the Compare Datacalls button in front of users on prod and every click lands on the error state until the backend ships. Worth gating the button behind a flag or holding the merge until the endpoint exists.

**Conflict with #408.** This reads `selectedDataCallId` off useContextProp and passes it down, but #408 replaces `selectedDataCallId` with the `selectedDatacall` object on context. Whichever of these two lands second will break or conflict here, so we should pick an order and plan the small rebase (`selectedDatacall?.datacallid`).

One cleanup: `docs/ScoreDiffModal_Documentation.docx` is a binary committed to the branch. It is not diff reviewable and bloats the repo. Better as a wiki page or on the PR description than checked in.

Couple of smaller but real ones, not blockers:
- When the viewed datacall is the oldest or the only one, From defaults to null and the results area renders completely blank with no message, since the empty state is gated on both pickers being set. A "select two datacalls" line would cover it.
- The datacalls and questions caches are module level with a 10 min TTL and never invalidate, so a datacall or score change elsewhere shows stale here for up to 10 minutes. Minor, just noting it.

Rest of the notes are inline.

costacalvin:
> **Backend sequencing.** The test plan calls out that the diff table, empty state, null sides, and re-fetch are all unverified pending the backend `GET /scores/diff`, which is not in yet. As written, merging this to main puts the Compare Datacalls button in front of users on prod and every click lands on the error state until the backend ships. Worth gating the button behind a flag or holding the merge until the endpoint exists.

Yep this PR should wait to merge until the backend `GET scores/diff` is in main/prod.

> **Conflict with #408.** This reads `selectedDataCallId` off useContextProp and passes it down, but #408 replaces `selectedDataCallId` with the `selectedDatacall` object on context. Whichever of these two lands second will break or conflict here, so we should pick an order and plan the small rebase (`selectedDatacall?.datacallid`).

#408 should be merged first then a small rebase/conflict resolution with (`selectedDatacall?.datacallid`) replacing `selectedDataCallId`.

> One cleanup: `docs/ScoreDiffModal_Documentation.docx` is a binary committed to the branch. It is not diff reviewable and bloats the repo. Better as a wiki page or on the PR description than checked in.

We can get rid of this file.

> * When the viewed datacall is the oldest or the only one, From defaults to null and the results area renders completely blank with no message, since the empty state is gated on both pickers being set. A "select two datacalls" line would cover it.

Agree!

I see the other comments as well, with this many changes do you think me creating a new PR is best practice here? @danielbowne Thanks for the review!

danielbowne:
@costacalvin we are going to merge  #408 → #409 → #410 then see where we land for this one.  Don't worry about another PR for now as mack will have to rehome it and 2 more reviews.

MackOverflow:
Thanks for the thorough pass, @danielbowne! Addressed the three inline code issues in ec6d0f4:

**AbortController on diff fetch (line 216)**
Added `AbortController` to the diff `useEffect` — `{ signal: controller.signal }` is passed to the `get` call and `() => controller.abort()` is returned from the effect. The `.catch` and `.finally` handlers also guard on `controller.signal.aborted` so state isn't written after the modal unmounts or pickers change mid-flight. Matches the pattern from #410.

**isAuthHandled silent empty-state (line 231)**
The early `return` after `isAuthHandled` now sets `error` before returning, so a 403 surfaces the "Failed to load diff" alert rather than falling through to the empty-state message.

**Duplicate React key (line 533)**
Switched from `key={entry.functionid}` to `key={\`${entry.functionid}-${i}\`}` so two changed entries under the same function don't collide and silently drop one.

---

On the two above-the-code items:

- **Backend sequencing / button gating** — Noted. That's a call for the team on whether to flag-gate the button or hold the merge. Flagging here so it stays visible for the merge decision.
- **Conflict with #408 (`selectedDatacall` rename)** — Also noted. Whichever lands second will need the one-line rebase (`selectedDatacall?.datacallid`). Happy to do that rebase once the ordering is decided.
- **`docs/ScoreDiffModal_Documentation.docx`** — Agreed it shouldn't live in the repo. Will move to the wiki or PR description in a follow-up.

MackOverflow:
## Status recap — 2026-06-22

### Review fixes (Danny's inline comments, commit ec6d0f4)

Three code issues from @danielbowne's review were addressed:

- **AbortController on diff fetch** — added `AbortController` to the diff `useEffect`, passes `{ signal }` to axios, guards `.catch`/`.finally` against writing state after abort or picker change. Matches the pattern from #410.
- **isAuthHandled silent empty-state** — the early-return path on `isAuthHandled` now sets `error` first, so a 403 surfaces the failure alert instead of silently rendering "No changes."
- **Duplicate React key** — `key={entry.functionid}` → `key={\`${entry.functionid}-${i}\`}` so two changed entries under the same function don't silently drop one.

Danny's above-the-code items (backend sequencing, `docs/` .docx, #408 rename) were acknowledged in a reply comment.

### Conflict check and rebase (commits fec1176 → 8d44e38)

Checked open PRs for conflicts. **#408** (Calvin's datacall context shape change) had already merged to main, causing a direct conflict in `QuestionnairePage.tsx` — the `useContextProp()` destructure referenced `selectedDataCallId` (number) and `selectedDatacall` (string), both of which #408 replaced with `selectedDatacall` (full `datacall` object).

Rebased onto current main. Conflict resolved by merging the destructures:

```ts
const {
  userInfo,
  selectedDatacall,     // now datacall | null (from #408)
  latestDataCallId,
  latestDatacall,       // from #408
  latestDeadline,       // from #408
  fismaSystems,         // added by this PR
} = useContextProp()
```

`ScoreDiffModal` prop updated to derive the "To" picker seed from the object:

```ts
selectedDataCallId={selectedDatacall?.datacallid}
```

A comment at that call site references `#417`/`#408` so the rename context is visible to future readers.

**#409** (async/await cleanup) also merged — no conflict with this PR.  
**#410** (AbortController refactor) still open — no conflict expected.  
**#418** (Onix's 403 fix) — touches different parts of `Title.tsx` only, no conflict.

### Still pending before merge

- Backend `GET /api/v1/scores/diff` endpoint — the Compare Datacalls button is live; every click hits the error state until this ships. Team should decide on feature-flag gating.
- `docs/ScoreDiffModal_Documentation.docx` — binary file committed to branch, should move to the wiki.
- #417 — open with the same branch name as #408; worth confirming whether it has additional commits that haven't landed yet.

costacalvin:
> * [Enhancement/indicate current datacall #417](https://github.com/CMS-Enterprise/ztmf-ui/pull/417) — open with the same branch name as [indicate most-recent and active vs closed status datacalls #408](https://github.com/CMS-Enterprise/ztmf-ui/pull/408); worth confirming whether it has additional commits that haven't landed yet.

Closing #417, was a fork to try and address feedback but you got those in for me. Thank you Mack!

danielbowne (commented):
Took another pass after the updates. The two that mattered are fixed:

- Stale-response race: the diff fetch now creates an AbortController, passes the signal, and aborts the previous request on cleanup, so a slow earlier response can no longer overwrite a newer one. The aborted guard in the catch and the abort-aware finally are both there.
- Row key: now `${entry.functionid}-${i}`, so multi-question functions no longer collide and drop rows.

Also good: the auth/error path now sets an error message instead of falling through to the "No changes" empty state, so a failed load no longer reads as a real empty result.

Two small leftovers, neither blocking:
- docs/ScoreDiffModal_Documentation.docx is still committed. Still think that belongs in the wiki or the PR description rather than the repo, but your call.
- When the viewed datacall is the oldest or the only one, From still defaults to null and the results area renders blank with no message. A "select two datacalls to compare" line would close that gap.

One thing to confirm before this merges, not a code issue: the backend GET /scores/diff. If that endpoint is not live yet, the Compare button ships to prod and errors on every click, so it is worth gating the button or holding the merge until the endpoint lands. If it is already in, disregard.

Looks good overall.

MackOverflow:
> One thing to confirm before this merges: the backend GET /scores/diff. If that endpoint is not live yet, the Compare button ships to prod and errors on every click.

Confirmed — the backend endpoint is **not live yet**. It's implemented in [CMS-Enterprise/ztmf#356](https://github.com/CMS-Enterprise/ztmf/pull/356) (in-repo rehome of #355, authored by @voidspooks), which is still open.

Holding merge on this PR until #356 lands. Once the backend ships, this is ready to go — no other blockers on the frontend side.

MackOverflow:
> `docs/ScoreDiffModal_Documentation.docx` is still committed. Still think that belongs in the wiki or the PR description rather than the repo, but your call.

Agreed — removed the `.docx` and replaced it with [`docs/ScoreDiffModal.md`](https://github.com/CMS-Enterprise/ztmf-ui/blob/feat/411-compare-datacalls-modal/docs/ScoreDiffModal.md). Same content, fully diffable, renders natively on GitHub, and stays in sync with the code since it lives next to it.

a-dipietro (approved):
Provided this was appropriately smoke tested, lgtm! Nice work on this one.

jsos3-cms:

jsos3-cms:
Reopened on the project board to backfill the Estimate (story points) for data attribution/reporting, then re-archived. No code, review, or status change.

Issue #403, opened 2026-06-17
Unprovisioned (IdP-authed) users get a misleading 'session expired' loop instead of a 'no account' message


danielbowne:
### Description

A user who authenticates successfully through IDM/Okta but has no corresponding row in the ZTMF `users` table is shown a misleading "Your session has expired. Please log in again." message and is put into an infinite loop. The IdP session is valid, so the ALB never re-prompts for login; the SPA calls `GET /api/v1/users/current`, the backend returns 401 because the authenticated identity resolves to no app account, and the frontend renders the session-expired screen. Clicking Sign in redirects to `/login`, the ALB still has a valid OIDC session, it bounces straight back, `/users/current` 401s again, and the screen returns. There is no signal that the real problem is "your account is not set up in ZTMF," and no path forward.

This is distinct from a genuinely expired session and from a 403 action-denied (tracked separately in #156 / #390). This is the never-provisioned (or stored-email-mismatch) case.

Confirmed in prod: backend logs show `Could not find user for request: no data when expected` (108 occurrences over ~5h while only a couple users were affected, i.e. each looping), with zero "deleted user" hits, so the 401 is a user-not-found on a valid IdP identity.

### Steps to Reproduce

1. Grant a user the ZTMF tile/group in IDM/Okta, but do not add them to the ZTMF `users` table.
2. Have them open the hosted app and authenticate through IDM.
3. They land directly on "Your session has expired. Please log in again." (no login prompt, since the IdP session is valid).
4. Click Sign in. A screen flashes briefly, then returns to the same message. Repeats indefinitely.

### Expected Behavior

When `/users/current` returns 401/403 for someone who just completed IdP auth and has no prior app session, show a clear, terminal message such as "Your account isn't set up in ZTMF yet. Contact your administrator to request access," with no retry loop. Do not present it as an expired session.

### Actual Behavior

Generic "Your session has expired. Please log in again." with a Sign in button that loops back to the same screen, because the IdP session is valid and the user can never satisfy `/users/current`.

### Environment

- Hosted: reproduced in prod (ztmf.cms.gov); applies to dev as well.
- Browsers: Chrome and Edge.
- Machines: CMS VDI and contractor laptop. Browser- and machine-independent, which (together with the backend logs) confirms it is account/server-side, not cache or browser.

### Acceptance Criteria

- [ ] A user who authenticates at the IdP but has no ZTMF account sees a distinct, accurate message (not "session expired") and is not looped.
[...]

PR #360, opened 2026-05-20
Align frontend scoring with HHS scale (1.0-5.0) and read tier from API

danielbowne:
## Summary

Frontend pair for `CMS-Enterprise/ztmf#318` (branch `feature/hhs-scoring-alignment`). Tracked under `CMS-Enterprise/ztmf-misc#175`. The backend authoritative plan is at `~/projects/prod/ztmf/tmp/hhs-scoring/scoring-alignment-plan.md`.

Both PRs auto-deploy to dev for joint testing, then merge to main within a short window of each other to keep the prod cutover gap small.

## What the backend now sends

`/api/v1/scores/aggregate` and `/api/v1/scores/aggregate?include_pillars=true` keep their JSON shape, with two additions and one semantic change:

- `data[].systemscore` is on the HHS 1.0 to 5.0 scale (was 1.0 to 4.0)
- `data[].systemtier` is a new string field
- `data[].pillarscores[].score` is on the same HHS 1.0 to 5.0 scale
- `data[].pillarscores[].tier` is a new string field on each pillar entry

Tier values: `Optimal`, `Advanced`, `Initial`, `Traditional`, `Not Assessed`. Boundaries: Optimal at 4.1+, Advanced at 3.1+, Initial at 2.1+, Traditional at 1.01+, Not Assessed below 1.01. The string is authoritative; the frontend no longer derives tier from the numeric score.

### Centralized types and shared style map

- `src/types.ts`: new `ScoreTier` union, `PillarScore`, `ScoreAggregate`, and `SystemScoreEntry` types. `FismaTableProps.scores` updated from `Record<number, number>` to `Record<number, SystemScoreEntry>`. The two tier fields (`systemtier` and `pillarscores[].tier`) are optional so a transient mismatch between deployed backend and frontend during the cutover does not throw a runtime TypeError; tier chips simply hide until the backend ships tier strings.
- `src/utils/tierStyles.ts` (new): single `TIER_STYLES` map of `{ color, backgroundColor }` keyed by tier. Color tokens lifted from the prior `getMaturityLevel` ladder so chips render identically once data flips. Exports a `styleForTier` helper for safe undefined handling.
- `src/utils/tierStyles.test.ts` (new): regression contract pinning all five tier keys, validating hex tokens, and covering `styleForTier`. Failing tests on this file are the signal that the tier enum drifted.

### Pillar Scores modal

[...]

PR #418, opened 2026-06-22
Handle account not provisioned errors and gate branded header

tarratsco:
## Summary

Closes the FE half of #403.

Pre-existing bug: an IdP-authenticated identity with no ZTMF account (or a soft-deleted one) landed on LoginPage with "Your session has expired" and an infinite Sign in loop. The IdP session was valid so the ALB never re-prompted; the SPA 401'd on `/users/current` and rendered the wrong copy.

This PR adds a `NO_ACCOUNT` sign-in reason driven by a `403 + ACCOUNT_NOT_PROVISIONED` signal (delivered by the paired BE PR), renders a terminal "contact your administrator" state with no retry CTA, and preserves the existing `EXPIRED` handling for genuine 401s.

## Backend dependency

Pairs with the BE middleware change: CMS-Enterprise/ztmf#360. The `NO_ACCOUNT` terminal state only renders once the BE starts emitting the new `code` field.

**Deploy order: FE first, BE second.** This PR is forward-compatible: with old BE deployed, every new code path stays dormant and behavior matches today. The reverse order (BE first) is worse than the current bug: unprovisioned users see a confusing toast on a stuck page instead of the familiar (if misleading) loop. Coordinated single-deploy is best.

## Changes

* **`utils/authCodes.ts`** (new) — mirrors the BE auth-package constants (`UNAUTHORIZED`, `ACCOUNT_NOT_PROVISIONED`, `FORBIDDEN_ORIGIN`) plus FE-side `SignInReasons` (`NO_ACCOUNT`, `EXPIRED`). Single source of truth on the FE side.
* **`utils/authInterceptor.ts`** — branches on `response.data.code`. 403 with `ACCOUNT_NOT_PROVISIONED` redirects to `/signin` with `reason=NO_ACCOUNT` and suppresses the OOS toast. `FORBIDDEN_ORIGIN` logs and shows a defensive toast. 401 carries `reason=EXPIRED` on its existing redirect. Controller-level 403 (no code) path unchanged.
* **`router/authLoader.ts`** — discriminated return: `NO_ACCOUNT` and `EXPIRED` for the two auth-failure shapes. `NO_ACCOUNT` surfaces the BE `data.error` body as `message` so LoginPage can render it verbatim (the BE distinguishes unprovisioned vs soft-deleted via different message text under the same code). `EXPIRED` stays silent because the loader can't tell an expired session from a first-time visit. `serverError` path unchanged.
* **`views/LoginPage/LoginPage.tsx`** — `NoAccountTerminal` component renders the BE message verbatim with no retry affordance on `NO_ACCOUNT`. Otherwise the existing legacy / IdP-lookup variants render. Hedged placeholder copy from the earlier Path A commit removed.
* **`utils/apiErrors.ts`** — `parseApiError` surfaces the `code` field for callers that need to branch.
* **`views/Title/Title.tsx`** — useEffect gate, swap local `PromiseType` for the shared `AuthLoaderData`, and a header-gate follow-up (see below).

### Header gate fix (incidental, #399 follow-up)

The branded header now hides whenever LoginPage is the body (`loaderData.status !== 200`), not only on the `/signin` URL. Matches the datacall sub-bar's existing gate. Cold-load LoginPage at `/` was rendering under the branded chrome, which contradicted the "please sign in" prompt. Caught while smoke-testing #403; bundled here because Title.tsx was already in the diff.

## Test plan

Local-dev verification — no extra tooling required, only Vite dev server + DevTools console.

### Interceptor message rendering (DevTools console)

These exercise the exact same code path the interceptor uses in production: `router.navigate('/signin', {state})`. Paste in the console (Vite dev mode serves the router module via dynamic import):

```js
const router = (await import('/src/router/router.tsx')).default

// EXPIRED — what the interceptor sends on 401
router.navigate('/signin', { replace: true, state: {
  message: 'Your session has expired. Please log in again.',
  reason: 'EXPIRED',
}})

// NO_ACCOUNT — unprovisioned user
router.navigate('/signin', { replace: true, state: {
  message: 'Your ZTMF account is not set up. Contact your administrator to request access.',
  reason: 'NO_ACCOUNT',
}})

// Soft-deleted variant (same reason, different BE copy)
router.navigate('/signin', { replace: true, state: {
  message: 'Your ZTMF account is no longer active. Contact your administrator.',
  reason: 'NO_ACCOUNT',
}})
```

* [ ] EXPIRED — URL becomes /#/signin; "Your session has expired. Please log in again." renders in red above the email form.
* [ ] NO_ACCOUNT — URL becomes /#/signin; terminal state renders the BE message verbatim with no retry CTA, no email form, no branded header.
* [ ] Soft-deleted variant — same terminal state, different BE copy ("no longer active") → confirms FE renders BE message verbatim rather than a hardcoded FE string.

### Real-scenario checks (no mocking, no console)

* [ ] Cold-load EXPIRED — with no bearer token configured in env → reload → email form renders with NO "session expired" copy above it. Intentional: loader path stays silent because it can't distinguish an expired session from a first-time visit.
* [ ] Active-session redirect — with a valid bearer token, manually visit /#/signin → bounces to /.
* [ ] 5xx — stop the BE container (docker compose stop api or equivalent) → reload → ServerErrorPage renders.
* [ ] Header gate — with no bearer token, reload → branded ZTMF header is hidden above the LoginPage email form.

### Unit tests

* authInterceptor.test.ts - all code branches (UNAUTHORIZED, ACCOUNT_NOT_PROVISIONED, FORBIDDEN_ORIGIN, controller-level 403, skipAuthHandling bypass, 5xx pass-through, network error pass-through).
* LoginPage.test.tsx (new) - active-session redirect, NO_ACCOUNT terminal state from location.state and from the loader, fallback message when none provided, EXPIRED state for both Legacy and IDP_ENABLED variants, regression check that cold-load EXPIRED shows no session-expired copy.
* EmailModal.test.tsx - updated assertion for the new 401 state shape.

make check and make test green (163 passing).

## Notes for reviewers

Loader path stays silent on 401; interceptor path carries the message. The loader (cold-load 401) can't distinguish "expired session" from "first-time visit," so it doesn't pass a message — the BE's expiry-specific copy ("Your session has expired. Please sign in again.") would lie to a first-time visitor. The interceptor (mid-session 401) has ground truth that a session existed, so it carries ERROR_MESSAGES.expired via location.state. Both surface via the same <SessionMessage> component in LoginPage when the message is present.

### DIFF
--- a/src/components/EmailModal/EmailModal.test.tsx
+++ b/src/components/EmailModal/EmailModal.test.tsx
@@ -77,7 +77,7 @@ test('success path fires the success snackbar and surfaces sent emails', async (
   expect(mockedNavigate).not.toHaveBeenCalled()
 })
 
-test('401 redirects to sign-in with the expired-session message', async () => {
+test('401 redirects to sign-in with the expired-session message and reason', async () => {
   mock.onPost('/massemails').reply(401)
 
   renderWithProviders(<EmailModal openModal={true} closeModal={jest.fn()} />)
@@ -86,7 +86,7 @@ test('401 redirects to sign-in with the expired-session message', async () => {
   await waitFor(() => {
     expect(mockedNavigate).toHaveBeenCalledWith(Routes.SIGNIN, {
       replace: true,
-      state: { message: ERROR_MESSAGES.expired },
+      state: { message: ERROR_MESSAGES.expired, reason: 'EXPIRED' },
     })
   })
   // No generic fallback snackbar fires on top of the redirect.
--- a/src/router/authLoader.ts
+++ b/src/router/authLoader.ts
@@ -1,6 +1,7 @@
 import 'core-js/stable/atob'
 import { userData, UserRole } from '@/types'
 import axiosInstance from '@/axiosConfig'
+import { AuthCodes, SignInReasons, type SignInReason } from '@/utils/authCodes'
 /**
  * Auth state loader for react-router data routes.
  * @module router/authLoader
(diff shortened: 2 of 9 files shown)


danielbowne:
Fork PR so it can't run our CI (no secrets). Rehomed all three of your commits onto an internal branch with authorship preserved, rebased on current main. Title.tsx auto-merged cleanly with the #425 datacall-allowlist change, both intact. Reviewed locally plus a second pass, forward-compat checks out, full suite green (183). Continuing in #428 with two small non-blocking follow-up notes for you in the description. Thanks @tarratsco.

PR #420, opened 2026-06-22
fix(system-detail): parse timestamp-format ATO dates

### LINKED ISSUE #419: ATO Expiration shows "Invalid Date" on system detail page
## Describe the bug

On the System Detail page, the **ATO Expiration** field renders "Invalid Date" for systems whose enrichment data stores the ATO date as a full timestamp (for example `2026-12-13 00:00:00.000`).

## Steps to reproduce

1. Open a system detail page for a system whose enrichment ATO date includes a time component.
2. Observe that **ATO Expiration** shows "Invalid Date" instead of a formatted date.

## Expected behavior

A formatted date, or an em dash placeholder when no ATO date is present.

## Root cause

In `src/views/SystemDetailPage/CfactsRecordCard.tsx`:

- `parseDateOnly()` splits the string on `-` and treats the third token as a bare day. For `2026-12-13 00:00:00.000` the third token is `13 00:00:00.000`, so `Number(...)` is `NaN` and the resulting `Date` is invalid.
- `parseDate()` only routes to `new Date()` when the string contains `T`. These values use a space separator, so they fall through to `parseDateOnly()`.
- `formatDate()` then returns the string `"Invalid Date"` (which is truthy), so the existing `|| '—'` fallback never fires.
- The same `parseDateOnly()` feeds `getAtoColor()`, so the expired/expiring ATO color highlighting is also silently broken for these values.

This is a parsing/error-handling bug, not bad data: the value is a valid timestamp. Systems with no ATO date present correctly show the placeholder.

## Proposed fix

- In `parseDateOnly()`, take only the date portion before a space or `T`: `dateStr.split(/[ T]/)[0].split('-')`.
- In `formatDate()`, guard `isNaN(date.getTime())` and return `null` so the placeholder shows.
- Add a unit test covering the `YYYY-MM-DD HH:MM:SS.mmm` timestamp format.

## Notes

Affects dev and prod (shared component), so the fix targets `main`. Related: the planned rename of `CfactsRecordCard` to `SystemEnrichmentCard` (tracked in the cross-cutting backlog).

danielbowne:
Closes #419.

ATO Expiration on the System Detail page rendered the literal "Invalid Date" for systems whose enrichment ATO date is stored as a full timestamp like `2026-12-13 00:00:00.000` instead of a bare `YYYY-MM-DD`. The value is valid data, the parsing was the problem.

`parseDateOnly` split the whole string on `-`, so the day token came back as `13 00:00:00.000` and the Date was invalid. `parseDate` only routed to `new Date()` when the string contained `T`, and these use a space separator, so they fell through. `formatDate` then returned the string "Invalid Date", which is truthy, so the existing `|| "—"` placeholder never kicked in. The same parse path feeds `getAtoColor`, so the expired/expiring color highlight was silently broken for these values too.

Changes in `src/views/SystemDetailPage/CfactsRecordCard.tsx`:
- `parseDateOnly` takes only the date portion before a space or `T` before splitting on `-`.
- `formatDate` guards `isNaN(date.getTime())` and returns null so the placeholder shows.

Added two tests to `CfactsRecordCard.test.tsx`: one asserting a timestamp-format ATO date renders a formatted date and not "Invalid Date", and one asserting a missing date still renders the placeholder.

Shared component, so this affects dev and prod and targets main.

Verify: `npx tsc --noEmit` clean, and `yarn test --watchAll=false src/views/SystemDetailPage/CfactsRecordCard.test.tsx` (6 passing).

### DIFF
--- a/src/views/SystemDetailPage/CfactsRecordCard.test.tsx
+++ b/src/views/SystemDetailPage/CfactsRecordCard.test.tsx
@@ -94,3 +94,40 @@ test('500 renders the failed-to-load message', async () => {
     await screen.findByText(/failed to load ztmf insights data/i)
   ).toBeInTheDocument()
 })
+
+test('formats a timestamp-format ATO expiration date instead of "Invalid Date"', async () => {
+  mock.onGet(`/systemenrichment/${FISMA_UID}`).reply(200, {
+    data: {
+      fisma_uuid: FISMA_UID,
+      payload: {
+        authorization_package_name: 'Test Package',
+        ato_expiration_date: '2026-12-13 00:00:00.000',
+      },
+      synced_at: '2026-01-01T00:00:00Z',
+    },
+  })
+
+  renderWithProviders(<CfactsRecordCard fismaUid={FISMA_UID} />)
+
+  expect(await screen.findByText('Test Package')).toBeInTheDocument()
+  expect(screen.getByText('12/13/2026')).toBeInTheDocument()
+  expect(screen.queryByText('Invalid Date')).not.toBeInTheDocument()
+})
+
+test('renders the placeholder when no ATO expiration date is present', async () => {
+  mock.onGet(`/systemenrichment/${FISMA_UID}`).reply(200, {
+    data: {
+      fisma_uuid: FISMA_UID,
+      payload: {
+        authorization_package_name: 'Test Package',
+        ato_expiration_date: null,
+      },
+      synced_at: '2026-01-01T00:00:00Z',
+    },
+  })
+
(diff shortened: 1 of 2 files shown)


MackOverflow (approved):
Looks good, one thing: 
 isPastDeadline fallback defaults to true when latestDeadline is missing (QuestionnairePage.tsx:296)

setIsPastDeadline(
  latestDeadline ? new Date() > new Date(latestDeadline) : true
)
Defaulting to true (read-only) on a missing deadline is defensively correct, but it could silently make the questionnaire appear read-only if latestDeadline is an empty string due to an unrelated bug. Consider logging a warning or surfacing an error state when latestDeadline is empty rather than silently locking the user out.

PR #425, opened 2026-06-23
fix(#421): render datacall selector only where needed

danielbowne:
## Summary

Fixes #421. Rehomed from #424 by @costacalvin (Calvin Costa) onto an internal branch so CI can run with secrets; original authorship is preserved on the commit. Rebased onto current `main` (picks up #423, which was merged after #424 was opened).

The datacall sub-bar previously rendered on every authenticated route because `Title.tsx` only excluded the sign-in page (`!isSignInRoute`), so it leaked onto admin, user management, and any future routes.

This flips to an allowlist: the sub-bar renders only on the three routes where datacall context is meaningful.

| Route | Datacall bar |
|---|---|
| `/` (dashboard) | Full selector |
| `/questionnaire/*` | Full selector |
| `/systems/:id` | Read-only text (unchanged) |
| Everything else | Hidden |

An allowlist is self-closing: new admin/settings routes get no datacall bar by default, so this does not need revisiting each time a route is added.

## Changes

- `src/views/Title/Title.tsx` — replace `!isSignInRoute` denylist with `isHomeRoute || isQuestionnaireRoute || isSystemDetail` allowlist

## Verification

- typecheck, lint, full test suite (172 passed) green locally on the rebased branch
- Route prefixes confirmed against `src/router/constants.ts` (`/`, `/questionnaire/:fismaacronym/...`, `/systems/:fismasystemid`)
- `isSignInRoute` still used for the header block, no dead code

Credit: @costacalvin. Supersedes #424.

PR #428, opened 2026-06-23
fix(#403): handle account-not-provisioned auth errors and gate branded header

danielbowne:
## Summary

Closes the FE half of #403. Rehomed from #418 by @tarratsco (Onix Tarrats Calderon) onto an internal branch so CI can run with secrets; original authorship is preserved across all three commits. Rebased onto current `main` (Title.tsx auto-merged cleanly with the #425 datacall-allowlist change; both survived).

Pre-existing bug: an IdP-authenticated identity with no ZTMF account (or a soft-deleted one) landed on LoginPage with "Your session has expired" and an infinite sign-in loop. The IdP session was valid so the ALB never re-prompted; the SPA 401'd on `/users/current` and rendered the wrong copy.

This adds a `NO_ACCOUNT` sign-in reason driven by a `403 + ACCOUNT_NOT_PROVISIONED` signal from the paired BE PR, renders a terminal "contact your administrator" state with no retry CTA, and preserves the existing `EXPIRED` handling for genuine 401s. Also gates the branded header to hide whenever LoginPage is the body, not just on the `/signin` URL.

## Backend dependency

Pairs with CMS-Enterprise/ztmf#360. The `NO_ACCOUNT` terminal state only renders once the BE emits the new `code` field.

**Deploy order: FE first, BE second.** Forward-compatible: with old BE deployed, every new path stays dormant and behavior matches today.

## Changes

- `utils/authCodes.ts` (new) — mirrors BE auth-package constants + FE `SignInReasons`
- `utils/authInterceptor.ts` — branches on `response.data.code` (NO_ACCOUNT, FORBIDDEN_ORIGIN, controller-403 unchanged)
- `router/authLoader.ts` — discriminated return for the two auth-failure shapes
- `views/LoginPage/LoginPage.tsx` — terminal `NoAccountTerminal`, BE message verbatim
- `utils/apiErrors.ts` — surface `code` for callers
- `views/Title/Title.tsx` — header gate + shared `AuthLoaderData` type

## Verification

- typecheck, lint, full suite green on rebased branch (183 passing; new LoginPage + authInterceptor suites)
- Forward-compat confirmed: old BE → `code` undefined → 401 stays EXPIRED, 403 falls through to existing controller-403 toast
- Title.tsx merge verified: #425 allowlist (datacall sub-bar) and the new header gate both intact

## Non-blocking follow-ups (from review, optional)

1. Interceptor's empty-body `NO_ACCOUNT` fallback uses the permission copy, while the loader path lets LoginPage own the fallback — divergent copy only if BE sends an empty `error` body (paired BE always sends one). Could pass `reason` only and let LoginPage own the fallback.
2. `AuthCodes.UNAUTHORIZED` is declared for BE-mirror docs but not branched on (any 401 is treated as EXPIRED). Harmless.

Credit: @tarratsco. Supersedes #418.

### DIFF
--- a/src/components/EmailModal/EmailModal.test.tsx
+++ b/src/components/EmailModal/EmailModal.test.tsx
@@ -77,7 +77,7 @@ test('success path fires the success snackbar and surfaces sent emails', async (
   expect(mockedNavigate).not.toHaveBeenCalled()
 })
 
-test('401 redirects to sign-in with the expired-session message', async () => {
+test('401 redirects to sign-in with the expired-session message and reason', async () => {
   mock.onPost('/massemails').reply(401)
 
   renderWithProviders(<EmailModal openModal={true} closeModal={jest.fn()} />)
@@ -86,7 +86,7 @@ test('401 redirects to sign-in with the expired-session message', async () => {
   await waitFor(() => {
     expect(mockedNavigate).toHaveBeenCalledWith(Routes.SIGNIN, {
       replace: true,
-      state: { message: ERROR_MESSAGES.expired },
+      state: { message: ERROR_MESSAGES.expired, reason: 'EXPIRED' },
     })
   })
   // No generic fallback snackbar fires on top of the redirect.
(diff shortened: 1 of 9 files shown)


a-dipietro (approved):
g2g

PR #429, opened 2026-06-24
Hide unrelated Pillars on SaaS environments.

### LINKED ISSUE #426: Filter Devices and Applications pillars from SaaS system questionnaire
## Feature Type

UI Component

## Workflow Description

When a user views the ZTMF questionnaire for a SaaS system type, only the applicable ZT pillars should be displayed. Specifically, the Devices and Applications pillars should be removed from both the question flow and the left-side navigation menu, as they do not apply to SaaS systems.

## Acceptance Criteria

- [ ] The Devices pillar is hidden from the questionnaire and left-side navigation for SaaS system types
- [ ] The Applications pillar is hidden from the questionnaire and left-side navigation for SaaS system types
- [ ] Remaining applicable pillars display correctly for SaaS systems
- [ ] PR must include a screenshot of the updated UI demonstrating the filtered view

## Additional Context

Child of #312. The HHS-developed reduced questionnaire for SaaS systems defines which pillars apply.

tarratsco:
## Summary

Closes #426.

For SaaS-typed systems, Devices and Applications pillars are scored at the provider rather than the consumer, so showing those pillars on the questionnaire asks the wrong audience. This PR hides them on the FE for any system whose `datacenterenvironment === 'SaaS'`. Every other environment renders unchanged.

## Changes

* **`src/constants.ts`** — new `SAAS_EXCLUDED_PILLARS = ['Devices', 'Applications'] as const` constant, doc-commented alongside the existing `PILLAR_ORDER`. Names match the API's `question.pillar.pillar` value exactly.
* **`src/utils/filterPillarsForSystem.ts`** (new) — `filterPillarsForSystem(pillars, datacenterenvironment)`. Returns a new array; pass-through when env is anything other than `'SaaS'` (including null/undefined/empty); preserves input order.
* **`src/utils/filterPillarsForSystem.test.ts`** (new) — 8 unit tests covering: null/undefined env pass-through, non-SaaS env pass-through, SaaS dropping the two excluded pillars, case sensitivity on `'SaaS'` (saas / SAAS do not trigger), order preservation, empty input, non-mutation of the input array.
* **`src/views/QuestionnairePage/QuestionnairePage.tsx`** — wraps the existing `sortPillars(...)` call with `filterPillarsForSystem(..., systemInfo?.datacenterenvironment)`. Adds the env string to the fetch effect's dep array so a system change re-runs the question fetch with the right filter.
* **`src/views/QuestionnareModal/QuestionnareModal.tsx`** — same wrap, using `system?.datacenterenvironment`. Existing dep array already lists `system` so no dep-array change needed.

### Unit tests

`yarn jest src/utils/filterPillarsForSystem` — 8/8 passing locally.

`make check` and `make test` both green (191 passing).

### Regression checks (non-SaaS systems unchanged)

* [ ] Open the questionnaire for any non-SaaS seeded system (e.g., Imperial-Fleet, On-Prem, etc.) → all six pillars render as before.
* [ ] Open the modal for the same system → all six pillars present.
* [ ] Sort order unchanged across both flows (Identity → Devices → Networks → Applications → Data → CrossCutting on non-SaaS).

## Notes for reviewers

* Case-sensitive match on 'SaaS'. Anything else (case variants, null, empty, other envs) passes through. Guards against a silent break if the BE ever ships a typo or normalizes casing; we'd see pillars come back instead of disappear.
* Two pillars listed in SAAS_EXCLUDED_PILLARS today; if a future scope adds more environments or excludes different pillars, the helper's pass-through-by-default shape keeps the surface centralized.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -57,6 +57,17 @@ export const CONFIRMATION_MESSAGE_QUESTION =
   'Your changes will not be saved! Are you sure you want to leave question without saving your changes?'
 
 /**
+ * Pillars hidden on the questionnaire for SaaS systems. Names must match the API's
+ * `question.pillar.pillar` value exactly.
+ */
+export const SAAS_EXCLUDED_PILLARS = ['Devices', 'Applications'] as const
+
+// If we wanted to filter out some pillars for other environments, we could add them here.
+// For now, only SaaS triggers any filtering.
+
+// ---------
+
+/**
  * The fixed order for pillars on the questionnaire, no matter what order the
  * API returns them in. Each name has to match the API's
  * `question.pillar.pillar` value exactly (same keys as {@link PILLAR_FUNCTION_MAP}).
--- a/src/utils/filterPillarsForSystem.test.ts
+++ b/src/utils/filterPillarsForSystem.test.ts
@@ -1,0 +1,69 @@
+import { filterPillarsForSystem } from './filterPillarsForSystem'
+
+const FULL_PILLARS = [
+  'Identity',
+  'Devices',
+  'Networks',
+  'Applications',
+  'Data',
+  'CrossCutting',
+]
+
+describe('filterPillarsForSystem', () => {
+  it('passes through when datacenterenvironment is null', () => {
+    expect(filterPillarsForSystem(FULL_PILLARS, null)).toEqual(FULL_PILLARS)
+  })
+
+  it('passes through when datacenterenvironment is undefined', () => {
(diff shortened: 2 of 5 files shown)


tarratsco:
SaaS vs Non-SaaS system questionnaire pillar display:
https://github.com/user-attachments/assets/d259e09d-261a-4ef7-98eb-f678dccf6fdd

a-dipietro:
Rehomed to #430 .

Issue #366, opened 2026-05-27
Issues with list of systems in the User Assignment Picker


eschwein-usds:
### Description

There are systems that are not decommissioned, but appear in the user assignment picker that are not in the main list.  At least some of them need to be decommissioned.

### Steps to Reproduce

2. Click on '...'
3. Scroll down to '...'

### Expected Behavior

brb

### Actual Behavior

brb

### Environment

all

### Acceptance Criteria

- [ ] Bug no longer occurs when following steps above
- [ ] No regression in related functionality
- [ ] Tests added/updated to prevent recurrence

### Severity

None

### Logs / Error Messages

```shell

```

eschwein-usds:
NEver mind, this was user error -- the system appeared at the end of the list in the System Dashboard.

PR #435, opened 2026-06-26
Bugfix/add guard against admins deleting themselves

### LINKED ISSUE #433: Admins can delete their own account
## Recommended labels

`bug`, `area/frontend`, `lang/typescript`, `cross-repo`

## Summary

The UserTable's delete affordance is rendered for every row including the current user's own row. A logged-in admin can click the Delete icon on their own row, confirm the dialog, and the FE issues `DELETE /api/v1/users/<their-own-id>` which the backend accepts. After the call returns, the next API request 401s (the BE no longer recognizes the deleted user), the centralized interceptor redirects to `/signin` with "session expired," and the user is locked out — unable to log back in because their account is gone.

Self-deletion is destructive and irreversible from the user's own perspective (they'd need another admin to restore them via `PUT /users/<id>/restore`). It should be impossible from the UI, and the BE should reject the request as defense-in-depth.

## Root cause

The delete handler in `src/views/UserTable/UserTable.tsx`:

- `handleDeleteClick` (line 348) accepts any row id with no identity check.
- `handleConfirmDelete` (line 353) issues `axiosInstance.delete('/users/${target.userid}')` unconditionally.
- The columns definition renders the Delete icon for every row whose role grants the current admin delete rights, with no per-row "is this me?" guard.

The BE (`DELETE /api/v1/users/<id>`) accepts the request and performs the soft-delete without checking whether the caller is deleting themselves.

### FE (this ticket)

Compare `target.userid` against the current user's id from context. Two layered defenses, both inside `UserTable.tsx`:

1. **Hide the Delete icon on the current user's own row.** In the columns/`getActions` block that renders the row-level action icons, gate the Delete icon on `row.userid !== userInfo.userid`. Replace with a no-op `null` (or a disabled icon with a tooltip "You can't delete your own account") so the affordance is visibly absent.
2. **Guard `handleConfirmDelete` as a backstop.** Even with the icon hidden, if the handler is invoked with the current user's id (programmatic API misuse, future refactor mistake), short-circuit before the axios call:
   ```ts
   if (target.userid === userInfo.userid) {
     notify("You can't delete your own account.", 'error')
     return
   }
   ```

Source of truth for the current user: `userInfo.userid` from the outlet context (already available in UserTable via `useOutletContext` per the existing role-check code paths).

### BE (companion ticket, file separately)

Reject `DELETE /api/v1/users/<id>` when the path id equals the authenticated caller's id. Return 400 or 403 with a clear error message ("You cannot delete your own account") so the FE can surface it on the off-chance the FE guard is bypassed (older client, direct API call).

## Out of scope

- **Restore flow self-guarding.** A deleted user can't restore themselves (they're locked out), so the equivalent guard on `handleConfirmRestore` isn't a user-facing issue. Worth adding for consistency but not blocking.
- **Audit logging of attempted self-deletes.** If the BE side ends up rejecting the request, logging the attempt would be defensive-security nice-to-have; out of scope here.
- **Last-admin guard.** A different but related concern: even deleting a *different* admin can leave the system with zero admins. That's a separate ticket — it requires counting admins in scope, which is orthogonal to the self-delete fix.
- **UI for the "request access" or "recover deleted account" path** that a deleted user could see when they're locked out. The fix here prevents the situation; recovering from it (if it happens) is a separate ticket.

### Functional

1. Log in as an admin user (call them `A`).
2. Visit the Users table.
3. Find `A`'s own row → confirm no Delete icon (or a disabled one with tooltip).
4. Find another admin's row (call them `B`) → confirm Delete icon present and works as today.
5. Open DevTools console, attempt to programmatically trigger the delete handler against `A`'s userid:
   ```js
   // get the handler off the UserTable instance via React DevTools, or
   // attempt: fetch('/api/v1/users/<A-id>', {method: 'DELETE', credentials: 'include'})
   ```
   - With the FE guard in place: the in-component handler should notify "You can't delete your own account" and not issue the request.
   - With BOTH FE and BE guards in place: even a direct fetch should be rejected by the BE with 400/403.

### Regression

- Deleting another user still works as today.
- Restore flow unchanged.
- The Users table renders normally for non-admin (read-only) viewers — no delete icons at all on any row, consistent with prior behavior.

### Edge cases

- A deleted user attempting to delete themselves via API while their session is still valid (very short window between BE delete and next API call) — covered by the BE-side guard once that companion ticket lands.
- An OWNER deleting an OWNER (themselves vs. another) — guard should fire only on the self-delete case; deleting other OWNERs continues to be allowed (or restricted separately if product wants OWNER protection).

## Notes

- **Two-layer fix is intentional.** Hiding the icon is the right primary defense (no affordance → no accidental click). Guarding `handleConfirmDelete` is a backstop against future refactors that might wire a different code path to the handler (e.g., a keyboard shortcut, batch-delete action, etc.) and accidentally bypass the column-level visibility check.
- **Cross-repo dependency**: the BE half of this fix is filed (or to be filed) in `ZTMF-backend-issues/` as a defense-in-depth ticket. Order doesn't matter — FE-first ships the visible fix; BE-first hardens the API.
- Related: `[[37 - Sign-out affordance with backend logout endpoint]]` (another auth-state hardening ticket).
- Surfaced during ad-hoc smoke testing.

## Steps to reproduce

1. Log in as an admin (ADMIN or OWNER role).
2. Navigate to the Users table (`/users`).
3. Find your own row.
4. Click the Delete icon on your row.
5. Confirm the dialog.

### Expected Behavior

**Expected**: the Delete icon either isn't rendered on your own row, or is disabled with a tooltip explaining why.

### Actual Behavior

**Actual**: the delete succeeds. The row disappears from the table. The next API call (e.g., reload, or any save) 401s. The interceptor redirects to `/signin` with "Your session has expired." A "Sign in" click takes the user through the IdP, but the BE rejects them (no `users` row → either 401 today, or 403 `ACCOUNT_NOT_PROVISIONED` once `#403` ships). Either way, the user can't get back in without another admin running `PUT /users/<id>/restore`.

### Environment

brave (chromium based browser)

## Acceptance criteria

- [ ] The Delete icon does not render on the current user's own row in the Users table, or is rendered disabled with an explanatory tooltip.
- [ ] `handleConfirmDelete` short-circuits with a user-facing notify if invoked against the current user's id.
- [ ] All other rows continue to behave as before — admins can still delete other users normally.
- [ ] `make check` and `make test` pass; new unit/integration test covers both the icon-hidden and the handler-guard paths.

### Severity

High - Significantly impacts usage

### Affected Component

_No response_

### Screenshots

https://github.com/user-attachments/assets/5d7a38db-b61a-49a6-b776-eb3e29f62833

### Logs / Error Messages

```shell

```

### Workaround

_No response_

a-dipietro:
In-repo version of #434 by @tarratsco, moved so the org-secret CI gates (Snyk,
Infrastructure/Deploy) can run — they can't on a fork PR. Commits pushed unchanged;
authorship preserved.

Closes #433.

## Summary
Two-layer defense against an admin self-deleting:

1. Visible: the Delete row-action icon on the current user's own row is disabled, with a
   tooltip explaining why ("You can't delete your own account"). Icon color shifts black →
   gray to reinforce the disabled state. Other rows render the Delete icon as today with a
   new "Delete User" tooltip for consistency with the other row actions.
2. Backstop: `handleConfirmDelete` short-circuits with a notify toast if invoked against the
   current user's id. This catches programmatic invocation or future refactors that wire a
   new entry point to the handler.

## Why two layers
Self-delete locks the admin out of the app with no FE-recoverable path (no other admin
nearby = locked out until a sibling admin runs `PUT /users/<id>/restore`). The visible
affordance prevents the accidental click; the backstop catches everything else.

## Test plan
Manual verification (UserTable has no trivial equality checks; not worth standing up MUI
DataGrid + outlet context + axios mocking just for this PR):
- Log in as admin, visit `/users`, hover own Delete icon → tooltip reads "You can't delete
  your own account", icon is greyed.
- Hover another user's Delete icon → tooltip reads "Delete User", icon is black, click →
  delete modal opens normally.
- Trigger `handleConfirmDelete` programmatically against own user id via the console →
  notify toast fires, network call is blocked.

## Follow-ups (out of scope)
- BE companion: reject `DELETE` where target id === caller id (defense in depth) — ztmf#366.
- Restore self-guarding and last-admin protections, per the #433 out-of-scope section.

Refs #434

### DIFF
--- a/src/views/UserTable/UserTable.tsx
+++ b/src/views/UserTable/UserTable.tsx
@@ -354,6 +354,14 @@ export default function UserTable() {
     const target = pendingDeleteRow
     setPendingDeleteRow(null)
     if (!confirm || !target) return
+    // Backstop: the row-action icon for the current user is already
+    // disabled, but guard the handler in case it's invoked some other
+    // way (programmatic call, future refactor wiring a new entry point).
+    // Self-delete locks the user out of the app with no recovery path.
+    if (target.userid === userInfo.userid) {
+      notify("You can't delete your own account.", 'error')
+      return
+    }
     try {
       await axiosInstance.delete(`/users/${target.userid}`)
       setRows((prev) => prev.filter((row) => row.userid !== target.userid))
@@ -677,13 +685,37 @@ export default function UserTable() {
               color="inherit"
             />
           </Tooltip>,
-          <GridActionsCellItem
-            key={`delete-${params.id}`}
-            icon={<DeleteIcon sx={{ color: 'black' }} />}
-            label="Delete"
-            onClick={handleDeleteClick(params.id)}
-            color="inherit"
-          />,
+          <Tooltip
+            title={
+              params.row.userid === userInfo.userid
+                ? "You can't delete your own account"
+                : 'Delete User'
+            }
+            key={`tooltip-delete-${params.id}`}
+            placement="right-start"
+          >
+            {/* span wrapper lets Tooltip listen to events even when the
+                child is disabled (MUI requirement). */}
+            <span>
+              <GridActionsCellItem
+                key={`delete-${params.id}`}
+                icon={
+                  <DeleteIcon
+                    sx={{
+                      color:
+                        params.row.userid === userInfo.userid
+                          ? 'gray'
+                          : 'black',
+                    }}
+                  />
+                }
+                label="Delete"
+                onClick={handleDeleteClick(params.id)}
+                color="inherit"
+                disabled={params.row.userid === userInfo.userid}
+              />
+            </span>
+          </Tooltip>,
         ]
       },
     },
--- a/src/views/UserTable/UserTable.tsx
+++ b/src/views/UserTable/UserTable.tsx
@@ -650,6 +650,8 @@ export default function UserTable() {
           ]
         }
 
+        const isSelf = params.row.userid === userInfo.userid
+
         return [
           <GridActionsCellItem
             icon={<EditIcon />}
@@ -686,11 +688,7 @@ export default function UserTable() {
             />
           </Tooltip>,
           <Tooltip
-            title={
-              params.row.userid === userInfo.userid
-                ? "You can't delete your own account"
-                : 'Delete User'
-            }
+            title={isSelf ? "You can't delete your own account" : 'Delete User'}
             key={`tooltip-delete-${params.id}`}
             placement="right-start"
           >
@@ -699,20 +697,11 @@ export default function UserTable() {
             <span>
               <GridActionsCellItem
                 key={`delete-${params.id}`}
-                icon={
-                  <DeleteIcon
-                    sx={{
-                      color:
-                        params.row.userid === userInfo.userid
-                          ? 'gray'
-                          : 'black',
-                    }}
-                  />
-                }
+                icon={<DeleteIcon sx={{ color: isSelf ? 'gray' : 'black' }} />}
                 label="Delete"
                 onClick={handleDeleteClick(params.id)}
                 color="inherit"
-                disabled={params.row.userid === userInfo.userid}
+                disabled={isSelf}
               />
             </span>
           </Tooltip>,


jsos3-cms (approved):
LGTM! The only thing I would want to potentially add as a catch would be if any attempts at deletion are logged in the admin event panel; however completely deleting this functionality for the logged in user is effectively the same. Likely useful in other areas (logging attempted destructive actions, even if impossible to do), but not a blocker for this due to implementation.

a-dipietro:
Passed feedback to Onix for review/comment.

tarratsco:
@jsos3-cms Sorry for the noobness, do we have an admin event panel? I think we probably are fine from a FE perspective, however I added an equivalent ticket for a backend check: https://github.com/CMS-Enterprise/ztmf/issues/366

I think we can add proper logging for destructive action attempts there.

tarratsco:

PR #438, opened 2026-06-27
Enhancement/batch opdiv grants

jsos3-cms:
In-repo version of #437 by @costacalvin. Carries the full superset of #432 (`make idp and Opdiv fields editable`, commit `e709c76`) plus a second commit, `batch Opdiv grants` (`389230f`), so the org-secret-gated CI can run — Snyk and the dev deploy (via the AWS OIDC role) do not run on a fork PR. Both commits are unchanged and authorship is preserved (`Calvin Costa`); this repo is squash-merge-only, so please keep @costacalvin credited as the author on merge.

This PR currently carries #432's commit underneath the batch change. Per the author's note, #432 merges first; once it lands, this branch will be rebased down to the batch-only commit. Closes #395. Refs #406, #432, #437.

## Summary

The batch commit replaces the per-OpDiv POST/DELETE loop in `OpDivGrantModal` with a single `PUT /users/{userid}/opdivs` batch request. The backend reconciles the full desired set against current grants in one transaction and re-derives `identity_provider` once.

- **`userOpdivs.ts`** — removes `grantOpDiv`/`revokeOpDiv`; adds `setUserOpDivs(userid, opdivIds[])` → `PUT /users/{userid}/opdivs` with body `{ opdiv_ids: [...] }`.
- **`OpDivGrantModal`** — replaces immediate per-toggle writes with local state + Save/Cancel; the modal closes only on success. An OPDIV_ADMIN's scope is enforced by filtering the PUT body to the OpDivs they hold, which clears the backend's scope gate without revoking out-of-scope grants the target user holds from another admin (the backend only reconciles within the caller's scope).
- **`UserTable`** — wires `setUserOpDivs` into the new-user create flow (one PUT in place of the per-item loop) and stops a redundant refresh when the modal is cancelled.

The first commit (OpDiv/IdP fields editable at create) is the change previously carried by #432; see #432 / #431 for its screenshots and the "OpDiv grant derivation overrides explicit IdP" note.

## Depends on

- Backend CMS-Enterprise/ztmf#352 — widens the `identity_provider` override gate so HHS-wide admins can set Entra at create. Green in dev.
- Backend CMS-Enterprise/ztmf#369 — adds the `PUT /users/{userid}/opdivs` batch endpoint and its scope-aware reconciliation. Must be deployed to dev alongside this UI; until it is, the modal Save and new-user-with-OpDivs paths return 404 in dev.

## Test plan

[...]

PR #444, opened 2026-07-01
fix(login): distinguish lookup outages from unknown-email; enable withCredentials

jsos3-cms:
## Summary

Hardens how the pre-auth login page handles the email→IdP lookup. Today a transport or server failure (timeout, network error, 4xx/5xx, or a malformed response) is folded into the same `null` as a genuine "no IdP" answer, so an outage is indistinguishable from a not-found and the user gets no retry signal. This splits the lookup into a discriminated result so an outage shows a distinct "temporarily unavailable, please try again" message, while unknown, unprovisioned, and soft-deleted emails keep the single generic message — preserving the non-enumeration contract. It also enables `withCredentials` on the axios instance defensively.

These changes are independent of the backend Entra 401 under investigation; they harden the login UX regardless of why a given login fails.

## Changes

- `src/utils/authLookup.ts` — `lookupIdpForEmail` now returns `{ idp } | { unavailable: true }`. A clean backend answer (including the deliberate `null`) returns `{ idp }`; a transport/server failure or a malformed 2xx body returns `{ unavailable: true }`.
- `src/views/LoginPage/LoginPage.tsx` — branch on the result: the Okta/Entra redirects are unchanged; `null` keeps the existing generic message; `unavailable` shows the new retry message. The `unavailable` state fires independently of whether an email exists, so it carries no enumeration signal.
- `src/axiosConfig.ts` — set `withCredentials: true` explicitly. Defensive only: same-origin `/api/v1/` already sends cookies, so this has no effect today; it guards a future separate-origin API.
- Tests — new `src/utils/authLookup.test.ts` (okta/entra/null, timeout, network error, 4xx/5xx, malformed body) and added `LoginPage` submit cases that assert a genuine `null` and an outage stay on distinct messages.

## Non-enumeration contract

Unknown email, unprovisioned, soft-deleted, and a genuine `null` all render identically. Only a transport/server failure surfaces a distinct, retryable state, and that state does not depend on whether the email maps to an account. The tests lock this so a later change can't quietly reintroduce an enumeration oracle.

## Deploy note

Opened as a draft on purpose. `orchestration-dev.yml` deploys to dev on a non-draft PR, and we don't want a UI change landing on dev while the backend Entra 401 verification is still in flight — a half-verified fix could mask that signal. Mark this ready for review once the backend verification is complete, at which point the dev deploy runs.

**Rebase onto `main` before marking ready / merging.** This branch was cut from `7b32b30` (#434), one commit before #438 (batch OpDiv grants) landed on main, so its tree still carries the pre-#438 `OpDivGrantModal`/`userOpdivs`/`UserTable`. Without the rebase, marking ready would deploy a tree missing #438 to dev (and the eventual merge would ship it stale). The changed files are disjoint — #438 touches `userOpdivs`/`OpDivGrantModal`/`UserTable`, this PR touches `axiosConfig`/`authLookup`/`LoginPage` — so the rebase is conflict-free.

## Follow-up

Client-side outage telemetry — a fire-and-forget beacon on the `unavailable` branch plus a backend sink and alarm — is tracked in CMS-Enterprise/ztmf#378 and intentionally left out of this PR so the UX fix is not blocked on the backend piece. Refs CMS-Enterprise/ztmf#378.

## Test plan

- [x] Rebase onto `main` to pick up #438 (batch OpDiv grants) before marking ready / merging — branch predates #438; disjoint files, so conflict-free
- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn test:ci` — `authLookup` and `LoginPage` suites green, existing suites unaffected
- [ ] Manual: a healthy lookup still redirects to `/login` or `/login/entra`; a forced failure shows the "temporarily unavailable" message; an unknown email still shows the generic message

<details>
<summary>Background: dev Entra login RCA (excerpt)</summary>

Entra login on dev broke in two distinct ways, both obscured by a logging blind spot. The first was an Entra-side block — the test user was not assigned to the Entra enterprise application, so Microsoft stopped the sign-in at its own page and never redirected back; removing the assignment requirement cleared that block. The second is a backend 401: after the ALB completes the OIDC handshake and forwards the request, the backend rejects the Entra token. Infrastructure, the ALB↔IdP handshake, and the UI all checked out healthy.

What was ruled out on the frontend: the live lookup returned the correct shape; the deployed UI had the Entra path enabled (`VITE_IDP_ENABLED=true`); the redirect was a correct full-page navigation. Okta login succeeded end-to-end, proving the ALB OIDC → backend session-mint → session path works. In other words, the UI was not the cause — which is why the changes in this PR are hardening, not a fix for the outage.

Leading backend hypothesis (handled separately in CMS-Enterprise/ztmf): the Entra branch pins the `tid` claim, which the ALB-forwarded userinfo-derived token does not carry, so the tenant check rejects it. That work is staged in the backend repo and awaiting a real Entra login to confirm.

This PR is the frontend follow-up: it does not depend on the backend cause, and it makes a lookup outage legible to the user instead of silently collapsing into "no IdP".

</details>

### DIFF
--- a/src/axiosConfig.ts
+++ b/src/axiosConfig.ts
@@ -25,7 +25,11 @@ const axiosInstance: AxiosInstance = axios.create({
   headers: {
     'Content-Type': 'application/json',
   },
-  // withCredentials: true,
+  // Same-origin requests to /api/v1/ already carry cookies, so this has no
+  // effect while the API shares the app's origin. It is set explicitly so
+  // credentialed requests keep working if the API is served from a
+  // separate origin.
+  withCredentials: true,
 })
 
 // Local development only: bypass auth with token from .env
--- a/src/utils/authLookup.test.ts
+++ b/src/utils/authLookup.test.ts
@@ -87,4 +87,11 @@ describe('lookupIdpForEmail - malformed 2xx is not a genuine null', () => {
       unavailable: true,
     })
   })
+
+  it('treats a present envelope with an absent idp marker as unavailable', async () => {
+    mock.onGet('/auth/lookup').reply(200, { data: {} })
+    await expect(lookupIdpForEmail('user@example.com')).resolves.toEqual({
+      unavailable: true,
+    })
+  })
 })
--- a/src/utils/authLookup.test.ts
+++ b/src/utils/authLookup.test.ts
@@ -1,0 +1,90 @@
+import MockAdapter from 'axios-mock-adapter'
+
+// Replace @/axiosConfig with a fresh axios instance. The production module
+// reads import.meta.env at top level and swc/jest leaves that literal in
+// the CommonJS output, which throws "Cannot use 'import.meta' outside a
+// module" on load. No interceptor is registered: every lookup call passes
+// skipAuthHandling: true, so the interceptor would rethrow untouched and
+// its absence does not change the behavior under test.
+jest.mock('@/axiosConfig', () => {
+  const axios = require('axios').default
+  return { __esModule: true, default: axios.create({ baseURL: '/api/v1/' }) }
+})
+
+import axiosInstance from '@/axiosConfig'
+import { lookupIdpForEmail } from './authLookup'
+
+const mock = new MockAdapter(axiosInstance)
+
+afterEach(() => {
+  mock.reset()
+})
+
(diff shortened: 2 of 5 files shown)


MackOverflow (approved):
looks clean

PR #445, opened 2026-07-01
Require updates to notes when users change their answers in the questionnaire

### LINKED ISSUE #436: Require update to text answer if maturity level changes
### Feature Type

UI Component

### Workflow Description

When a user changes a multiple choice answer in the questionnaire, they must also update the text answer for them to be allowed to save the answer.
It is fine if the text answer is updated without changes to the multiple choice.

### Acceptance Criteria

- [ ] Changed multiple choice answers are not saved without a corresponding update to the text answer.
- [ ] The user is informed why they cannot save the answer and are encouraged to update the text answer
- [ ] [Bonus] Text answer is checked to see if there is a substantial change, e.g., not just adding a space to the end of the string.

### Additional Context

_No response_

tarratsco:
## Summary

Closes #436.

When a user changes their multiple-choice answer on a questionnaire question, the save is now blocked until they substantially update the notes field. The blocked state is signaled inline (red border + helper text under the notes field) and the Next button is disabled until resolved. Notes-only edits stay valid; unanswered questions and read-only sessions are unaffected.

## Motivation

Today the questionnaire lets a user change a maturity-level radio answer and save it via Next without updating the notes. That produces score records whose free-text explanation no longer matches the chosen tier, which is bad for audit and downstream analysis. The ticket asks the FE to block this shape of save and tell the user why.

## What changed

* **`src/views/QuestionnairePage/saveGuard.ts`** (two new exports):
  * `isSubstantialNotesChange(curr, init)` (trim-comparison): rejects the "just added a space" case (the ticket's bonus criterion) without over-committing to a stricter rule. If we later need to reject internal-whitespace-only edits or short diffs, this is the one place to tighten.
  * `needsNotesUpdateForChoiceChange(state)`: returns `true` when the radio changed but the notes did not substantially change; returns `false` for unanswered questions and for cases where the choice is unchanged.
* **`src/constants.ts`**: new `NOTES_UPDATE_REQUIRED_MSG = 'Please update your notes to reflect the changed answer.'` alongside `CONFIRMATION_MESSAGE_QUESTION`.
* **`src/views/QuestionnairePage/QuestionnairePage.tsx`**:
  * Derives `needsNotesUpdate` once per render, right before the JSX return.
  * Notes `CssTextField` gets `error` and `helperText` properties based on that value.
  * Next button gets `disabled` property based on that value.
* **Back button / sidebar dirty-check paths (unchanged)**: They warn about unsaved changes via `ConfirmDialog`; they don't invoke `saveResponse`, so the new rule doesn't apply. Discard-navigating from an invalid state clears it via the existing `questionId` effect re-seed.

### Automated

* `saveGuard.test.ts` extended with two new `describe` blocks (13 new cases):
  * `isSubstantialNotesChange`: identical strings, trailing-space added, leading-space added, internal edit, both empty, add from empty.
  * `needsNotesUpdateForChoiceChange`: unanswered case, choice unchanged (regardless of notes), choice changed with unchanged notes, choice changed with only trailing whitespace, choice changed with only leading whitespace, choice changed with substantive edit, choice changed from empty-notes baseline.
* `make check` and `make test` green (210 passing, up from 197).

### Manual (any non-read-only questionnaire session)

* [ ] **Fresh answer + notes**: unanswered question → pick a choice → type notes → click Next. Score POSTs, moves forward.
* [ ] **Choice change without notes**: previously answered question → change the radio, don't touch notes. Notes field shows red border + "Please update your notes to reflect the changed answer." Next is disabled.
* [ ] **Choice change with substantial notes update**: same starting state → edit notes meaningfully → error clears, Next enables → click Next. PUT fires, moves forward.
* [ ] **Notes-only change**: keep the radio, edit notes → Next stays enabled, no error → click Next. PUT fires (unchanged behavior).
* [ ] **Whitespace-only notes change**: change radio → append a space or newline → error persists, Next stays disabled.
* [ ] **Navigate away with invalid state**: in the blocked state, click a sidebar item or Back → ConfirmDialog still opens; discarding clears the invalid state.
* [ ] **Read-only mode**: verify no error state ever appears (ChoiceList disabled means `selectQuestionOption` can't diverge from `initQuestionChoice`).

## Notes for reviewers

* **Trim-only substantial-change definition** matches the ticket's stated bar literally. Stricter definitions (normalize internal whitespace, min char-diff threshold) can be dropped in behind the same helper name later, meaning no call-site changes are needed.
* **Eager UX**: the error appears as soon as the user diverges the choice without matching notes, not deferred to a Next click. Rationale: the user has already taken a deliberate action (changed the radio), so guidance is most useful at that moment. Avoids introducing a `showValidationError` state and the bookkeeping around when to clear it (choice re-selection, question change, back navigation, etc.). Matches the existing notes-counter pattern, which is also eager.
* **Backstop in `saveResponse()`** is redundant with the disabled button today. It's there so a future save trigger (autosave, dedicated Save button, route-leave hook) doesn't silently regress the rule.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -56,6 +56,9 @@ export const CONFIRMATION_MESSAGE =
 export const CONFIRMATION_MESSAGE_QUESTION =
   'Your changes will not be saved! Are you sure you want to leave question without saving your changes?'
 
+export const NOTES_UPDATE_REQUIRED_MSG =
+  'Please update your notes to reflect the changed answer.'
+
 /**
  * Pillars hidden on the questionnaire for SaaS systems. Names must match the API's
  * `question.pillar.pillar` value exactly.
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -29,6 +29,7 @@ import {
   STATUS_MESSAGES,
   MAX_QUESTIONNAIRE_NOTES_LENGTH,
   CONFIRMATION_MESSAGE_QUESTION,
+  NOTES_UPDATE_REQUIRED_MSG,
 } from '@/constants'
 import { isAuthHandled, notify } from '@/utils/notify'
 import { sortPillars } from '@/utils/sortPillars'
@@ -40,7 +41,10 @@ import ScoreDiffModal from '@/components/ScoreDiffModal/ScoreDiffModal'
 import { useContextProp } from '../Title/Context'
 import { isAdmin, isReadOnlyAdmin } from '@/utils/userRoles'
 import LastEditedFooter from './LastEditedFooter'
-import { shouldPersistResponse } from './saveGuard'
+import {
+  shouldPersistResponse,
+  needsNotesUpdateForChoiceChange,
+} from './saveGuard'
 type Category = {
   name: string
   steps: FismaQuestion[]
@@ -206,6 +210,19 @@ export default function QuestionnarePage() {
   const saveResponse = async () => {
     if (
       !shouldPersistResponse({
+        selectQuestionOption,
+        initQuestionChoice,
+        notes,
+        initNotes,
+      })
+    ) {
+      return
+    }
+    // Backstop: the Next button is disabled when this fires, but a future
+    // save trigger (autosave, dedicated Save, route-leave hook) would
+    // reach saveResponse without knowing about the rule. Cheap insurance.
+    if (
+      needsNotesUpdateForChoiceChange({
         selectQuestionOption,
         initQuestionChoice,
         notes,
@@ -472,6 +489,15 @@ export default function QuestionnarePage() {
       </>
     )
   }
+  // Inline validation: when the user has flipped their answer without
+  // substantially editing the notes, we block the save (Next button) and
+  // surface the reason under the notes field. See saveGuard.ts for the rule.
+  const needsNotesUpdate = needsNotesUpdateForChoiceChange({
+    selectQuestionOption,
+    initQuestionChoice,
+    notes,
+    initNotes,
+  })
   return (
     <>
       <Box
@@ -615,6 +641,10 @@ export default function QuestionnarePage() {
                     fullWidth
                     value={notes}
                     disabled={isReadOnly}
+                    error={needsNotesUpdate}
+                    helperText={
+                      needsNotesUpdate ? NOTES_UPDATE_REQUIRED_MSG : undefined
+                    }
                     inputProps={{ maxLength: MAX_QUESTIONNAIRE_NOTES_LENGTH }}
                     onChange={(e) => {
                       setNotes(e.target.value)
@@ -708,6 +738,7 @@ export default function QuestionnarePage() {
                         }
                         setLoadingQuestion(false)
                       }}
+                      disabled={needsNotesUpdate}
                       style={{ marginBottom: '8px', marginTop: '8px' }}
                     >
                       {selectedIndex ===
--- a/src/views/QuestionnairePage/saveGuard.test.ts
+++ b/src/views/QuestionnairePage/saveGuard.test.ts
@@ -1,4 +1,9 @@
-import { shouldPersistResponse, ResponseState } from './saveGuard'
+import {
+  shouldPersistResponse,
+  isSubstantialNotesChange,
+  needsNotesUpdateForChoiceChange,
+  ResponseState,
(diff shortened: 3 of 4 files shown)


MackOverflow:
I have a question here @tarratsco, is this workflow required once an answer is saved? At what point is the notes update requirement triggered?

There's been some questions around the ability to review and update answers, and I can see a world where a user might re-assess their multiple choice answer without necessarily having new notes to add. 

How are we defining 'substantially update' for the notes field? I'm just trying to track if this is an added restriction (or a rework of something we have currently) and ensure that it doesn't cause undue friction.

tarratsco:
@MackOverflow As is, it triggers a requirement to update notes as soon as you change your answer to a multiple choice question. So it does not let you proceed/go to next question till you update the notes. 

And "substancialChange" as of now, is adding more than just a space at the beginning/end of the notes. Is nothing comprehensive as its currently written, but it can be expanded if requirements change in the future. For now, is just detecting that you dont try to trigger a change by adding whitespace.

tarratsco:
image
Control buttons are disabled.


https://github.com/user-attachments/assets/fba740ca-7bbd-49c4-9e55-d3ed4685968a

danielbowne (dismissed):
Hey @tarratsco, this is looking good. Nice call pulling the guard logic out into saveGuard.ts with its own tests, made it easy to follow. Ran it locally, typecheck is clean and all 18 jest tests pass. Two things I want to sort out before we merge, plus a small test gap.

**1. First-time answers get blocked (the main one)**

`needsNotesUpdateForChoiceChange` only short-circuits when the current selection is empty (`selectQuestionOption === -1`). It doesn't short-circuit when there was no prior answer (`initQuestionChoice === -1`). On load both get set together, so a question that has never been answered starts at `initQuestionChoice === -1`, and the first time someone answers it with an empty notes field the guard fires: Next/Complete goes disabled and they get the "please update your notes to reflect the changed answer" message.

That trips on a real path, not just a corner case. Scores come back per data call, so on any new data call every question starts unanswered, and users would be forced to write a note on every single first answer. That's broader than the PR title ("change their answers"), and it also runs against your own docstring, which says the rule should only apply once a response has already been answered.

So it's really a question of intent:

- If the rule is meant to fire only when someone changes an existing answer (which the title and docstring both suggest), then the guard just needs to also bail early when there's no prior answer:

  ```ts
  if (s.initQuestionChoice === -1) return false
  ```

- If you actually want notes required on every answer including the first, then let's update the docstring and the "changed answer" wording so they match, since right now they describe change-only behavior.

Either way is fine, just want them lined up.

**2. Clearing the notes slips past the guard**

`isSubstantialNotesChange` treats any trimmed difference as a change, and that includes deleting the notes. So someone can change their answer, wipe the notes field, and save with no explanation, which is the thing this is trying to prevent:

```
{ selectQuestionOption: 7, initQuestionChoice: 5, notes: '', initNotes: 'old reason' } -> needsNotesUpdate === false
```

If the goal is to guarantee an explanation rides along with a changed answer, this probably wants a non-empty check too, something like also requiring `notes.trim().length > 0` when the choice changed.

**3. Test coverage**

The suite covers whitespace-only edits and text-added-from-empty on an already-answered baseline, but neither of the two cases above has a test. Adding them would lock in whatever behavior we land on for 1 and 2.

Once 1 is settled and 2 has a decision, I'll re-review quickly and we're good to go. Thanks for this.

tarratsco:
@danielbowne Thanks for the thorough review! And yeah the intent is both to:
- fire only when someone changes an existing answer  
- prevent someone from completely deleting the notes and moving on

I'll apply the recommended changes accordingly 👍

tarratsco:
@danielbowne  Updated!

https://github.com/user-attachments/assets/683fa162-3d46-4375-9b87-be972bdb35c0


https://github.com/user-attachments/assets/2f1bdab5-1c2c-4ccc-9fd4-a79ec5b8ddf4

danielbowne (approved):
Re-reviewed at `d63ad6ad5`. Both fixes are exactly right, clearing my change request.

- First-time answers: the `initQuestionChoice === -1` short-circuit means a fresh data call no longer forces a note on every question, and the docstring now matches the behavior.
- Cleared notes: the `notes.trim().length === 0` check blocks wiping the notes on a changed answer, so an explanation has to ride along.

Ran it locally, typecheck is clean and all 21 saveGuard tests pass, including the new cases for the first-time answer and the cleared-notes path. Videos look good too.

Approving. Thanks for turning this around.

PR #447, opened 2026-07-02
Hhs datacall schema integration

tarratsco:

Issue #390, opened 2026-06-12
Auth error handling follow-ups: parseApiError domain-layer consistency + outOfScope 403 copy


danielbowne:
Follow-up to #329 / PR #388, which centralized 401/403 handling behind the axios interceptor. Two items were intentionally deferred so #388 could merge on review; tracking them here.

## 1. Finish parseApiError domain-layer wiring

The interceptor now owns 401/403. The per-call domain branch (400 field-error maps, 404/5xx fallback) should run through `parseApiError` (`src/utils/apiErrors.ts`). `UserTable`, `OpDivAdmin`, and `OpDivGrantModal` already do. Three views still hand-roll their non-auth catch with a generic `tryAgain` snackbar / inline 400 handling:

- `src/views/AssignSystemModal/AssignSystemModal.tsx`
- `src/views/EditSystemModal/EditSystemModal.tsx`
- `src/views/QuestionnairePage/QuestionnairePage.tsx`

Wire them through `parseApiError` so 400 field-error rendering is consistent across the app.

## 2. Resolve the outOfScope 403 copy

`ERROR_MESSAGES.outOfScope` (`src/constants.ts:27`) is now a dead constant — #378 used it for OpDiv-scope 403s, but #388 routes all 403s through the interceptor, which surfaces the backend `response.data.error` or the generic `ERROR_MESSAGES.permission`. Decide:

- delete the unused constant, or
- restore the OpDiv-scope-specific 403 copy via the interceptor default (or confirm the backend returns scope-specific text in `response.data.error`).

Refs #329
Refs #388

tarratsco:
Hey @danielbowne would this be related to https://github.com/CMS-Enterprise/ztmf-ui/issues/389?

danielbowne:
@tarratsco you're right, this is a duplicate of #389, and yours is the better-scoped ticket. It correctly splits the notify() migration from the parseApiError propagation, where this one had lumped them together and miscategorized AssignSystemModal/QuestionnairePage as parseApiError work when they are really just notify() conversions.

The only item here not already covered by #389 is the outOfScope 403 copy: `ERROR_MESSAGES.outOfScope` (`src/constants.ts`) is a dead constant now that the interceptor owns 403s, so it is either delete it or restore OpDiv-scope-specific 403 copy. If you want that tracked, add it as a comment or a line on #389. Otherwise it is safe to drop.

Closing this as a duplicate of #389.

tarratsco:
[...]

PR #451, opened 2026-07-07
Add OpDiv selector and HHS expansion fields to Add/Edit System modal

### LINKED ISSUE #449: Add/Edit System form: OpDiv selector + 12 new HHS fields; resolve ISSO / Data Call Contact semantics
## Context
The HHS multi-OpDiv expansion added an OpDiv relationship and 12 new `fismasystems` columns. The Add / Edit System modal has neither, so systems cannot be created or edited correctly for HHS OpDivs.

## Scope

**1. OpDiv selector (required), top priority.** `opdiv_id` is a NOT NULL FK; the form has no OpDiv control today. Add a required select populated from `GET /api/v1/opdivs` (value `opdiv_id`, label `name`/`code`). Without it, non-CMS systems cannot be created.

**2. Add the 12 new fields** (Add + Edit): `isso_name`, `hva`, `fips`, `system_type`, `cloud_system`, `cloud_service_model`, `cloud_vendor`, `system_operator`, `goco_coco_gogo`, `system_owner`, `system_owner_email`, `legacy`. All `varchar(255)`, text inputs for now. RBAC-gate them to match the backend (unscoped admins write; scoped OpDiv admins read-only, mirroring `clearHHSMetadata` / `copyHHSMetadata` so a scoped edit does not appear to wipe them). `isso_name` depends on backend migration `0044` (#389); coordinate so the field is not shipped before the column exists.

**3. Resolve ISSO vs Data Call Contact semantics (product decision).** With `isso_name` added, the ISSO is now a name + email pair (`isso_name` + `issoemail`); present them together as "ISSO Name" and "ISSO Email". Separately, `datacallcontact` is a single email-validated field (`isValidEmail` on the backend) despite its "Data Call Contact" (person) label. Decide whether Data Call Contact is a name, an email, or both, and fix the label vs validation disagreement. Needs an owner.

**4. CMS-specific fields (product decision).** `Group Name`, `Division Name`, `Group Acronym`, `Component` were CMS-context fields with some CMS data already loaded. Decide whether they apply to HHS-OpDiv systems and whether they stay required (`Component` is currently required), become optional, or hide for non-CMS OpDivs.

## Current form (for reference)
_Screenshot of today's Add Fisma System modal to be attached here._

Fields present today: Fisma Name*, Data Call Contact*, Fisma Acronym*, Group Acronym, Component*, ISSO Email*, Group Name, Fisma UID*, Division Name, Datacenter Environment*, Fisma Subsystem, Sync to SDL toggle. No OpDiv control and none of the 12 new fields.

## Out of scope
`notes_is_ai_summary` (questionnaire AI badge) is a separate surface.

## Acceptance criteria
- Required OpDiv selector on Add + Edit, populated from `/opdivs`.
- All 12 new fields present, RBAC-gated to the backend write rules.
- Recorded product decisions for the ISSO / Data Call Contact semantics and the CMS-specific fields.

Image

MackOverflow:
## What this PR does

This PR implements [issue #449](https://github.com/CMS-Enterprise/ztmf-ui/issues/449) — the Add/Edit System form changes required for the HHS multi-OpDiv expansion — and folds in all of the related work from [PR #447](https://github.com/CMS-Enterprise/ztmf-ui/pull/447) so both land together in a single, reviewed changeset.

---

## From issue #449: Add/Edit System form

The HHS multi-OpDiv expansion added an `opdiv_id` foreign key and 12 new columns to `fismasystems`. The existing Add/Edit modal had neither, so HHS OpDiv systems could not be created or edited correctly. This PR fixes that.

### Required OpDiv selector
Adds a full-width, required `opdiv_id` select field at the top of the Add/Edit System modal, populated from `GET /api/v1/opdivs`. Because `opdiv_id` is a NOT NULL FK on the backend, this field is required — the Save/Create button stays disabled until an OpDiv is chosen.

### 12 new HHS metadata fields
All 12 fields from the issue scope are now present on both Add and Edit:

| Field | Label |
|---|---|
| `isso_name` | ISSO Name |
| `hva` | HVA |
| `fips` | FIPS Impact Level |
| `system_type` | System Type |
| `legacy` | Legacy |
| `cloud_system` | Cloud System |
| `cloud_service_model` | Cloud Service Model |
| `cloud_vendor` | Cloud Vendor |
| `system_operator` | System Operator |
| `goco_coco_gogo` | GOCO/COCO/GOGO |
| `system_owner` | System Owner |
| `system_owner_email` | System Owner Email |

### RBAC gating
The HHS Metadata section is only shown and editable for HHS-wide admins (`hasUnscopedRead` — OWNER, HHS_ADMIN). Scoped OpDiv admins see the modal without the HHS section; the backend enforces the same boundary server-side. This mirrors the approach from #447 (`hhsEditable` prop) for consistency.

> **`isso_name` note:** This field depends on backend migration `0044` (#389). It is present in the form but should not ship to production before that column exists. Coordinate with the backend timeline.

---

## From PR #447: HHS datacall schema integration

#447 (by @onix-tarrats-calderon) introduced complementary HHS work that was not yet merged. Rather than land two overlapping PRs, its changes are incorporated here with #447 taking architectural priority. The following is a summary of what #447 contributed:

### Bug fix
- **`FismaTable.tsx`** — `issoemail` can be `null` on HHS-onboarded systems. The previous code called `.split('@')` on the raw value unconditionally, crashing the DataGrid on scroll whenever a null row rendered. Now guarded with an early return.

### HHS Metadata section on the System Detail page
- **`fieldConfig.ts`** — adds an `hhs` section to the field registry. All 12 HHS fields (plus `isso_name` added in this PR) are defined here. The `hhs` section drives both the modal and the System Detail page so labels and field order stay in sync.
- **`SystemDetailReadView.tsx`** — renders a new "HHS Metadata" card in the read view, populated from `fieldConfig`.
- **`SystemDetailEditView.tsx`** — renders a "HHS Metadata" card in the edit view. Fields are editable for HHS-wide admins, disabled with a tooltip for everyone else.
- **`SystemDetailPage.tsx`** — passes `hhsEditable={hasUnscopedRead(userInfo)}` down to the edit view.
- **`Title.tsx`** — passes `hhsEditable={hasUnscopedRead(userInfo)}` to the Add System modal.

### AI Summary badge
- **New `AISummaryBadge` component** — a small chip that reads "AI Summary" with an `AutoAwesome` icon and a tooltip explaining that the note was AI-generated from HHS narrative data.
- **`QuestionnairePage.tsx` and `QuestionnareModal.tsx`** — badge shown next to the notes character counter when `notes_is_ai_summary === true`.
- **`ScoreDiffModal.tsx`** — badge shown inline with score notes in the diff view.
- **`types.ts`** — `notes_is_ai_summary?: boolean` added to `QuestionScores` and `ScoreDiffSide`.

### DataCall modal — ZTM format support
- **`DataCallModal.tsx`** — the validation pattern was previously hard-coded to `FYYYYY QN` (9 characters exactly). HHS uses an annual cadence named `FY23 ZTM` / `FY25 ZTM`. The pattern is now widened to accept both formats, the character limit raised from 9 to 10, the submit-button gate corrected to match, and the modal's state now resets on close so stale errors don't bleed into the next open.

### ValidatedTextField — `disabled` prop
- **`ValidatedTextField.tsx`** — adds a `disabled` prop so the component can be rendered read-only, used by the HHS Metadata edit fields for scoped admins.

---

## Files changed (15)

| File | What changed |
|---|---|
| `src/types.ts` | 12 HHS fields + `isso_name` on `FismaSystemType`; `hhsEditable` on `editSystemModalProps`; `notes_is_ai_summary` on `QuestionScores` + `ScoreDiffSide` |
| `src/views/EditSystemModal/emptySystem.ts` | Initializes all new fields to `null` |
| `src/views/EditSystemModal/EditSystemModal.tsx` | OpDiv selector; `hhsEditable`-gated HHS section using `fieldConfig`; `opdiv_id` in POST/PUT |
| `src/views/EditSystemModal/ValidatedTextField.tsx` | `disabled` prop |
| `src/views/SystemDetailPage/fieldConfig.ts` | `hhs` section with all 13 fields |
| `src/views/SystemDetailPage/SystemDetailEditView.tsx` | HHS Metadata card, editable/disabled by `hhsEditable` |
| `src/views/SystemDetailPage/SystemDetailReadView.tsx` | HHS Metadata read card |
| `src/views/SystemDetailPage/SystemDetailPage.tsx` | Passes `hhsEditable` |
| `src/views/Title/Title.tsx` | Passes `hhsEditable` to Add modal |
| `src/components/AISummaryBadge/AISummaryBadge.tsx` | New component |
| `src/components/ScoreDiffModal/ScoreDiffModal.tsx` | AI badge in diff view |
| `src/views/FismaTable/FismaTable.tsx` | Null `issoemail` crash fix |
| `src/views/QuestionnairePage/QuestionnairePage.tsx` | AI badge next to notes |
| `src/views/QuestionnareModal/QuestionnareModal.tsx` | AI badge next to notes |
| `src/views/DatacallModal/DataCallModal.tsx` | ZTM format + state reset |

Closes #449
Incorporates #447


costacalvin:
@MackOverflow There is a small gap I noticed in the #447 PR regarding the AI badge that can be added here: 

Any notes sections with AI summaries would be labeled as AI even if a user subsequently modifies it to remove the AI summary. Both QuestionnareModal.tsx (just noticed it has a typo) and QuestionnairePage.tsx PUT notes without `notes_is_ai_summary: false`, so the backend preserves true and the AI badge stays after a human edits the notes. Fix is to add `notes_is_ai_summary: false` to the PUT body in both save handlers as the users are modifying the AI summary and should therefore no longer be considered AI. This would still keep the AI label if a user "modifies" the notes but they are identical in content.

QuestionnairePage.tsx L225:

+ Add `notes_is_ai_summary: false,`

QuestionnarePage.tsx L162:

+ Add `notes_is_ai_summary: false,`

src/views/QuestionnairePage/QuestionnairePage.tsx
costacalvin, line 224: `notes_is_ai_summary: false,` here would remove an AI summary badge if notes content is modified

src/views/QuestionnareModal/QuestionnareModal.tsx
costacalvin, line 161: adding notes_is_ai_summary: false, here would remove an AI summary badge if notes content is modified

danielbowne (dismissed):
Thanks for pulling #449 and #447 into one changeset. The modal path for #449 is solid and the RBAC gating reads clean across all the surfaces, defaulting `hhsEditable` to false is the right call. Typecheck is green locally. Before this merges I have three things, two of them blockers on the System Detail edit page, plus a deploy ordering note.

**1. Save is permanently disabled on the System Detail edit page (hits every write admin)**

`system_owner_email` is `type: 'email', required: false` in `fieldConfig.ts:174-179`, but the edit view renders email fields through `ValidatedTextField` + `emailValidator`. That validator's mount effect (`ValidatedTextField.tsx:24-31`) runs on the empty value, and `emailValidator` returns "This field is required" for empty (`validators.ts:2-3`), which pushes `system_owner_email: false` into `formValid`. `isFormValid()` requires every entry to be true (`SystemDetailPage.tsx:208-210`), so Save stays greyed out and a red required error shows on an optional field.

Since it is a brand new column it is empty on every existing system, so this hits every write admin (OWNER, HHS_ADMIN, OPDIV_ADMIN) the moment they open edit, not just HHS admins. The disabled field still fires the mount validation. A few ways out: give optional email fields a validator that passes on empty, exclude non-required fields from the `isFormValid()` gate, or render the HHS fields as plain inputs like the modal does.

**2. HHS metadata edited on the detail edit view never saves**

`handleSave` on the detail page PUTs a fixed allowlist (`SystemDetailPage.tsx:272-285`) that includes none of the 12 HHS keys, but the edit view renders those fields editable for HHS admins and mutates `editedSystem`. So an HHS admin can type into ISSO Name or HVA, hit Save, get the green toast, and nothing persists. The modal's `handleSave` does this right (sends the HHS keys when `hhsEditable`, `EditSystemModal.tsx:274-278, 323-327`), the detail page just needs to match it, gated on `hasUnscopedRead`. Or lock the detail card read-only so it does not imply editability. Item 1 currently hides this since you cannot reach a successful save, fixing 1 surfaces 2.

**3. Deploy dependency is wider than the isso_name note**

The body flags `isso_name` / migration 0044, but `opdiv_id` is the harder gate. It is sent on every create and edit unconditionally (`EditSystemModal.tsx:272, 319`, outside the hhsEditable block) and blocks Save until chosen, and the dropdown is filled from `GET /opdivs` (`opdivs.ts`, whose comment still says there is no production caller yet). No feature flag anywhere. If `/opdivs` is not live in the target env the selector is empty, `formValid.opdiv_id` never flips true, and Save/Create is disabled for everyone, so nobody can create or edit any system. If `/opdivs` exists but the columns do not and the backend rejects unknown fields, the write 400s. Can we confirm `/opdivs` plus all 13 columns (opdiv_id + 12 HHS) are live in dev and prod before this promotes? Worth adding to the deploy checklist alongside 0044.

Related: existing systems all have null `opdiv_id` today, so on the modal edit path they cannot be saved until an OpDiv is assigned (`EditSystemModal.tsx:178`). If the backend backfilled existing rows that is fine, if not then every edit of a legacy system forces an OpDiv pick. Worth confirming the backfill.

**Non-blocking**

- Read-view HHS Metadata card is not gated (`SystemDetailReadView.tsx:205+`), so scoped OpDiv admins, readonly, and ISSO/ISSM all see it, and it renders even on CMS systems where every field is empty, giving an empty card on every non-HHS system. Edit is gated but read is not. If the data is not view sensitive this is just cosmetics, consider hiding the card when all fields are blank.
- The opdiv fetch abort controller is inert (`EditSystemModal.tsx:115-122`): the signal is never passed to `fetchOpDivs`, so the cleanup abort does nothing and `setOpDivs` can fire after close. The error is also swallowed by an empty catch, so an `/opdivs` failure leaves Save silently stuck with no message. Either thread the signal and surface the error, or drop the controller.
- Two lists define the HHS fields: `HHS_METADATA_KEYS` drives the modal save loop, `fieldConfig`'s hhs section drives the render. They match today at 12 each, but a field added to one and not the other would silently render-but-not-save. Deriving the save list from fieldConfig removes the drift risk.
- Small stuff: `issoemail` is still typed `string` in `types.ts` though it is nullable (the runtime guard covers it, but it could be `string | null`), the modal HHS inputs use hardcoded `create-${key}` ids and "Set at create time only" copy that reads oddly in edit mode, and `DataCallModal`'s `Q\d` accepts Q0 and Q5-9 if you want to tighten to `Q[1-4]` (not a regression).

FismaTable null fix and the DataCallModal ZTM widening both look correct. Not re-raising the AI badge point since costacalvin already has that covered.

Nice work overall, most of this is in good shape. Once the two detail-page save bugs and the deploy ordering are sorted I will take another pass.

stu-aq:
OS and OIG are currently under OpDivs even though they are technically staffing divisions, but we are confirming they can stay put with the other OpDivs.  For the Add Fisma Form, it might make sense to add Division/Staffing Name, but this would be a small detail.

Is it possible to make FIPS, HVA, cloud model, GOCO dropdowns instead of free text inputs?

tarratsco:
Some of the issues that surfaced while looking through the PR:

* **In `src/views/SystemDetailPage/SystemDetailPage.tsx:272-285` (`handleSave` PUT body drops all HHS fields and `opdiv_id`) - Data-loss bug.**
  An HHS admin opens `/systems/:id`, edits the HHS Metadata card (say, sets `hva='Yes'`, `system_owner='J. Doe'`), and clicks Save. The "Saved!" toast fires. Reload the page: every HHS change is gone. The PUT body enumerates only the pre-existing 12 fields (`fismauid`, `fismaacronym`, ..., `sdl_sync_enabled`) and never sends the HHS keys. `EditSystemModal.tsx` was updated to conditionally include `HHS_METADATA_KEYS` in its own PUT/POST (diff 293-316), but the parallel `SystemDetailPage.handleSave` was missed. The same gated block needs to go here.

* **In `src/views/SystemDetailPage/SystemDetailEditView.tsx` (HHS `system_owner_email` field) - Save button gets stuck disabled on any system with an empty email.**
  Blocks all edits. `emailValidator` returns `'This field is required'` on empty input (`src/views/EditSystemModal/validators.ts:2-4`). The HHS section renders `system_owner_email` as a `ValidatedTextField` (`email` type in `fieldConfig`). Its mount effect calls `validator('')` → error → `onChange(false, '')` → `handleValidatedFieldChange` sets `formValid.system_owner_email=false`. `isFormValid` uses `Object.values(formValid).every(v => v === true)`, so Save is disabled forever. This fires for every system whose `system_owner_email` is null, which is most HHS-loaded systems and every existing non-HHS system. It fires even for non-HHS admins because the disabled path (`<Tooltip><Box>{renderEditField(field, props, true)}</Box></Tooltip>`) still mounts the `ValidatedTextField` and runs its effect.

MackOverflow:
Pushed b605171 addressing the review feedback. Summary of what changed:

### Blockers — System Detail edit page
- **Save permanently disabled (@danielbowne, @tarratsco):** Optional email fields now use an `optionalEmailValidator` (empty is valid), disabled `ValidatedTextField`s no longer contribute to the form-valid gate, and `ValidatedTextField` honors a `required` prop instead of hardcoding it. Save is no longer stuck on systems with an empty `system_owner_email`.
- **HHS / `opdiv_id` data loss on `handleSave` (@danielbowne, @tarratsco):** `SystemDetailPage.handleSave` now sends `opdiv_id` on every PUT and includes the HHS keys when `hasUnscopedRead`, mirroring the modal save path. Detail-page HHS edits now persist.

### AI summary badge (@costacalvin)
- Both notes PUTs (`QuestionnairePage`, `QuestionnareModal`) now send `notes_is_ai_summary: false`, so a human edit clears the badge. The existing dirty-checks keep the badge when the note is saved unchanged.

### Non-blocking cleanups (@danielbowne)
- `HHS_METADATA_KEYS` is now derived from `fieldConfig` (single source of truth), removing the render-but-not-save drift risk.
- Read-view HHS Metadata card is hidden when no HHS fields are populated (no more empty card on non-HHS systems).
- Abort signal threaded into `fetchOpDivs`; `/opdivs` load failures now surface via a toast instead of being swallowed.
- `issoemail` typed `string | null`.
- Modal HHS input ids + helper copy are now mode-aware.
- Datacall quarter pattern tightened to `Q[1-4]`.

### Deploy dependency — need backend confirmation before promoting (@danielbowne #3)
`opdiv_id` is required on every create/edit and the selector is populated from `GET /opdivs`. Before this promotes, can the backend team confirm the following are live in **dev and prod**:
1. `GET /api/v1/opdivs` returns the reference list in both envs.
2. All 13 new columns exist on `fismasystems` and the write endpoints accept them without 400-ing: `opdiv_id` + the 12 HHS fields.
3. Migration 0044 (#389) for `isso_name` is applied.
4. Existing `fismasystems` rows have been backfilled with an `opdiv_id` (otherwise every legacy-system edit forces an OpDiv pick).

If any of these lags the UI, I'll gate the OpDiv selector + HHS section behind a feature flag instead.

### Product note (@stu-aq)
Re: OS/OIG as staffing divisions under OpDivs — noted, leaving them filed under OpDivs for now pending your confirmation. I'll spin the "Division/Staffing Name" field on the Add form into a separate follow-up issue rather than expand this PR's scope.

---
Separately, an agency-agnostic language cleanup is being tracked as its own effort and will not land in this PR.

costacalvin:
UI/Code verifications: 

1. "Save permanently disabled" ✅  - Verified system_owner_email is no longer required + allows for a Save.
2. "HHS / opdiv_id data loss on handleSave" ✅ - Verified metadata edits on the details page persists after a save.
3. "AI summary badge" ✅  - Verified badge gets removed after user edits/removes the AI summary
4. "Read-view HHS Metadata card" ✅ - Verified Metadata section is hidden for systems with no metadata fields populated.

Deploy dependency verifications:
Backend PR (ztmf#390) has been merged. All four checklist items are confirmed resolved and in Prod through github actions:

1. **`GET /api/v1/opdivs` in dev and prod** ✅  — Route is registered (`router.go`)

2. **All 13 columns on `fismasystems` + write endpoints accept them** ✅  — Migration 0043 added the 11 HHS fields, migration 0044 added `isso_name`. All 13 columns (including `opdiv_id`) are in the model's column list and the UPDATE/INSERT paths handle them. No 400s on write.

3. **Migration 0044 (`isso_name`)** ✅ — Applied. Note: the PR description links `#389` which auto-resolves to `ztmf-ui#389` — the correct issue reference is [ztmf#389](https://github.com/CMS-Enterprise/ztmf/issues/389) with the PR being ztmf#390.

4. **Existing rows backfilled with `opdiv_id`** ✅ — Migration 0028 backfilled all pre-existing `fismasystems` rows to the CMS `opdiv_id`, and migration 0029 enforced `NOT NULL`. No legacy system has a null `opdiv_id`, so editing an existing system will not force an OpDiv re-pick.

stu-aq:
Not entirely sure when in the process this data is pushed into SDL, but we need to ensure the `CORE.ZTMF_SCORES` field length in SDL is increased from 1000 to 2000, or the push will either truncate the data at 1000 or cause an error.

MackOverflow:
@danielbowne we'll for sure need to address Seth's comment above!

MackOverflow:
@danielbowne for requested changes:

 1. Save permanently disabled on the System Detail edit page — Fixed. Added optionalEmailValidator (empty passes), made ValidatedTextField report valid for disabled fields so they no longer gate the form, and it now honors a real required prop instead of hardcoding it. system_owner_email empty no longer locks Save.

2. HHS metadata edited on the detail view never saves — Fixed. SystemDetailPage.handleSave now sends opdiv_id on every PUT and includes the HHS keys when hasUnscopedRead(userInfo), mirroring the modal path.

Non-blocking — all folded in ✅
Read-view HHS card gated (hasAnyHhsData, hidden when all blank)
Inert AbortController fixed (fetchOpDivs(false, controller.signal) + error surfaced)
Single source of truth (HHS_METADATA_KEYS derived from fieldConfig)
issoemail: string | null
Mode-aware modal ids/copy (${mode}-${field.key})
Q[1-4] tightened

tarratsco:
Re-tested locally:

- **Data-loss on `SystemDetailPage.handleSave`**: PUT now sends `opdiv_id` and the HHS keys when `hasUnscopedRead` - HHS edits persist across reload. ✅
- **Save blocked on empty `system_owner_email`**: three-pronged fix (`optionalEmailValidator`, `ValidatedTextField` skips validator when disabled, honors a `required` prop) - works for both HHS-admin and scoped-admin paths. ✅
- **`HHS_METADATA_KEYS` drift**: now derived from `fieldConfig` as single source of truth. ✅
- **Unused `AbortController`**: signal threaded into `fetchOpDivs`, and OpDivs load failures now surface via a toast instead of being swallowed -nice bonus. ✅
- **AI badge stickiness** (@costacalvin's find): both notes PUTs send `notes_is_ai_summary: false`, dirty-check preserves the badge on identical "edits". ✅

Also walked through the full manual test plan against a fresh DB + mock addon:

- OpDiv selector - populated, gates Create button, toast on `/opdivs` failure ✅
- Add modal HHS section - visible to OWNER only, values persist on new system ✅
- HHS Metadata save on detail page - `hva` / `system_owner` / `opdiv_id` persist through reload ✅
- Save no longer blocked on empty `system_owner_email` (OWNER + OPDIV_ADMIN paths) ✅
- Read view - HHS card hidden on non-HHS systems, shown on MOCK-* ✅
- AI Summary badge display + clear on human edit + preserved on identical "edit" ✅
- AI Summary badge inline in `ScoreDiffModal` ✅
- `FismaTable` renders MOCK-* rows with null `issoemail` without crashing ✅

danielbowne (approved):
Re-reviewed at `b60517172`. Everything from the last pass is addressed. Clearing the change request.

What I verified:
- Detail-page Save no longer locks on the optional email. `optionalEmailValidator` plus the `disabled` short-circuit in `ValidatedTextField` handle it, and the edit view picks the validator by `field.required`.
- Detail-page HHS edits persist now. `handleSave` sends the HHS keys gated on `hasUnscopedRead`, matching the modal.
- `HHS_METADATA_KEYS` is derived from the fieldConfig hhs section, so the render and save lists can't drift.
- OpDiv fetch threads the abort signal and surfaces failures via `notify` instead of swallowing them.
- Read-view HHS card is hidden when the system has no HHS data, so non-HHS systems don't show an empty card.
- Small ones all in: `issoemail` is `string | null`, `Q[1-4]`, mode-aware ids and helper copy, and `notes_is_ai_summary: false` on both notes save paths.

Typecheck is clean and jest passes on the touched areas locally.

On the deploy dependency from my last comment: the backend side is all in `main` as of 07-06, `/api/v1/opdivs` is registered, and the fismasystems model has `opdiv_id` plus the 12 HHS columns and `isso_name`. `opdiv_id` is backfilled and NOT NULL (migrations 0028/0029), so existing systems are covered and there's no risk of them being unsavable, disregard that part of my earlier note. One doc fix: the `isso_name` migration (0044) is #390, not #389 as the description says. The only open item is operational, not code, just make sure the backend carrying these migrations is promoted to prod before this frontend goes to prod.

Approving. Thanks for the quick turnaround.

danielbowne:
> Not entirely sure when in the process this data is pushed into SDL, but we need to ensure the `CORE.ZTMF_SCORES` field length in SDL is increased from 1000 to 2000, or the push will either truncate the data at 1000 or cause an error.

This has been done, but would be unrelated to what we are doing with data load honestly (we do not sync HHS data to SDL)

PR #454, opened 2026-07-07
Require updates to notes when users change their answers in the questionnaire (rehome of #445)

### LINKED ISSUE #436: Require update to text answer if maturity level changes

danielbowne:
Rehome of #445 by @tarratsco so the Snyk and deploy checks can run. Fork PRs cannot reach the org secrets, so those required checks sit pending forever. These are his exact commits with authorship preserved, no code changes.

Review and approval are already on #445, and I re-verified locally on this head: typecheck clean, all 21 saveGuard tests passing. Once CI is green here I will close #445 pointing at this one.

Original description below.

---

## Summary

Closes #436.

When a user changes their multiple-choice answer on a questionnaire question, the save is now blocked until they substantially update the notes field. The blocked state is signaled inline (red border + helper text under the notes field) and the Next button is disabled until resolved. Notes-only edits stay valid; unanswered questions and read-only sessions are unaffected.

## Motivation

Today the questionnaire lets a user change a maturity-level radio answer and save it via Next without updating the notes. That produces score records whose free-text explanation no longer matches the chosen tier, which is bad for audit and downstream analysis. The ticket asks the FE to block this shape of save and tell the user why.

## What changed

* **`src/views/QuestionnairePage/saveGuard.ts`** (two new exports):
  * `isSubstantialNotesChange(curr, init)` (trim-comparison): rejects the "just added a space" case (the ticket's bonus criterion) without over-committing to a stricter rule. If we later need to reject internal-whitespace-only edits or short diffs, this is the one place to tighten.
  * `needsNotesUpdateForChoiceChange(state)`: returns `true` when the radio changed but the notes did not substantially change; returns `false` for unanswered questions and for cases where the choice is unchanged.
* **`src/constants.ts`**: new `NOTES_UPDATE_REQUIRED_MSG = 'Please update your notes to reflect the changed answer.'` alongside `CONFIRMATION_MESSAGE_QUESTION`.
* **`src/views/QuestionnairePage/QuestionnairePage.tsx`**:
  * Derives `needsNotesUpdate` once per render, right before the JSX return.
  * Notes `CssTextField` gets `error` and `helperText` properties based on that value.
  * Next button gets `disabled` property based on that value.
* **Back button / sidebar dirty-check paths (unchanged)**: They warn about unsaved changes via `ConfirmDialog`; they don't invoke `saveResponse`, so the new rule doesn't apply. Discard-navigating from an invalid state clears it via the existing `questionId` effect re-seed.

### Automated

* `saveGuard.test.ts` extended with two new `describe` blocks (13 new cases):
  * `isSubstantialNotesChange`: identical strings, trailing-space added, leading-space added, internal edit, both empty, add from empty.
  * `needsNotesUpdateForChoiceChange`: unanswered case, choice unchanged (regardless of notes), choice changed with unchanged notes, choice changed with only trailing whitespace, choice changed with only leading whitespace, choice changed with substantive edit, choice changed from empty-notes baseline.
* `make check` and `make test` green (210 passing, up from 197).

### Manual (any non-read-only questionnaire session)

* [ ] **Fresh answer + notes**: unanswered question → pick a choice → type notes → click Next. Score POSTs, moves forward.
* [ ] **Choice change without notes**: previously answered question → change the radio, don't touch notes. Notes field shows red border + "Please update your notes to reflect the changed answer." Next is disabled.
* [ ] **Choice change with substantial notes update**: same starting state → edit notes meaningfully → error clears, Next enables → click Next. PUT fires, moves forward.
* [ ] **Notes-only change**: keep the radio, edit notes → Next stays enabled, no error → click Next. PUT fires (unchanged behavior).
* [ ] **Whitespace-only notes change**: change radio → append a space or newline → error persists, Next stays disabled.
* [ ] **Navigate away with invalid state**: in the blocked state, click a sidebar item or Back → ConfirmDialog still opens; discarding clears the invalid state.
* [ ] **Read-only mode**: verify no error state ever appears (ChoiceList disabled means `selectQuestionOption` can't diverge from `initQuestionChoice`).

## Notes for reviewers

* **Trim-only substantial-change definition** matches the ticket's stated bar literally. Stricter definitions (normalize internal whitespace, min char-diff threshold) can be dropped in behind the same helper name later, meaning no call-site changes are needed.
* **Eager UX**: the error appears as soon as the user diverges the choice without matching notes, not deferred to a Next click. Rationale: the user has already taken a deliberate action (changed the radio), so guidance is most useful at that moment. Avoids introducing a `showValidationError` state and the bookkeeping around when to clear it (choice re-selection, question change, back navigation, etc.). Matches the existing notes-counter pattern, which is also eager.
* **Backstop in `saveResponse()`** is redundant with the disabled button today. It's there so a future save trigger (autosave, dedicated Save button, route-leave hook) doesn't silently regress the rule.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -56,6 +56,9 @@ export const CONFIRMATION_MESSAGE =
 export const CONFIRMATION_MESSAGE_QUESTION =
   'Your changes will not be saved! Are you sure you want to leave question without saving your changes?'
 
+export const NOTES_UPDATE_REQUIRED_MSG =
+  'Please update your notes to reflect the changed answer.'
+
 /**
  * Pillars hidden on the questionnaire for SaaS systems. Names must match the API's
  * `question.pillar.pillar` value exactly.
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -29,6 +29,7 @@ import {
   STATUS_MESSAGES,
   MAX_QUESTIONNAIRE_NOTES_LENGTH,
   CONFIRMATION_MESSAGE_QUESTION,
+  NOTES_UPDATE_REQUIRED_MSG,
 } from '@/constants'
 import { isAuthHandled, notify } from '@/utils/notify'
 import { sortPillars } from '@/utils/sortPillars'
@@ -40,7 +41,10 @@ import ScoreDiffModal from '@/components/ScoreDiffModal/ScoreDiffModal'
 import { useContextProp } from '../Title/Context'
 import { isAdmin, isReadOnlyAdmin } from '@/utils/userRoles'
 import LastEditedFooter from './LastEditedFooter'
-import { shouldPersistResponse } from './saveGuard'
+import {
+  shouldPersistResponse,
+  needsNotesUpdateForChoiceChange,
+} from './saveGuard'
 type Category = {
   name: string
   steps: FismaQuestion[]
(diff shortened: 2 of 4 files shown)


MackOverflow (approved):
clean, looks great

Issue #395, opened 2026-06-16
OpDivGrantModal: set OpDiv grants in one batch call (remove per-grant write loop)


danielbowne:
## Workflow

OpDivGrantModal grants and revokes a user's OpDiv memberships by firing one API call per OpDiv (a per-item write loop). Once the backend batch endpoint lands, send the full desired grant set in a single request, removing the write-side N+1.

This was deferred from the users-list read N+1 fix (UserTable, which now reads `assignedopdivids` inline from the list response). Lower severity than that read path: user-triggered, small N, and bounded by the connection pool.

## Acceptance Criteria

- [ ] OpDivGrantModal saves all grant add/remove changes in one request via the new batch endpoint
- [ ] Resulting grants and the post-save `identity_provider` refresh are unchanged from today
- [ ] No per-OpDiv request loop remains in the modal

## Depends on

Backend batch endpoint: CMS-Enterprise/ztmf#345 (paired)

PR #459, opened 2026-07-08
Datacenter environment vocabulary + extended system fields (ztmf#392)

### LINKED ISSUE #456: HHS metadata edits in the system detail edit view are silently discarded on save
## Description

On the system detail page, HHS-wide admins can edit the HHS Metadata fields (HVA, FIPS, System Type, Cloud System, Cloud Service Model, Cloud Vendor, System Operator, GOCO/COCO/GOGO, System Owner, System Owner Email, Legacy) in the edit view — the inputs render enabled for `hasUnscopedRead` users. But `handleSave` in `src/views/SystemDetailPage/SystemDetailPage.tsx` sends a fixed field list on the PUT that omits all 11 HHS keys, so those edits never reach the backend. The optimistic local-state update right after the PUT writes the edited values onto the on-screen system, so the change appears to save — until a refetch or reload, when the fields revert to their stored values. These fields can be set at create time (the create POST includes them) but cannot be edited afterward.

## Steps to Reproduce

1. Sign in as an HHS-wide admin (a role with `hasUnscopedRead`, e.g. OWNER).
2. Open a system's detail page and click Edit.
3. Change any HHS Metadata field (e.g., System Owner Email).
4. Click Save — a success toast appears and the new value shows on screen.
5. Reload the page (or trigger a systems-list refetch).
6. The HHS field reverts to its previous stored value.

## Expected Behavior

Editing an HHS Metadata field as an HHS-wide admin and saving persists the change to the backend and survives a reload.

## Actual Behavior

The save succeeds visually (success toast + optimistic update), but the HHS fields are omitted from the PUT body, so the backend never receives them. The edit is silently lost on the next refetch or reload.

## Environment

ztmf-ui `main` (React SPA), any browser. Present since #451 merged the HHS metadata feature.

## Acceptance Criteria

- [ ] `handleSave` in `SystemDetailPage.tsx` includes the 11 HHS keys in the PUT body for HHS-wide admins, mirroring the create path's `HHS_METADATA_KEYS` handling.
- [ ] An HHS-wide admin's edit to an HHS field persists across a reload.
- [ ] Non-HHS admins remain unable to edit these fields (existing role gate unchanged).
- [ ] Test added/updated to cover HHS field persistence on edit.
- [ ] No regression in the non-HHS fields already saved by `handleSave`.

## Severity

High - Significantly impacts usage (silent data-integrity failure on an admin edit path)

## Additional Context

The same field-editing path exists in `EditSystemModal`'s edit-mode PUT (currently unreachable — its only caller uses create mode) and should be fixed alongside to avoid a latent trap. Surfaced while rehoming the neutralize-language work (#453 → #455), which renames these fields but does not change this behavior. Originated in the HHS metadata feature (#447), merged via #451.

cc @tarratsco @costacalvin

danielbowne:
Frontend for ztmf#392. Partner of backend PR CMS-Enterprise/ztmf#394 (draft), which adds `GET /api/v1/datacenterenvironments` and populates the extended system fields. Both stay draft and ship together after local testing and approval, so the frontend does not deploy ahead of the endpoint.

## Round 1 — backend-driven environment vocabulary
- New `src/utils/dataCenterEnvironments.ts`: fetch the environment list once, plus pure selectors for the dropdown options and the raw-to-category map. Fetched at the Title layout and shared via outlet context.
- The Add/Edit System dropdowns (modal + detail edit view) render the selectable environments labeled by category instead of the old hardcoded list, which is deleted.
- The questionnaire pillar filter now keys off the resolved scoring category rather than the raw value, so alias environments filter correctly. Raw values are unchanged on save.

## Round 2 — ISSO name, extended fields, legacy-env editing
- Editing a system whose environment is no longer selectable keeps the current value visible as a disabled option, so the required field is not blank.
- The ISSO Name column binds to the backend-resolved `isso_name` (populated for both CMS and HHS systems) and drops the `issoemail.split('@')` derivation, which fixes ztmf-ui#450 (blank for HHS, crashed the sort on null emails). Added a FIPS column.
- The extended metadata fields are now editable across all OpDivs (the org-wide-admin gate is removed) and are always included in the create/edit payload, sending null to leave a value unchanged so imported data is not clobbered. `isso_name` stays display-only (backend-resolved; the form still edits `issoemail`).

## Notes
- Deploy coordination: needs `/api/v1/datacenterenvironments` and the new columns live in the target environment before this is marked ready. The FIPS column shows an em dash until #394 populates it.
- Extended fields render as free text for now; validated select boxes and data cleanup are tracked separately in ztmf#395.
- Clearing a populated extended field is a no-op by design (backend writes only non-null values); proper clear/validation lands with #395.

## Test
- `npx tsc --noEmit` clean; `jest` full suite green (246 passing). New unit tests cover the fetch, the dropdown/category selectors, the legacy-value option, and the category-based pillar filtering.

Closes #456


MackOverflow (approved):
Approving the code, but I think we have a problem on 394 thats causing the jest suite to break, if I am correct that needs to be fixed before merge. 

Reviewed the frontend changes. Typecheck clean and the new util tests (fetch, dropdown/category selectors, legacy-value option, category pillar filter) pass locally — 19/19. The full jest suite can't run due to a pre-existing jsdom/@tootallnate/once bootstrap break on main (unrelated to this PR; no jest CI job covers it). Approving the code, contingent on product confirming that cross-OpDiv extended-metadata editing is the intended behavior now that the hasUnscopedRead gate is removed — the backend ([#394](https://github.com/CMS-Enterprise/ztmf/pull/394)) does not re-gate it.

PR #461, opened 2026-07-08
feat(datacall): sort by deadline; deadline-driven Current/Latest chip (#393)

voidspooks:
Part of CMS-Enterprise/ztmf#393 (the frontend half; paired with the backend PR CMS-Enterprise/ztmf#396).

## Problem
The datacall picker sorted by `datacallid` descending, so a re-imported historical call (higher serial id, older deadline) would out-sort the real current call and become the default selection once the HHS historical load lands.

## Changes
- `Title.tsx`: sort datacalls by `deadline` descending (`datacallid` only as a tiebreak); `sorted[0]` — the default selection and the badged call — is now the furthest-out deadline. All downstream consumers read this via Outlet context, so they inherit the fix.
- `ScoreDiffModal.tsx`: same deadline sort; the default "from" is now the next-older call by list order rather than by next-lower `datacallid`.
- Chip on the latest call now reads **"Current"** only while it is open and **"Latest"** once its deadline has passed (was always "Current").
- Added a `ScoreDiffModal` test asserting the furthest-out deadline is treated as latest even when another call has a higher `datacallid`.

## Testing
- `tsc --noEmit` and `eslint` clean; full changed-file jest suite green (incl. the new regression test).
- Verified end-to-end in the running local stack (vite dev server + dev API rebuilt from the backend branch, driven with Playwright): with a high-`datacallid`/past-deadline call injected, the picker defaults to the open furthest-out call badged "Current" and sorts the historical call into its deadline position; when the furthest-out call's deadline is in the past, its badge reads "Latest".

Refs CMS-Enterprise/ztmf#393

### DIFF
--- a/src/components/ScoreDiffModal/ScoreDiffModal.test.tsx
+++ b/src/components/ScoreDiffModal/ScoreDiffModal.test.tsx
@@ -183,6 +183,48 @@ describe('ScoreDiffModal', () => {
     expect(screen.getByLabelText('From datacall')).toHaveValue('FY2024')
   })
 
+  // #393 regression: a re-imported historical call can carry a higher
+  // datacallid than the real current call, so "latest" must follow the
+  // furthest-out deadline, not the id.
+  it('treats the furthest-out deadline as latest even when another call has a higher datacallid', async () => {
+    const OUT_OF_ORDER: datacall[] = [
+      // highest id but an already-passed deadline (the historical hijacker)
+      {
+        datacallid: 7,
+        datacall: 'FY23 ZTM',
+        datecreated: '2026-01-01T00:00:00Z',
+        deadline: '2023-06-30T00:00:00Z',
+      },
+      // lower id but the furthest-out deadline (the real current call)
+      {
+        datacallid: 6,
+        datacall: 'FY2025 Q3',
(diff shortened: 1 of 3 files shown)


danielbowne:
@voidspooks nice fix, this nails the #393 ordering. Reviewed it and it holds up: the deadline-first comparator with datacallid as a tiebreak is consistent across Title and ScoreDiffModal, latest derives from the top of the sorted list in both, and the from/to defaults line up with the new order. Your regression test covers the exact case (higher id but a passed deadline must not win). Typecheck clean, tests green.

Since this came from a fork it can't run our Snyk and deploy checks, so I rehomed it onto an in-repo branch as #462 with your commit and authorship intact, rebased onto current main. I added one commit on top that pulls the deadline comparator into a small shared util and hardens the empty/malformed deadline case, so a bad value can't fall back to the id tiebreak and quietly reintroduce the bug. Closing this in favor of #462.

One heads up, not on you: this fixes the frontend selectors, but the backend `/datacalls/latest` still orders by `datacallid DESC` and QuestionnareModal reads from it, so we still need a companion backend change (deadline-ordered `FindLatestDataCall` plus an Emberfall case) to fully close #393. I'll take that piece.

Thanks for the contribution.

PR #462, opened 2026-07-08
feat(datacall): sort by deadline; deadline-driven Current/Latest chip (rehome of #461)

danielbowne:
Rehome of #461 by @voidspooks so the Snyk and deploy checks can run (fork PRs cannot reach the org secrets). His commit is preserved with authorship intact, rebased onto current main.

One follow-up commit on top hardens the deadline comparator: it is extracted into a shared, unit-tested `sortDatacallsByDeadline` util, and an empty or malformed deadline now sinks to the bottom instead of falling back to the datacallid tiebreak (which would quietly reintroduce the #393 bug). Local: typecheck clean, full jest suite green.

Note: this fixes the frontend selectors (Title + ScoreDiffModal). The backend `/datacalls/latest` endpoint still orders by `datacallid DESC` and QuestionnareModal consumes it, so a companion backend change (deadline-ordered FindLatestDataCall + an Emberfall case) is needed to fully close #393.

Original description below.

---

Part of CMS-Enterprise/ztmf#393 (the frontend half; paired with the backend PR CMS-Enterprise/ztmf#396).

## Problem
The datacall picker sorted by `datacallid` descending, so a re-imported historical call (higher serial id, older deadline) would out-sort the real current call and become the default selection once the HHS historical load lands.

## Changes
- `Title.tsx`: sort datacalls by `deadline` descending (`datacallid` only as a tiebreak); `sorted[0]` — the default selection and the badged call — is now the furthest-out deadline. All downstream consumers read this via Outlet context, so they inherit the fix.
- `ScoreDiffModal.tsx`: same deadline sort; the default "from" is now the next-older call by list order rather than by next-lower `datacallid`.
- Chip on the latest call now reads **"Current"** only while it is open and **"Latest"** once its deadline has passed (was always "Current").
- Added a `ScoreDiffModal` test asserting the furthest-out deadline is treated as latest even when another call has a higher `datacallid`.

## Testing
- `tsc --noEmit` and `eslint` clean; full changed-file jest suite green (incl. the new regression test).
- Verified end-to-end in the running local stack (vite dev server + dev API rebuilt from the backend branch, driven with Playwright): with a high-`datacallid`/past-deadline call injected, the picker defaults to the open furthest-out call badged "Current" and sorts the historical call into its deadline position; when the furthest-out call's deadline is in the past, its badge reads "Latest".

Refs CMS-Enterprise/ztmf#393

### DIFF
--- a/src/components/ScoreDiffModal/ScoreDiffModal.test.tsx
+++ b/src/components/ScoreDiffModal/ScoreDiffModal.test.tsx
@@ -183,6 +183,48 @@ describe('ScoreDiffModal', () => {
     expect(screen.getByLabelText('From datacall')).toHaveValue('FY2024')
   })
 
+  // #393 regression: a re-imported historical call can carry a higher
+  // datacallid than the real current call, so "latest" must follow the
+  // furthest-out deadline, not the id.
+  it('treats the furthest-out deadline as latest even when another call has a higher datacallid', async () => {
+    const OUT_OF_ORDER: datacall[] = [
+      // highest id but an already-passed deadline (the historical hijacker)
+      {
+        datacallid: 7,
+        datacall: 'FY23 ZTM',
+        datecreated: '2026-01-01T00:00:00Z',
+        deadline: '2023-06-30T00:00:00Z',
+      },
+      // lower id but the furthest-out deadline (the real current call)
+      {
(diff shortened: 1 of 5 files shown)


MackOverflow (approved):
looks clean!

PR #464, opened 2026-07-08
add drafts for questionnaire

### LINKED ISSUE #457: Questionnaire: add autosave / draft persistence so in-progress answers survive a crash or accidental navigation
### Feature Type
UI Component

### Workflow Description
Today the questionnaire only persists a response when the user clicks Next or Complete. There is a single `saveResponse()` call, wired to the Next/Complete button (`src/views/QuestionnairePage/QuestionnairePage.tsx`). Typing a radio answer or notes updates local React state only, so the inline validation reacts on change but nothing is written to the backend until the user advances. If the browser crashes, the tab is closed, or the user navigates away (Back and the sidebar open a discard confirm) before advancing, everything entered on that question since the last Next is lost. There is no debounced autosave, no save-on-blur, and no `beforeunload` guard for a tab close or refresh.

For a questionnaire where users write real justification text per question, this is a data-loss risk. Raised by the customer during data-call entry.

Recommended approach (lightest first pass): a debounced on-change save of the active question's answer and notes to a local draft (localStorage) keyed by system + question + datacall, restored on reload and cleared on a successful save. Pair it with a visible save-status indicator (Saved / Unsaved) and a `beforeunload` guard so closing or refreshing the tab warns before losing work. A debounced on-change model is preferred over a fixed-interval timer: it captures the latest edit within a second or two of the user pausing, whereas a timer can lose everything typed since the last tick. A heavier alternative is a debounced backend autosave (PUT /scores on blur), but that adds score-record churn and must respect the changed-answer notes rule, so the local draft is the recommended starting point.

### Acceptance Criteria
- In-progress answer and notes for the active question survive a page reload or browser crash.
- A restored draft is clearly indicated in the UI and clears once the response is saved.
- Closing or refreshing the tab with unsaved changes prompts the user before the work is lost.
- The draft/autosave path does not bypass the existing changed-answer notes-required guard on final save.
- Read-only sessions are unaffected.

### Additional Context
- Save trigger today: a single `saveResponse()` at the Next/Complete button, issuing PUT/POST to `/scores`.
- Related work: the changed-answer notes-required guard added recently is validation only (no persistence) and left a code comment anticipating a future autosave trigger.
- Suggested local-draft key: fismasystemid + questionid + datacallid.

costacalvin:
## Summary
Closes #457 
Implements localStorage-based draft persistence for questionnaire responses. Answers and notes are autosaved locally so a hard reload, crash, or accidental navigation no longer loses in-progress work.

- **Debounced autosave** — 1 second after the user pauses editing, the current answer and notes are written to `localStorage` keyed by `ztmf_draft_{fismasystemid}_{functionid}_{datacallid}`
- **14-day TTL** — stale drafts are evicted automatically on load
- **Draft restoration** — on return to a question, any saved draft is reapplied and a banner prompts the user to save permanently via Next/Complete
- **Auto-clear on save** — draft is removed from `localStorage` after a successful backend save, or when the user clicks Next/Complete with no changes
- **`beforeunload` guard** — browser warns before tab close or hard refresh when there are unsaved edits
- **Read-only sessions excluded** — no drafts are saved, restored, or shown when `isReadOnly` is true; existing stale drafts are eagerly evicted on read-only load
- **Option validation** — a restored draft referencing a `functionoptionid` that no longer exists is silently evicted rather than applied, preventing a stale value from reaching the backend

## Files Changed

- `src/views/QuestionnairePage/draftStore.ts` *(new)* — thin `localStorage` wrapper (`saveDraft`, `loadDraft`, `clearDraft`) with TTL handling; all operations are try/catch-wrapped for private-browsing/quota resilience
- `src/views/QuestionnairePage/draftStore.test.ts` *(new)* — 18 unit tests covering key format, TTL boundary, expiry eviction, malformed JSON, and `localStorage` unavailability
- `src/views/QuestionnairePage/QuestionnairePage.tsx` — debounce effect, restoration logic in `fetchOptions`, draft banner, `beforeunload` effect, and draft eviction on save/navigation

## UI
Original
image

Draft saved after edits
image

Refresh triggers popup
image

Draft restored on reload
image

Changing questions through sidebar brings up confirm modal
image

If confirmed the draft is cleared, and reverted to original state.
image

## Test Plan

- [ ] Open a question, select an answer, type notes — wait 1 second then hard reload (`Cmd+Shift+R`). Banner should show "Draft restored" with fields pre-filled
- [ ] Click Next/Complete after restoring a draft — banner disappears and the response is saved to the server
- [ ] Click Next/Complete with no changes from the server state — draft is cleared, no banner on next visit
- [ ] Restore a draft and attempt to close the tab before clicking Next — browser should show the "Leave site?" dialog
- [ ] Open a question in a closed (read-only) datacall — no draft banner, no autosave, existing drafts for that question are evicted
- [ ] Run unit tests: `npx jest --testPathPattern="draftStore"`

### DIFF
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -508,19 +508,28 @@ export default function QuestionnarePage() {
             !isReadOnly && sys && questionId && datacallID > 0
               ? loadDraft(sys, questionId, datacallID)
               : null
-          if (
-            draft &&
-            choices.some((c) => c.value === draft.selectQuestionOption)
-          ) {
-            choices.forEach(
-              (c) => (c.defaultChecked = c.value === draft.selectQuestionOption)
-            )
-            setSelectQuestionOption(draft.selectQuestionOption)
-            setNotes(draft.notes)
-            setDraftStatus('restored')
+          if (draft) {
+            if (draft.selectQuestionOption === -1) {
+              // Notes-only draft — restore notes without pre-selecting an answer.
+              setNotes(draft.notes)
+              setDraftStatus('restored')
+            } else if (
+              choices.some((c) => c.value === draft.selectQuestionOption)
+            ) {
+              choices.forEach(
+                (c) =>
+                  (c.defaultChecked = c.value === draft.selectQuestionOption)
+              )
+              setSelectQuestionOption(draft.selectQuestionOption)
+              setNotes(draft.notes)
+              setDraftStatus('restored')
+            } else {
+              // Draft references an option that no longer exists — evict it.
+              if (sys && questionId && datacallID > 0)
+                clearDraft(sys, questionId, datacallID)
+              setDraftStatus('idle')
+            }
           } else {
-            if (draft && sys && questionId && datacallID > 0)
-              clearDraft(sys, questionId, datacallID)
             setDraftStatus('idle')
           }
           setOptions(choices)
@@ -553,10 +562,6 @@ export default function QuestionnarePage() {
       if (draftStatusRef.current !== 'idle') setDraftStatus('idle')
       return
     }
-    // Never draft-save when no answer is selected: -1 is the "no answer"
-    // sentinel that shouldPersistResponse rejects, so a draft with -1 could
-    // never be promoted to a real save, leaving a permanent orphan in storage.
-    if (selectQuestionOption === -1) return
     if (selectQuestionOption === initQuestionChoice && notes === initNotes)
       return
     const timer = setTimeout(() => {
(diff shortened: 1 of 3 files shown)


MackOverflow:
@costacalvin really nice! A couple thoughts from Maude: 
- **Notes typed before an answer is selected are never drafted.** The debounce guards `if (selectQuestionOption === -1) return` ([QuestionnairePage.tsx:559](src/views/QuestionnairePage/QuestionnairePage.tsx)). Rationale is sound (a `-1` draft can't be promoted to a real save), but a user who types notes first, then reloads, silently loses that text — which is exactly the "don't lose in-progress work" case this PR targets. Worth confirming this is the intended tradeoff, or restoring notes-only drafts while still blocking their save.
- **`beforeunload` may not prompt on older/legacy browsers.** Line 584 uses only `e.preventDefault()`. Per current spec that's sufficient for Chrome/Firefox, but Safari and older engines historically need `e.returnValue = ''` too. Cheap to add both.
- **Data-at-rest consideration.** Questionnaire notes for a CMS security-posture tool now sit in plaintext `localStorage` for up to 14 days, including on shared/kiosk machines. Read-only eviction helps, but consider whether the retention window and the lack of clearing on logout are acceptable for this data class. I would think we're ok, but worth considering.

costacalvin:
@MackOverflow 

- **Notes typed before an answer is selected are never drafted.**  Added conditional to allow notes with no option selected to be saved as a draft 
- **`beforeunload` may not prompt on older/legacy browsers.** Added the `e.returnValue = ''` to handle older browsers

- **Data-at-rest consideration.** We could tie the drafts to user by adding userID to the key, then on app mount, sweep any draft not belonging to the current user.

MackOverflow:
looks great! Rehoming now!

MackOverflow:
Moving this to #469 — fork PRs can't access the org secrets required for Snyk and the deploy CI gates. Your commits and authorship carry over unchanged. Thanks @costacalvin!

PR #470, opened 2026-07-08
feat(system-detail): Target Maturity Level card (#398 MVP)

voidspooks:
Frontend half of CMS-Enterprise/ztmf#398 (paired with backend PR CMS-Enterprise/ztmf#400 — merge that first; this UI calls its `PUT /fismasystems/{id}/target-maturity` endpoint).

## What

A **Target Maturity Level** card on the system detail page (right column, between Data Lake Export and Organization) showing the risk-based target tier and its justification, editable through the existing page-level Edit/Save flow.

## Behavior

- Options: **2 — Initial / 3 — Advanced (default) / 4 — Optimal** (tier NAME is the stored value; numerals are display-only; Traditional deliberately not offered).
- **No stored default**: a system with no asserted target renders the Advanced default explicitly labeled (`(default)` chip + "no target has been set" caption); the DB stays NULL until someone saves.
- **Single edit flow**: no card-level buttons — the page Edit unlocks the card, page Save persists it. The target PUT fires **only when the draft changed**, so untouched saves record no spurious audit event. Changing the tier requires a justification before Save enables; Cancel prompts on a dirty target.
- **ISSO/ISSM support**: the page Edit button now shows for system-scoped users (new `isSystemScoped` helper) — for them only the target card unlocks; the system form stays admin-only. Display gating is by tier per the app convention (the backend scopes an ISSO's system list and re-checks assignment on write).
- Fields use `variant="standard"` matching `SystemDetailEditView` (the CMS design-system CSS breaks MUI's outlined floating label).

## Test plan

- [x] `tsc --noEmit`, `eslint` clean
- [x] Jest: card read/edit states, option vocabulary, required + over-limit justification (plus full SystemDetailPage suite green)
- [x] Validated in the local stack against the backend branch: admin full-edit, ISSO target-only edit on an assigned system, no Edit button on unassigned systems, persistence across refresh

## Screenshots

Screenshot 2026-07-08 at 4 53 24 PM
Screenshot 2026-07-08 at 4 54 41 PM

Refs CMS-Enterprise/ztmf#398

PR #471, opened 2026-07-08
feat(system-detail): Target Maturity Level card (#398 MVP)

MackOverflow:
> **Note:** In-repo copy of #470 by @voidspooks. Moved from the fork so Snyk
>
> Refs #470

---

Frontend half of CMS-Enterprise/ztmf#398 (paired with backend PR CMS-Enterprise/ztmf#401 — merge that first; this UI calls its `PUT /fismasystems/{id}/target-maturity` endpoint).

## What

A **Target Maturity Level** card on the system detail page (right column, between Data Lake Export and Organization) showing the risk-based target tier and its justification, editable through the existing page-level Edit/Save flow.

## Behavior

- Options: **2 — Initial / 3 — Advanced (default) / 4 — Optimal** (tier NAME is the stored value; numerals are display-only; Traditional deliberately not offered).
- **No stored default**: a system with no asserted target renders the Advanced default explicitly labeled (`(default)` chip + "no target has been set" caption); the DB stays NULL until someone saves.
- **Single edit flow**: no card-level buttons — the page Edit unlocks the card, page Save persists it. The target PUT fires **only when the draft changed**, so untouched saves record no spurious audit event. Changing the tier requires a justification before Save enables; Cancel prompts on a dirty target.
- **ISSO/ISSM support**: the page Edit button now shows for system-scoped users (new `isSystemScoped` helper) — for them only the target card unlocks; the system form stays admin-only. Display gating is by tier per the app convention (the backend scopes an ISSO's system list and re-checks assignment on write).
- Fields use `variant="standard"` matching `SystemDetailEditView` (the CMS design-system CSS breaks MUI's outlined floating label).

## Test plan

- [x] `tsc --noEmit`, `eslint` clean
- [x] Jest: card read/edit states, option vocabulary, required + over-limit justification (plus full SystemDetailPage suite green)
- [x] Validated in the local stack against the backend branch: admin full-edit, ISSO target-only edit on an assigned system, no Edit button on unassigned systems, persistence across refresh

## Screenshots

Screenshot 2026-07-08 at 4 53 24 PM
Screenshot 2026-07-08 at 4 54 41 PM

Refs CMS-Enterprise/ztmf#398

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -103,6 +103,10 @@ export type FismaSystemType = {
   system_owner?: string | null
   system_owner_email?: string | null
   legacy?: string | null
+  // Risk-based target maturity (ztmf#398). null = no target asserted yet;
+  // the UI presents the Advanced default.
+  target_maturity_tier?: string | null
+  target_maturity_justification?: string | null
 }
 export type FismaSystems = {
   fismaSystems: FismaSystemType[]
--- a/src/utils/userRoles.ts
+++ b/src/utils/userRoles.ts
@@ -86,6 +86,15 @@ export const hasAdminRead = (user: UserLike): boolean =>
 export const hasUnscopedRead = (user: UserLike): boolean =>
   !!user && UNSCOPED_READ_ROLES.has(user.role as UserRole)
 
+// System-scoped tiers (ISSO / ISSM) - the users whose write scope is their
+// per-system assignment rather than an OpDiv or admin tier. Like the other
+// gates here this is display-only: the backend narrows the fismaSystems list
+// to the user's assignments and re-checks IsAssignedFismaSystem on write, so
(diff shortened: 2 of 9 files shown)


jsos3-cms (dismissed):
Sent along some non-blocking nits for review. Adding my approval for the PR, and a re-review if it gets cleared.

MackOverflow:
Good fixes by @jsos3-cms 

Trim mismatch — justificationTooLong now trims before comparing to the max, matching the page's isTargetValid gate (and mirroring the [#401](https://github.com/CMS-Enterprise/ztmf/pull/401) API fix).
PUT reorder — target-maturity PUT now fires before the full-system PUT, so a failure leaves nothing silently persisted.

jsos3-cms (approved):
Thank you so much Mack!

PR #472, opened 2026-07-08
Enhancement/encrypt drafts

### LINKED ISSUE #399: Encrypt Draft text of freeform box at rest
### Feature Type

General Feature

### Workflow Description

ztmf#496/ztmf#494 introduces the ability for ISSO users to capture incomplete/drafted responses to free text boxes in current data calls. Some of this information might include sensitive or classified system information, organizational context, or implementation details. 

This feature request asks to encrypt and decrypt these drafts when stored in the users local browser. 

### Acceptance Criteria

- [ ] A draft text during an open data call is saved in local browser state
- [ ] Draft text is not human readable
- [ ] There is not a UI or API type indicator that may hint to a malicious actor what encryption protocol is used
- [ ] Can we implement a PQC compliant encryption here? (Explain if out of scope, might be a stretch to implement)

### Additional Context

_No response_

costacalvin:
Builds off of https://github.com/CMS-Enterprise/ztmf-ui/pull/469 and closes https://github.com/CMS-Enterprise/ztmf/issues/399

IN PROGRESS

### DIFF
--- a/config/jest/setupTests.ts
+++ b/config/jest/setupTests.ts
@@ -1,7 +1,15 @@
 import '@testing-library/jest-dom'
 import 'jest-fetch-mock'
 import { TextEncoder, TextDecoder } from 'util'
+import { webcrypto } from 'crypto'
 
 // Polyfill for React Router 7 which requires TextEncoder/TextDecoder
 global.TextEncoder = TextEncoder
 global.TextDecoder = TextDecoder as typeof global.TextDecoder
+
+// Polyfill Web Crypto API — JSDOM does not expose crypto.subtle
+Object.defineProperty(globalThis, 'crypto', {
+  value: webcrypto,
+  configurable: true,
+  writable: true,
+})
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -245,7 +245,7 @@ export default function QuestionnarePage() {
 
   const clearCurrentDraft = () => {
     if (system && questionId && datacallID > 0) {
-      clearDraft(system, questionId, datacallID)
+      clearDraft(userInfo.userid, system, questionId, datacallID)
     }
     setDraftStatus('idle')
   }
@@ -502,11 +502,12 @@ export default function QuestionnarePage() {
           // server-side values set above. Skipped for read-only sessions.
           const sys = systemRef.current
           // Eagerly evict any stale draft when the session is now read-only.
+          const uid = userInfo.userid
           if (isReadOnly && sys && questionId && datacallID > 0)
-            clearDraft(sys, questionId, datacallID)
+            clearDraft(uid, sys, questionId, datacallID)
           const draft =
             !isReadOnly && sys && questionId && datacallID > 0
-              ? loadDraft(sys, questionId, datacallID)
+              ? await loadDraft(uid, sys, questionId, datacallID)
               : null
           if (draft) {
             if (draft.selectQuestionOption === -1) {
@@ -526,7 +527,7 @@ export default function QuestionnarePage() {
             } else {
               // Draft references an option that no longer exists — evict it.
               if (sys && questionId && datacallID > 0)
-                clearDraft(sys, questionId, datacallID)
+                clearDraft(uid, sys, questionId, datacallID)
               setDraftStatus('idle')
             }
           } else {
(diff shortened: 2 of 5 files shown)


jsos3-cms:
### #472 "encrypt drafts" — review findings

Findings from a pre-merge review of #472 (AES-256-GCM encryption of questionnaire drafts, closes ztmf#399). No rehome yet — this is for the work to continue on #472.

#### Truly blocking

**1. The encryption doesn't yet deliver the confidentiality the ticket asks for.**
The AES key is derived only from the user's `userid` — a UUID that isn't secret; `GET /users/current` returns it in plaintext — plus a fixed, app-wide salt. The threat this feature targets is someone reading drafts "at rest" in the browser. But anyone who can read the browser profile's `localStorage` can also read that same `userid` from the same profile, then re-derive the key with the app's own shipped client code and decrypt every draft. So the current design achieves "not human-readable at a glance in DevTools," but not ztmf#399's actual goal of protecting sensitive information against device access.

This is a design decision, not a knob to tune. Two honest ways forward:
- **Introduce a genuine secret into the key derivation** — e.g. a value bound to the authenticated session, or a randomly generated per-device key stored as a *non-extractable* `CryptoKey` in IndexedDB (so an attacker must operate inside the live browser rather than doing an offline decrypt). Note the inherent limit: anything that must survive a hard reload without re-authenticating has to live on disk, and disk-level access reaches it too — client-only crypto can raise the bar but can't fully solve "immune to a compromised device."
- **Or reframe the ticket's scope** to what client-side encryption can honestly deliver ("obfuscated at rest, not readable at a glance"), so no one signs off on it as protecting classified/sensitive data when it doesn't.

Either is fine — but the current framing (satisfies #399's security intent) shouldn't ship, because it creates false assurance.

**2. A failed save still tells the user "Draft saved."**
`saveDraft` swallows every error internally, so if the write fails — storage quota hit, Safari private browsing, a non-secure context — the promise still resolves and the UI shows "Draft saved," with nothing logged. That's a silent data-loss path, and it fails in exactly the storage-failure case the feature is supposed to guard against. `saveDraft` should report real success vs. failure, the UI should show a distinct "not saved" state, and the swallowed error should at least be logged.

#### Non-blocking (hardening — worth doing, none gate merge)

- **Static salt** → use a random per-entry salt stored next to the IV/ciphertext.
- **PBKDF2 iterations** are 100k; current OWASP guidance is 600k+ — only matters once the derivation is reworked with a real secret.
- **Autosave vs. explicit clear race:** an in-flight debounced save can land *after* a "Next"/clear and resurrect a just-cleared draft. A generation counter or in-flight tracking closes it.
- **`clearStaleUserDrafts` is misnamed** (it purges other users' drafts, with no TTL check) and leans on an implicit "no user ID is a prefix of another" invariant. Safe today (UUIDs), but rename it and drop the fragile assumption.
- **No shape check on decrypted JSON** — a future schema change could surface `undefined` fields; add a lightweight shape/version check.
- **`localStorage` key names embed the raw `userid`** — visible in DevTools on a shared machine; consider hashing it into the key.

#### Sound as-is (for balance)

AES-256-GCM via Web Crypto (not hand-rolled), a fresh 12-byte IV per encryption (never reused), correct base64 round-tripping, decrypt that fails closed (returns null, no partial-plaintext leak), and a genuinely solid `draftStore` unit-test suite (wrong-key isolation, TTL boundary, quota exceptions, cross-user isolation).

#### On post-quantum (brief)

AES-256-GCM is already the right, quantum-appropriate choice here — no change needed and nothing to add. Quantum computers only *halve* symmetric strength (Grover), so AES-256 keeps ~128-bit security; NSA's CNSA 2.0 keeps AES-256 for the post-quantum era. The quantum-vulnerable class is *asymmetric* key exchange and signatures (RSA/ECC, broken by Shor) — which this feature doesn't use, so there's nothing here to post-quantum-harden. Adding PQC would bring no value on this feature and isn't practical client-side anyway (the browser's Web Crypto has no ML-KEM/ML-DSA, and the JS/WASM libraries aren't FIPS-validated). The important point: #472's real weakness is the *absence of a genuine secret* (blocking item 1) — post-quantum resistance is orthogonal to that and wouldn't fix it.

## Where this leaves us

No rehome. Keep iterating on #472 (still a draft); #469 continues in parallel. Once item 1's direction is settled (add a real secret vs. reframe the scope) and item 2 is fixed, it's ready to rehome for CI and merge.

jsos3-cms:
476 is merged, this is closed

Issue #475, opened 2026-07-09
Encrypted questionnaire drafts: confidentiality hardening and persistence fixes


jsos3-cms:
## Description

Successor to CMS-Enterprise/ztmf#399 ("Encrypt Draft text of freeform box at rest"), moved into ztmf-ui where the feature lives. The initial encrypted-drafts feature shipped via the #472 rehome (questionnaire draft persistence from #469 plus AES-256-GCM encryption from #472). ztmf#399's stated acceptance criteria are met (draft saved locally, not human readable, no protocol hint), and its post-quantum question is answered below. This ticket tracks the remaining work to reach real confidentiality and to fix the persistence blockers found in review.

## Post-quantum question (from ztmf#399)

AES-256-GCM is already quantum-appropriate for data at rest. Quantum computers only halve symmetric strength (Grover), so AES-256 keeps about 128-bit security, and NSA CNSA 2.0 keeps AES-256 for the post-quantum era. The post-quantum standards (ML-KEM, ML-DSA) target asymmetric key exchange and signatures, which this feature does not use, so PQC is out of scope here. The real gap is the absence of a genuine secret in the key derivation (first blocking item below), which PQC would not address.

## Blocking items

Encryption and confidentiality:

- [ ] **Derive the key from a real secret.** The AES key currently derives only from the user's `userid`, a non-secret UUID that `GET /users/current` returns in plaintext, plus a fixed salt. Anyone who can read the browser profile can re-derive the key and decrypt drafts, so the feature obfuscates rather than protects against device access. Bind the key to the authenticated session, hold a non-extractable per-device key in IndexedDB, or encrypt server-side.
- [ ] **Report real save success or failure.** `saveDraft` swallows `setItem` errors, and the UI shows "Draft saved" even when the write failed (quota, private browsing), losing data silently.

Persistence correctness:

- [ ] **Validate the loaded draft shape.** `loadDraft` casts without a runtime check, so a malformed entry (`notes: undefined`) reaches render, throws on `notes.length`, and blanks the questionnaire. Evict on a mismatch.
- [ ] **Scope draft keys to the user.** The key omits the user id, so on a shared browser one user's draft restores into another's session and can submit under them. Encryption does not fix this. Add `userid` to the key.
- [ ] **Clear the draft when an answer returns to its original value.** Reverting A to B to A leaves the B draft in storage, where it can resurface within the 14-day TTL and overwrite a correct answer.

## Non-blocking hardening

[...]

PR #476, opened 2026-07-09
feat(questionnaire): encrypt saved drafts (AES-256-GCM)

jsos3-cms:
> **Note:** In-repo copy of #472 by @costacalvin, moved off the fork so the Snyk and deploy CI gates can run, which need org secrets unavailable to fork PRs. Commits and authorship are unchanged.
>
> Refs #472

---

## Summary

Adds an initial encrypted-drafts feature for the questionnaire. Drafts persist in the browser (from #469) and are encrypted at rest with AES-256-GCM via the Web Crypto API (from #472). This is a first cut: it obfuscates drafts at rest, and the confidentiality hardening plus persistence fixes are tracked in #475.

Delivers the initial version requested in CMS-Enterprise/ztmf#399. That issue closes in favor of the ztmf-ui successor #475, which tracks the remaining confidentiality hardening and persistence fixes, when this merges.

## Supersedes #469

This branch carries #469's commits (questionnaire draft persistence, @MackOverflow) with the AES-256-GCM layer (@costacalvin) on top, confirmed against git history (#469's head is an ancestor of this branch). Close #469 in its favor once this is accepted (it is a PR, so the `Closes` keyword will not auto-close it; close it by hand).

## What's included

- **Questionnaire draft persistence**, from #469, @MackOverflow.
- **AES-256-GCM draft encryption** via `crypto.subtle`, from #472, @costacalvin: new `draftStore.ts`, wiring in `QuestionnairePage.tsx` and `Title.tsx`, jest crypto setup in `config/jest/setupTests.ts`, and `draftStore.test.ts`.

## Known limitations (tracked in #475)

Pre-merge review found blockers that ship in this initial version. They are tracked in #475 rather than gating the first cut:

- Key derivation uses the user id (a non-secret) plus a fixed salt, so the encryption obfuscates rather than protects against device access.
- A failed save still reports "Draft saved," which can lose data silently.
- A malformed stored draft can crash the questionnaire view.
- Draft keys omit the user, so a shared browser can leak drafts across users.
- Reverting an answer to its original value does not clear the saved draft.

## Test plan

- [x] `yarn typecheck` clean
- [ ] `yarn test` green (includes `draftStore.test.ts`)
- [ ] Enter answers and notes, reload, the draft restores
- [ ] Inspect the stored draft in DevTools, it shows ciphertext, not plaintext
- [ ] Snyk and deploy CI gates pass (the reason for the rehome)

## Notes

- Authorship preserved. The ref-based rehome points at the exact commit stack, so both authors' commits are unchanged.
- The original fork #472 stays a draft and closes when this merges.

### DIFF
--- a/config/jest/setupTests.ts
+++ b/config/jest/setupTests.ts
@@ -1,7 +1,15 @@
 import '@testing-library/jest-dom'
 import 'jest-fetch-mock'
 import { TextEncoder, TextDecoder } from 'util'
+import { webcrypto } from 'crypto'
 
 // Polyfill for React Router 7 which requires TextEncoder/TextDecoder
 global.TextEncoder = TextEncoder
 global.TextDecoder = TextDecoder as typeof global.TextDecoder
+
+// Polyfill Web Crypto API — JSDOM does not expose crypto.subtle
+Object.defineProperty(globalThis, 'crypto', {
+  value: webcrypto,
+  configurable: true,
+  writable: true,
+})
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -245,7 +245,7 @@ export default function QuestionnarePage() {
 
   const clearCurrentDraft = () => {
     if (system && questionId && datacallID > 0) {
-      clearDraft(system, questionId, datacallID)
+      clearDraft(userInfo.userid, system, questionId, datacallID)
     }
     setDraftStatus('idle')
   }
@@ -502,11 +502,12 @@ export default function QuestionnarePage() {
           // server-side values set above. Skipped for read-only sessions.
           const sys = systemRef.current
           // Eagerly evict any stale draft when the session is now read-only.
+          const uid = userInfo.userid
           if (isReadOnly && sys && questionId && datacallID > 0)
-            clearDraft(sys, questionId, datacallID)
+            clearDraft(uid, sys, questionId, datacallID)
           const draft =
             !isReadOnly && sys && questionId && datacallID > 0
-              ? loadDraft(sys, questionId, datacallID)
+              ? await loadDraft(uid, sys, questionId, datacallID)
               : null
           if (draft) {
             if (draft.selectQuestionOption === -1) {
@@ -526,7 +527,7 @@ export default function QuestionnarePage() {
             } else {
               // Draft references an option that no longer exists — evict it.
               if (sys && questionId && datacallID > 0)
-                clearDraft(sys, questionId, datacallID)
+                clearDraft(uid, sys, questionId, datacallID)
               setDraftStatus('idle')
             }
           } else {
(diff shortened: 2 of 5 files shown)


MackOverflow:
Jono, this looks good. Here's a draft comment from your perspective, replying to Jono, folding in both notes — the MVP/scope confirmation and the encryption clarification:

> Confirming the plan: #476 rehomes #472 unchanged (same head as when you reviewed it) — it encapsulates both #469's draft persistence and #472's AES-256-GCM layer as the MVP cut. Per #475 (which you opened), items 1 (key derivation) and 2 (silent save failure) are tracked there rather than blocking this merge.
>
> Worth being precise on scope for anyone reading #399/#475 later: drafts *are* genuinely encrypted at rest — AES-256-GCM via Web Crypto, fresh IV per write, not plaintext, not just base64. The gap is the key, not the cipher: it derives from `userid` plus one static app-wide salt, and `userid` is readable in plaintext from the same browser profile that holds the ciphertext. So it defeats a glance at localStorage but not device access, which is #399's actual threat model. That's #475 item 1, and closing #399 in favor of #475 makes sense as long as that caveat travels with it rather than reading as "encryption done."

I think for MVP this is fine to move forward, but we'll want to revisit the 'spirit' of 399 again to ensure that we're answering the mail. This is already a big improvement, and if we can add some device access protections it can be even stronger.

MackOverflow (approved):
Good approach, 399 should be revisited, but this is a big improvement. two silent failures (non blocking) noted in body can be revisted.

MackOverflow:
Two things I'd like to sruface before merging, not blockers per se, but worth looking at. Both from reviewing the diff:

**1. Failed draft saves still report "Draft saved."**
In `draftStore.ts`, `saveDraft` catches all internal errors and resolves normally (never rejects):
```ts
} catch {
  // localStorage unavailable or crypto unsupported — degrade silently
}
```
In `QuestionnairePage.tsx`, the debounce effect does:
```ts
saveDraft(...).then(() => {
  if (draftStatusRef.current !== 'restored') setDraftStatus('saved')
})
```
Since `saveDraft` never rejects, the "saved" status fires unconditionally — a full `localStorage` quota, disabled storage, or a `crypto.subtle` failure will still show the user "Draft saved" with nothing actually persisted. This is called out in the PR description as a known limitation deferred to #475, but given it's a silent-data-loss path (not cosmetic), do we consider either fixing here or explicitly prioritized as the first item in #475 rather than bundled with the lower-severity items?

**2. Test-plan checkboxes are unchecked on the CI-enabling PR.**
`yarn test`, "enter answers and notes, reload, draft restores," "inspect stored draft shows ciphertext," and "Snyk/deploy CI gates pass" are all unchecked in the PR body — but this PR's stated purpose is specifically to unblock those CI gates. Can we get confirmation these were run (and the boxes checked, or an explanation of what's still pending) before merge?

MackOverflow (approved):
approving but with two nits, one large, one small. In comments.

costacalvin:
@MackOverflow 
1. This is definitely an issue and is being handled in #475, the PR is currently in draft.
2. I will update the PR body checkboxes. Let me just re-test it real quick 👍 Just remembered I dont have access haha, but just retested these three and all passed. Synk tests also passed looking at the check results.

- [x]  yarn test green (includes draftStore.test.ts)
- [x]  Enter answers and notes, reload, the draft restores
- [x]  Inspect the stored draft in DevTools, it shows ciphertext, not plaintext
- [x] Synk tests pass according to the check results

PR #480, opened 2026-07-09
Data Call Progress column on the dashboard

tarratsco:
### Summary

Adds a **Data Call Progress** column to the dashboard system list so OpDiv Admins can see, at a glance, which systems have actually updated their questionnaires this data call. This is the frontend half of `ztmf#299`; it consumes the new `GET /api/v1/scores/progress` endpoint from the companion backend PR, `CMS-Enterprise/ztmf#404`.

### What it shows

For each system, a fraction plus a status chip:

* `12/41 · Updated` — 12 questions genuinely edited this cycle
* `0/41 · Not updated` — answers carried over from last data call, nobody has touched it
* `—` — no progress data (fetch failed, or the system isn't covered)

Each cell carries a tooltip with the last-activity time. The column **sorts not-updated systems first**, so an admin with hundreds of systems sees the ones needing a nudge at the top without scrolling.

### Why "Not updated" isn't "0 answered"

The backend pre-populates every system's answers from the previous data call, so "has answers" is true for everyone on day one. The endpoint counts only answers *genuinely edited this cycle* (those with an audit event), so a fully carried-over questionnaire correctly reads `0/N · Not updated` even though it has a complete set of answers. The column reflects that faithfully — see `CMS-Enterprise/ztmf#404` for the mechanic.

### How it's wired

* `Home.tsx` fetches `/scores/progress?datacallid=${act existing aggregate fetch, on the same `activeDataCallId` dependency — so switching the active data call refetches progress automatically.
* The progress fetch is **independent of the score fetch map and the column falls back to em-dashes, without blocking score display or the page's loading gate.**
* Sort logic and tooltip text live in `progressHelpers.ts` so `progressColumn.tsx` stays a component-only module (FastRefresh).

### Testing

`progressColumn.test.tsx` (11 tests) covers:

* Triage sort ordering — not-updated first, then partial by fraction, complete last, unknowns last, with a divide-by-zero guard for systems with no applicable questions
* Tooltip states — last-activity line, explicit "no updates," missing-data, and unparseable-timestamp fallback
* Cell rendering — the `Updated` / `Not updated` / em-dash layouts, specifically checking the core `ztmf#299` case (a carried-over system renders `0/N · Not updated`)

### Files

| Area | File |
| --- | --- |
| Types | `src/types.ts` (`ScoreProgress`, `FismaTableProps.progress`) |
| Cell component | `src/views/FismaTable/progressColumn.tsx` |
| Sort + tooltip helpers | `src/views/FismaTable/progressHelpers.ts` |
| Column wiring | `src/views/FismaTable/FismaTable.tsx` |
| Data fetch | `src/views/Home/Home.tsx` |
| Tests | `src/views/FismaTable/progressColumn.test.tsx` |

2 commits, +242 / 6 files.

### Notes for reviewers

* Built in main's existing table idiom (plain MUI `Chip` layout) rather than reaching for the redesign's design-system components — this ships on `main` ahead of the redesign, which will restyle the column when it absorbs this commit.
* `progress` is an optional prop, so the table degrades gracefully if the progress array is unavailable; score display never depends on it.
* **Requires CMS-Enterprise/ztmf#404** to be deployed — until then the column renders all em-dashes.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -240,6 +240,19 @@ export type SystemScoreEntry = {
   tier?: ScoreTier
 }
 
+// One system's questionnaire progress within a data call, as returned by
+// GET /scores/progress. "Updated" counts answers genuinely touched this
+// cycle - answers pre-populated from the previous data call do not count
+// until a user saves them, so a carried-over untouched questionnaire reads
+// as not updated (ztmf#299).
+export type ScoreProgress = {
+  fismasystemid: number
+  questionsexpected: number
+  questionsupdated: number
+  lastupdatedat?: string | null
+  updatedsincestart: boolean
+}
+
 export type QuestionChoice = {
   label: string
   value: number
@@ -274,6 +287,10 @@ export type FormValidHelperText = {
 
 export type FismaTableProps = {
   scores: Record<number, SystemScoreEntry>
+  // Per-system questionnaire progress for the active data call, keyed by
+  // fismasystemid. Optional so the table degrades to an em-dash column if
+  // the progress fetch fails - score display must not depend on it.
+  progress?: Record<number, ScoreProgress>
 }
 
 export type ThemeColor =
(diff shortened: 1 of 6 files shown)


tarratsco:
Update progress on the Systems table:

image

MackOverflow:
Moving this to #483 — fork PRs can't access the org secrets required for Snyk and the deploy CI gates. Your commits and authorship carry over unchanged. Thanks @tarratsco!

PR #481, opened 2026-07-09
feat(questionnaire): encrypt drafts at rest — non-extractable IDB key + lifecycle hardening (#475)

### LINKED ISSUE #475: Encrypted questionnaire drafts: confidentiality hardening and persistence fixes

jsos3-cms:
In-repo rehome of #478 by @costacalvin (Calvin Costa) so the org-secret CI gates (dev deployment + checks) can run against the branch. The head commit is identical to #478 (`7d77b62`); authorship is preserved. The original PR stays open as a draft until this one merges.

Closes #475. Supersedes #476.

## Summary

Resolves all five blocking items and the remaining non-blocking hardening items from the encrypted questionnaire drafts issue. Replaces the PBKDF2-derived key and initial encryption approach from #476 with a non-extractable per-device key in IndexedDB, and hardens the draft lifecycle wiring throughout `QuestionnairePage`.

## Dependencies / predecessors

- **ztmf#399** — Original "Encrypt Draft text of freeform box at rest" ticket. Answered the post-quantum question (AES-256 is CNSA 2.0 compliant for data at rest; PQC standards target asymmetric primitives not used here). Closed in favor of #475.
- **ztmf-ui#476** — First encryption cut (landed on main). Introduced `draftStore.ts`, basic draft wiring in `QuestionnairePage`, and the initial `draftStore.test.ts`. Used PBKDF2 key derivation from a non-secret userid; #475 tracks the blockers. This PR replaces the key approach and addresses all tracked gaps.

## Blocking items resolved (#475)

| # | Issue | Resolution |
|---|-------|------------|
| 1 | Key derived only from non-secret `userid` — anyone with profile disk access could re-derive and decrypt | Replaced with a **non-extractable `AES-256-GCM` key** generated via `crypto.subtle.generateKey({ extractable: false })` and stored in **IndexedDB**. The key cannot be exported via the JS API; an attacker needs a live authenticated browser session, not just a disk dump. |
| 2 | `saveDraft` swallowed errors — UI showed "Draft saved" even on quota/private-browsing failure | `saveDraft` now returns `Promise<boolean>`. The debounce effect checks the return value and sets `draftStatus = 'error'` on failure, surfacing a visible error state in the UI. |
| 3 | `loadDraft` cast without runtime check — `notes: undefined` would throw on `notes.length` and blank the questionnaire | Added a runtime shape guard after decrypt: evicts and returns `null` if either field is missing or has the wrong type. |
| 4 | Draft key omitted `userid` — one user's draft could restore into another user's session on a shared machine | localStorage key is now `ztmf_draft_{sha256(userid)[0:16]}_{fismasystemid}_{functionid}_{datacallid}`. The userid segment is SHA-256 hashed (base64url, 16 chars) so raw UUIDs are not visible in DevTools. `clearOtherUserDrafts` is called on every authenticated app load to evict entries not matching the current user's prefix. |
| 5 | Reverting A→B→A left the B draft in storage to resurface later | The debounce effect detects when `selectQuestionOption === initQuestionChoice && notes === initNotes` and calls `clearDraft` — unless the draft was just auto-restored (`draftStatus === 'restored'`), in which case matching the server state does not indicate a user revert and the draft is preserved. |

## Non-blocking hardening

- **Random per-entry IV** — `crypto.getRandomValues(new Uint8Array(12))` on every `encryptDraft` call. No static salt (eliminated by the IDB key approach).
- **Generation counter (`saveGenRef`)** — incremented on every explicit `clearCurrentDraft()` call and on every navigation (`handleListItemClick`, Back button, Next/Complete button onClick). Any in-flight debounce `.then()` that fires after a clear or navigation is a no-op (`saveGenRef.current !== capturedGen`).
- **`isCurrent` callback in `saveDraft`** — checked at two points inside the async function: before encryption begins, and again before the final `localStorage.setItem`. Prevents a stale in-flight save from resurrecting a draft that was cleared during the ~10 ms AES-GCM encryption window.
- **Schema/version field** — `StoredDraft.v` (`DRAFT_VERSION = 1`). Entries with a mismatched version are evicted on first load, enabling safe future format migrations.
- **`questionScoresRef`** — stable ref updated every render; `fetchOptions` reads scores through the ref instead of enrolling `questionScores` as an effect dep, which previously caused a double-run of the effect after every `saveResponse` call.

## Additional correctness fixes

- **Datacall-switch batching** — `setQuestions`, `setDatacallID`, `setQuestionId`, and `setQuestionScores` are now batched together after **both** the questions and scores APIs return. Previously `setQuestionId` fired after the questions API alone, causing the `questionId` effect to run with stale scores from the prior datacall and then run a second time when scores arrived — creating a window where the debounce could incorrectly clear a restored draft.
- **Abort guards in `fetchOptions`** — two `controller.signal.aborted` checks bracket the `loadDraft` call. `crypto.subtle.decrypt` is genuinely async (~10–50 ms); without the guard, rapid datacall switching could write a stale draft's values into the newly active question's state.
- **`setLoadingQuestion` debounce gate** — `loadingQuestion` is set to `true` at the start of `fetchData` and cleared only after `fetchOptions` finishes. The debounce effect returns early while `loadingQuestion` is true, preventing stale saves from firing during system or datacall transitions.
- **`setLoadingQuestion(false)` on non-redirect auth errors** — all three `fetchData` catch blocks (questions fetch, scores fetch, outer) now call `setLoadingQuestion(false)` before returning on auth-handled errors. For 403s that show a toast without redirecting (FORBIDDEN_ORIGIN, plain controller-level 403), the component stays mounted and the spinner would otherwise be stuck indefinitely.
- **`setLoadingQuestion(false)` guard in `fetchOptions` finally** — wrapped in `if (!controller.signal.aborted)` so a newer in-flight request owns the loading state after rapid navigation.

## Files changed

| File | What changed |
|------|-------------|
| `src/views/QuestionnairePage/draftStore.ts` | Rewritten — IDB key management replaces PBKDF2, AES-256-GCM encrypt/decrypt, `saveDraft` returns `Promise<boolean>`, `isCurrent` callback, hashed userid keys, `DRAFT_VERSION` schema field, test-only escape hatches |
| `src/views/QuestionnairePage/draftStore.test.ts` | Extended to 33 tests — adds coverage for `isCurrent` early-abort at both checkpoints, `'error'` status on quota failure, version mismatch eviction, corrupt ciphertext eviction, cross-user isolation with hashed keys |
| `src/views/QuestionnairePage/QuestionnairePage.tsx` | `saveGenRef` generation counter, `questionScoresRef`, `setLoadingQuestion` debounce gate, state batching, abort guards in `fetchOptions`, `isCurrent` passed to `saveDraft`, `setLoadingQuestion(false)` on all auth-handled and error paths |
| `src/views/Title/Title.tsx` | `clearOtherUserDrafts` updated to use hashed prefix format |

## Test plan

- [x] Open a questionnaire question, type in the notes field. Confirm "Draft saved" banner appears after ~1 second.
- [x] Refresh the page. Confirm the draft is restored and "Draft restored" banner appears.
- [x] Verify draft content in DevTools → Application → Local Storage is ciphertext, not plaintext.
- [x] Verify DevTools → Application → IndexedDB → `ztmf-draft-keys` contains an opaque `CryptoKey` entry.
- [x] Verify localStorage keys use the format `ztmf_draft_<16-char-opaque-string>_...` (no raw userid visible).
- [x] Select an answer, switch to a different datacall question, switch back. Confirm the draft is still present and the correct answer/notes are shown for each question.
- [x] Select an answer, revert to the original server value. Confirm the draft is cleared (no "Draft restored" on next page load for that question).
- [x] Fill in a question, click Next. Confirm the draft is cleared on the submitted question.
- [ ] Fill in a question and immediately click Next before the 1-second debounce fires. Confirm no stale draft reappears on re-navigation to that question.
- [x] Open the questionnaire in a private browsing window (IDB and localStorage may be limited). Confirm no crash — the error banner appears if the draft cannot be saved.
- [x] With drafts saved, log in as a different user. Verify DevTools → Local Storage contains no `ztmf_draft_` entries for the previous user.
- [x] `npx jest --no-coverage src/views/QuestionnairePage/draftStore.test.ts` — all 33 tests pass.

---

*Rehomed from #478 (@costacalvin). Original commit `7d77b62` preserved verbatim.*

### DIFF
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -449,12 +449,16 @@ export default function QuestionnarePage() {
             )
           } catch (error) {
             if (controller.signal.aborted) return
-            if (isAuthHandled(error)) {
-              setLoadingQuestion(false)
-              return
+            if (!isAuthHandled(error)) {
+              console.error('Error fetching question scores:', error)
+              notify(ERROR_MESSAGES.tryAgain, 'error')
             }
-            console.error('Error fetching question scores:', error)
-            notify(ERROR_MESSAGES.tryAgain, 'error')
+            // Fall through to the batch below with an empty scores map so the
+            // sidebar/URL commit together with questions/datacallID/questionId,
+            // even when the scores call 403s without redirecting (auth-handled,
+            // component stays mounted). Returning here instead would leave the
+            // sidebar/URL on the new system while the content stayed on the old
+            // one. The [questionId] effect's finally clears the spinner.
           }
           // Batch questions + scores + datacallID + questionId so the questionId
           // effect fires exactly once with the correct scores already in
@@ -539,6 +543,15 @@ export default function QuestionnarePage() {
           const sys = systemRef.current
           const uid = userInfo.userid
           if (controller.signal.aborted) return
+          // Read-only sessions never load the draft again, so evict any lingering
+          // entry instead of letting it sit for the full TTL. Bump the save
+          // generation first: an autosave that fired just before isReadOnly
+          // flipped may still be mid-flight, and without the bump its isCurrent()
+          // checks would pass and rewrite the draft after this clear.
+          if (isReadOnly && sys && questionId && datacallID > 0) {
+            saveGenRef.current++
+            void clearDraft(uid, sys, questionId, datacallID)
+          }
           const draft =
             !isReadOnly && sys && questionId && datacallID > 0
               ? await loadDraft(uid, sys, questionId, datacallID)
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -426,6 +426,7 @@ export default function QuestionnarePage() {
             if (controller.signal.aborted) return
             if (isAuthHandled(error)) return
             notify(ERROR_MESSAGES.tryAgain, 'error')
+            setLoadingQuestion(false)
             return
           }
           if (questionsEmpty) {
@@ -462,6 +463,7 @@ export default function QuestionnarePage() {
           if (controller.signal.aborted) return
           if (isAuthHandled(error)) return
           console.error('Error fetching data:', error)
+          setLoadingQuestion(false)
         }
       }
       fetchData()
@@ -567,7 +569,7 @@ export default function QuestionnarePage() {
           if (isAuthHandled(error)) return
           console.error('Error fetching data:', error)
         } finally {
-          setLoadingQuestion(false)
+          if (!controller.signal.aborted) setLoadingQuestion(false)
         }
       }
       fetchOptions()
@@ -610,10 +612,14 @@ export default function QuestionnarePage() {
     const currentGen = saveGenRef.current
     const timer = setTimeout(() => {
       if (saveGenRef.current !== currentGen) return
-      saveDraft(userInfo.userid, system, questionId, datacallID, {
-        selectQuestionOption,
-        notes,
-      }).then((saved) => {
+      saveDraft(
+        userInfo.userid,
+        system,
+        questionId,
+        datacallID,
+        { selectQuestionOption, notes },
+        () => saveGenRef.current === currentGen
+      ).then((saved) => {
         if (saveGenRef.current !== currentGen) return
         if (saved) {
           if (draftStatusRef.current !== 'restored') setDraftStatus('saved')
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -142,12 +142,20 @@ export default function QuestionnarePage() {
   const [selectQuestionOption, setSelectQuestionOption] =
     React.useState<number>(-1)
   const [draftStatus, setDraftStatus] = React.useState<
-    'idle' | 'restored' | 'saved'
+    'idle' | 'restored' | 'saved' | 'error'
   >('idle')
   // Refs read inside setTimeout/event callbacks to get the current value
   // without enrolling the effects in those deps (prevents extra re-runs).
   const draftStatusRef = React.useRef(draftStatus)
   draftStatusRef.current = draftStatus
+  // Stable ref so fetchOptions always reads the latest scores without enrolling
+  // questionScores as a dep of the questionId effect. Stale-closure-safe because
+  // the ref updates synchronously on every render before any effect runs.
+  const questionScoresRef = React.useRef(questionScores)
+  questionScoresRef.current = questionScores
+  // Incremented on every explicit draft clear so in-flight debounced saves
+  // that fire after a clear don't resurrect the just-removed draft.
+  const saveGenRef = React.useRef(0)
   const unsavedRef = React.useRef({
     selectQuestionOption,
     initQuestionChoice,
@@ -238,14 +246,16 @@ export default function QuestionnarePage() {
     }
   }
   const handleListItemClick = (index: number) => {
+    saveGenRef.current++
     setLoadingQuestion(true)
     setSelectedIndex(index)
     setQuestionId(index)
   }
 
   const clearCurrentDraft = () => {
+    saveGenRef.current++
     if (system && questionId && datacallID > 0) {
-      clearDraft(userInfo.userid, system, questionId, datacallID)
+      void clearDraft(userInfo.userid, system, questionId, datacallID)
     }
     setDraftStatus('idle')
   }
@@ -312,6 +322,9 @@ export default function QuestionnarePage() {
       // view does not bleed into the next render when system changes.
       setNoQuestions(false)
       const fetchData = async () => {
+        // Gate the debounce effect during the entire system/datacall transition
+        // so stale pending saves don't fire while the question list reloads.
+        setLoadingQuestion(true)
         try {
           let questionsEmpty = false
           let datacall = ''
@@ -320,13 +333,11 @@ export default function QuestionnarePage() {
             selectedDatacall.datacallid !== latestDataCallId
           let activeDataCallId: number
           if (isHistorical && selectedDatacall) {
-            setDatacallID(selectedDatacall.datacallid)
             datacall = selectedDatacall.datacall.replaceAll(' ', '_')
             setDatacall(datacall)
             setIsPastDeadline(true)
             activeDataCallId = selectedDatacall.datacallid
           } else {
-            setDatacallID(latestDataCallId)
             datacall = latestDatacall.replaceAll(' ', '_')
             setDatacall(datacall)
             setIsPastDeadline(
@@ -334,6 +345,9 @@ export default function QuestionnarePage() {
             )
             activeDataCallId = latestDataCallId
           }
+          // Hoisted so both the questions block and the final batch can access them.
+          const questionData: Record<number, Question> = {}
+          let sortedFuncId: number[] = []
           try {
             const response = await axiosInstance.get(
               `/fismasystems/${system}/questions`,
@@ -352,7 +366,6 @@ export default function QuestionnarePage() {
               setLoadingQuestion(false)
             } else {
               const organizedData: Record<string, FismaQuestion[]> = {}
-              const questionData: Record<number, Question> = {}
               data.forEach((question: FismaQuestion) => {
                 if (!organizedData[question.pillar.pillar]) {
                   organizedData[question.pillar.pillar] = []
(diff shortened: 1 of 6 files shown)


MackOverflow (commented):
two nits, sending back for attention-if these are intentional, cool, but want to confirm. Nits: 

Correctness issues found
1. State-batching change can leave the page in an inconsistent half-updated state on a specific 403 path — src/views/QuestionnairePage/QuestionnairePage.tsx (the fetchData effect, ~lines 408–467)

To fix the "questionId effect fires twice with stale scores" race, setQuestions, setDatacallID, and setQuestionId were moved out of the questions-fetch block and into a single batch that only runs after the scores fetch also completes. But the scores fetch's catch block returns early on isAuthHandled(error) (documented in authInterceptor.ts as covering FORBIDDEN_ORIGIN and plain controller-level 403s — both cases where the component intentionally stays mounted rather than redirecting):

} catch (error) {
  if (controller.signal.aborted) return
  if (isAuthHandled(error)) {
    setLoadingQuestion(false)
    return   // <-- skips the batch below entirely
  }
  ...
}
setQuestions(questionData)
setQuestionScores(hashTable)
setDatacallID(activeDataCallId)
setQuestionId(sortedFuncId[0])
On that path, setCategories, setFunctionIdIdx, setStepFunctionId, setSelectedIndex, and navigate(...) have already fired earlier in the questions block — so the sidebar and URL move to the new system/datacall — but questions, datacallID, and questionId never update, and the spinner turns off (setLoadingQuestion(false)) so nothing signals it's incomplete. The old code set setQuestion/setDescription/setNotePrompt directly from questionData in the questions block, independent of the scores call, so this same failure previously left a working (if score-less) view. Now it leaves a broken one.

Failure scenario: user has access to a system's questions but the scores endpoint 403s without a code (e.g. a permission edge case on that specific call) → sidebar shows the new system, main pane shows stale/blank question content, no loading indicator, no error toast beyond the one that already faded.
No test exercises this path — there's no QuestionnairePage.test.tsx in the repo at all, and the PR's manual test-plan checklist doesn't cover it either.
2. Read-only draft eviction was dropped, not replaced — src/views/QuestionnairePage/QuestionnairePage.tsx (fetchOptions effect)

The old code:

if (isReadOnly && sys && questionId && datacallID > 0)
  clearDraft(uid, sys, questionId, datacallID)
was removed with no equivalent added elsewhere. I checked every remaining isReadOnly reference in the file (debounce effect, beforeunload handler, JSX) — none of them clear a draft when the session becomes read-only. The debounce effect explicitly skips both save and clear while isReadOnly is true.

Failure scenario: a user is editing a question, the datacall deadline passes (isPastDeadline flips → isReadOnly true) or they become a read-only admin, leaving an in-progress draft in localStorage. It's never loaded again (guarded by !isReadOnly) and now never evicted either — it just sits there for the full 14-day TTL (or indefinitely, since TTL eviction only runs inside loadDraft, which is never called for that key). Low severity (still encrypted, still user-isolated) but it's a straightforward regression against the PR's own stated hardening goal (item 5 in the table: stale drafts shouldn't resurface/linger).

costacalvin:
For number 2 **Read-only draft eviction was dropped, not replaced**, I did remove the IsReadOnly clearing as it was causing deletions of the drafts in certain race conditions, but I think it is safe to add back now after we added batch updates that is referenced in number 1. Tested it locally and the issue no longer occurs with the IsReadOnly clear.

QuestionnairePage.tsx L542 Add:
`if (isReadOnly && sys && questionId && datacallID > 0)
            void clearDraft(uid, sys, questionId, datacallID)`

jsos3-cms:
Thank you, taking the feedback and folding it in!

jsos3-cms:
Both confirmed and fixed on the branch — thanks for the careful review, @MackOverflow, and @costacalvin for the read-only diagnosis.

**1 — Half-updated view on a scores 403.** On the auth-handled scores path (`FORBIDDEN_ORIGIN` or a plain controller-level 403, where the page stays mounted), the catch returned before the batch. The sidebar, URL, and categories had already moved to the new system/datacall while `questions`/`datacallID`/`questionId` stayed on the old one, and the spinner switched off — so the stale content rendered with no error. It's reachable by switching the active datacall too, not just the system, and a save in that window could write the old question's answer and `scoreid` under the new system. The scores catch now surfaces a toast only on non-auth errors and falls through to the batch with an empty scores map, so the sidebar/URL and `questions`/`datacallID`/`questionId` commit together — the same score-less-but-consistent view the pre-batch code produced. The questionId effect clears the spinner as before.

**2 — Read-only draft eviction.** Re-added, with one addition on top of the inline clear: it bumps `saveGenRef` before clearing. An autosave whose 1s timer fired just before `isReadOnly` flipped can still be mid-flight — its `clearTimeout` is then a no-op — and its `isCurrent()` checks would otherwise pass and rewrite the draft right after the clear, resurrecting exactly what we evicted. Bumping the generation first invalidates that in-flight save. It sits after the abort guard, so a superseded run can't clear on behalf of a stale invocation.

**3 — Stale answer on a fast Next→Back.** Closed a related gap while in here: the post-save `fetchQuestionScores` is fire-and-forget, so returning to a just-saved question before that GET resolves seeds it from the pre-save snapshot. A newly-answered question then showed as unanswered with `scoreid = 0`, and re-answering it would POST a duplicate score. Added a narrow re-seed — when the scores map refreshes while idle on a question with no unsaved edits and no restored draft, the answer is re-derived from the fresh map. It never overwrites in-progress edits.

Extracted the answer-derivation and re-seed guard into a small helper with unit tests. The scores-403 and read-only paths are effect-level; there's no component test harness for this page yet, so dedicated coverage there is a worthwhile follow-up.

jsos3-cms:
Opened #487 to track the component-level (RTL) test coverage follow-up for the QuestionnairePage effect paths (scores-403 batch-commit, read-only eviction, out-of-band re-seed) — there's no rendered-component harness for this page yet.

MackOverflow (approved):
LGTM!

PR #482, opened 2026-07-09
Add non-production environment warning banner

### LINKED ISSUE #479: Add a "this is dev!" banner
### Feature Type

UI Component

### Workflow Description

We need a stronger visual representation in the development environment for users. A 508 compliant banner that informs users that this is dev, it will experience active deployments during business hours, and that some instability will be transient. Also maybe point to the cms.hhs.gov email endpoint or our feedback form in the banner

### Acceptance Criteria

- [x] Indicates to users this is dev, and is a potentially unstable environment
- [x] Points to method to contact us

### Additional Context

_No response_

danielbowne:
Closes #479.

## What
image

Adds a dismissible warning banner that renders in every non-production environment (dev, impl, local) and is hidden in production. It tells users they are not on the live site, warns that active deployments can cause brief glitches (pause and refresh), and surfaces a feedback form and contact link.

## How it works

- **Environment gate:** `CONFIG.IS_NONPROD` is derived from Vite's build mode (`import.meta.env.MODE !== 'production'`). `build:prod` uses `--mode production`, so the banner fails closed and never appears in prod.
- **Copy and links are build-time injected, not committed.** Three optional variables drive the testing-specific text and links:
  - `VITE_DEV_BANNER_MESSAGE` — overrides the default copy
  - `VITE_DEV_FEEDBACK_URL` — feedback form link (https-only; other schemes are ignored)
  - `VITE_DEV_CONTACT_EMAIL` — contact mailto
- When a variable is empty the banner falls back to a mode-neutral default and hides the corresponding link. The deployed dev build sources these from **dev GitHub Environment secrets** (wired in `ui.yml`), so the feedback URL and contact address stay out of the repo and can be cleared when a testing window ends. impl and local show the generic default.
- Mounted once in the app shell (`Title.tsx`), directly below the existing USA banner. Dismissal is in-memory by design, so it returns on reload.

## Testing

- New unit tests cover the production hide, the non-prod default, override copy, whitespace-only fallback, https-only link filtering, and dismissal.
- Full suite green (258 passing); typecheck and lint clean.

## Before this deploys to dev

- The dev Environment secrets (`DEV_BANNER_MESSAGE`, `DEV_CONTACT_EMAIL`, `DEV_FEEDBACK_URL`) still need to be set so the first dev deploy shows the full testing copy. The real feedback-form URL is still outstanding; until it is set, the feedback link stays hidden and everything else renders.

---

### Follow-up: only show once signed in

The banner first rendered on the pre-login page too. It is now gated to an active session (`loaderData.status === 200`): signed-in users see the full environment-specific copy, while pre-login visitors see only the generic non-production marker.

### Testing note

There is no good way to catch this locally. Local dev auto-authenticates (`VITE_LOCAL_DEV`), so `loaderData.status` is always 200 and the pre-login state never renders on a dev machine. Issues in the pre-login path only surface once deployed, where real unauthenticated sessions exist.

### DIFF
--- a/.env.development.example
+++ b/.env.development.example
@@ -23,6 +23,14 @@ VITE_AUTH_TOKEN3=your-jwt-token-here
 # Set to 'false' for local dev, 'true' for AWS environments
 VITE_IDP_ENABLED=false
 
+# Development banner overrides (all optional; leave empty to use built-in copy)
+# The banner shows in every non-production environment regardless of these.
+# Set these locally only to preview testing-window copy; real values for the
+# deployed dev environment are injected from CI secrets, never committed.
+VITE_DEV_BANNER_MESSAGE=
+VITE_DEV_FEEDBACK_URL=
+VITE_DEV_CONTACT_EMAIL=
+
 # Legacy proxy configuration (leave empty for default setup)
 TOKEN=
 PROXY_URI=
--- a/.env.example
+++ b/.env.example
@@ -8,3 +8,10 @@
 
 TOKEN = ''
 PROXY_URI = ''
+
+# Development banner overrides (optional). Blank by design in the repo; the
+# deployed dev environment injects real values from CI secrets during a testing
+# window and clears them afterward. See .github/workflows/ui.yml.
+VITE_DEV_BANNER_MESSAGE = ''
+VITE_DEV_FEEDBACK_URL = ''
+VITE_DEV_CONTACT_EMAIL = ''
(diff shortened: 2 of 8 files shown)


MackOverflow:
looks great Danny, no nits

MackOverflow (approved):
lgtm Danny, we just need to confirm the dev Environment secrets are set before the first dev deploy so the banner shows the intended copy rather than the generic fallback.

jsos3-cms (approved):
Clean and LGTM!

PR #485, opened 2026-07-09
Gate the dev banner to signed-in users

danielbowne:
Follow-up to #482. That PR merged before this gate landed, so `main` currently mounts the banner unconditionally and it renders on the pre-login page.

## Change

Adds an `authenticated` prop to `DevEnvironmentBanner`. `Title.tsx` passes `loaderData.status === 200`: signed-in users see the full environment-specific copy, pre-login visitors see only the generic non-production marker.

## Testing note

There is no good way to catch this locally. Local dev auto-authenticates (`VITE_LOCAL_DEV`), so `loaderData.status` is always 200 and the pre-login state never renders on a dev machine. Issues in the pre-login path only surface once deployed, where real unauthenticated sessions exist.

Component tests extended to cover the pre-login state (8 total). Typecheck and lint clean.

### DIFF
--- a/src/components/DevEnvironmentBanner/DevEnvironmentBanner.test.tsx
+++ b/src/components/DevEnvironmentBanner/DevEnvironmentBanner.test.tsx
@@ -37,10 +37,28 @@ test('shows the default notice in a non-production environment', () => {
   expect(screen.getByText(/non-production environment/i)).toBeInTheDocument()
 })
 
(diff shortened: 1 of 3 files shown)


MackOverflow (approved):
no nits, Consider adding the authenticated-empty-override → default-message test for symmetry/completeness, though it doesn't block on it.

PR #490, opened 2026-07-10
Dashboard filters, stat card sizing, and header alignment

### LINKED ISSUE #463: Dashboard summary cards overflow with long HHS system names (fixed height)
### Type
Bug (UI / layout)

### What happens
On the dashboard, the four summary cards (Total Systems, Average System Score, Highest System Score, Lowest System Score) have a fixed height. With the new HHS data, the Highest/Lowest cards now show long system names (e.g. "Administration for Native Americans ARP Smartsheet", "HHS Cares"), and the name text wraps and overflows below the card border instead of the card growing to fit it. The cards also no longer visually line up because one has spilled content.

### Cause
`src/views/StatisticBlocks/StatisticsBlocks.tsx` — `StatisticsPaper` is a styled `Paper` with a hard-coded `height: 120` (and the row wrapper uses a fixed height too). The label `Typography` uses `overflowWrap: 'break-word'`, so a long name wraps downward but the fixed height clips/overflows it.

### Desired behavior
- Cards resize to fit their content (the longest name) instead of overflowing.
- All summary cards stay the same size as each other (equal-height row), so the three score cards remain uniform.

### Suggested approach
- Replace the fixed `height` with a `minHeight` and let the row stretch (flex `align-items: stretch` / equal-height grid), so the tallest card sets the shared height and the others match.
- Optionally cap very long names with a max number of lines + ellipsis and a tooltip showing the full name, so an extreme value can't make the row unreasonably tall.

### Notes
Surfaced by the HHS load (system names are longer than the CMS set the layout was built for). Component: `src/views/StatisticBlocks/StatisticsBlocks.tsx`.

### LINKED ISSUE #489: Dashboard filters: Environment, OpDiv, and Data Call progress
### Feature Type

UI Component

### Workflow Description

Customer request (Elizabeth). Add filters to the dashboard system list so users can narrow the table without scrolling through everything:

- **Environment**
- **OpDiv**
- **Data Call progress** — the priority, so a user can isolate the systems that have **Not Updated** for the current data call.

The dashboard is already wide and overflows a 16" laptop, so filtering (rather than more table width) is the right lever to help users find what they need.

### Acceptance Criteria

- [ ] Filter the dashboard by Environment
- [ ] Filter the dashboard by OpDiv
- [ ] Filter by Data Call progress, including a way to isolate "Not Updated" systems
- [ ] Filters combine and can be cleared
- [ ] Accessible / 508 compliant (keyboard operable, labeled controls)

### Additional Context

Builds on the Data Call Progress column (#483). Related: #463 (dashboard overflow), CMS-Enterprise/ztmf#299 (updated systems identifiable).

danielbowne:
Closes #489. Closes #463.

## Dashboard filters (#489)
image

Client-side filters on the dashboard system list so users can narrow the table without scrolling:

- **Environment** and **OpDiv** — compact multi-selects (fixed height; the collapsed control shows a short summary and never grows with selections).
- **Not updated only** — a toggle that isolates systems with a questionnaire but zero updates for the active data call. Reuses the Data Call Progress column's own classifier so the filter and the column chip always agree.
- **Clear filters** — resets the three facets; disabled when nothing is active.

A facet filter is hidden when the visible systems span a single value. The filter logic lives in a pure, unit-tested module (`src/views/FismaTable/dashboardFilters.ts`); selections auto-prune when the option sets change (e.g. after Show Decommissioned refetches).

## Summary card sizing (#463)

The dashboard summary cards clipped long system names against a fixed height. Switched to a min-height with `align-items: stretch` so cards grow to fit their content and stay equal size.

## Header alignment

The USA banner and the development-environment banner text now left-align with the ZTMF logo, while the bars stay full width.

## Testing

- New unit tests for the filter logic (7), full suite green (309 passing), typecheck and lint clean.
- Verified locally against dev data (the local backend was rebuilt so the Data Call Progress endpoint is present).

## Note

The development banner's pre-login gate is tracked separately in #485 (still open); this branch does not include it, so that PR and this one both touch the banner mount and will need a trivial reconcile whenever both land.

### DIFF
--- a/src/views/FismaTable/FismaTable.tsx
+++ b/src/views/FismaTable/FismaTable.tsx
@@ -13,8 +13,18 @@ import {
   GridRowParams,
 } from '@mui/x-data-grid'
 import Tooltip from '@mui/material/Tooltip'
-import { Box, IconButton } from '@mui/material'
-import { useState } from 'react'
+import {
+  Box,
+  IconButton,
+  Checkbox,
+  FormControl,
+  Select,
+  MenuItem,
+  ListItemText,
+  Button,
+} from '@mui/material'
+import CloseIcon from '@mui/icons-material/Close'
+import { useState, useEffect, useMemo } from 'react'
 import FormControlLabel from '@mui/material/FormControlLabel'
 import Switch from '@mui/material/Switch'
 import FileDownloadSharpIcon from '@mui/icons-material/FileDownloadSharp'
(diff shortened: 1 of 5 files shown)


jsos3-cms (approved):
Rather than "Not Updated Only", which is hard to understand, how about "Yet to Start"? Is that a bit more descriptive of what that toggle does? Otherwise no blocking nits.

danielbowne:
> Rather than "Not Updated Only", which is hard to understand, how about "Yet to Start"? Is that a bit more descriptive of what that toggle does? Otherwise no blocking nits.

@jsos3-cms I didn't write the the code for that feature(but those that say not updated are actually completed), but in its current state not started would be incorrect.  Honestly the logic needs to be changed and there needs to be a migration when it was deployed.  I think I have something that is going to fix the real problem which is the select box filter for data calls, it needs to be bundled by years and then default to current year IF contains a data call.

Issue #467, opened 2026-07-08
Group the data-call selector by year (show paired CMS and HHS calls, don't merge)


danielbowne:
### Type
Enhancement / bug (UX)

### Problem
The HHS load added a second data call per year, so each year now has a paired CMS call and HHS call (e.g. 2025 has `FY2025 Q3` (CMS) and `FY25 ZTM` (HHS); 2023 has `FY2023 Q4` and `FY23 ZTM`). The data-call selector shows all of these in one flat list.

For a multi-system admin, picking the wrong one of a pair is confusing: CMS systems only have scores under the CMS call and HHS systems only under the HHS call, so selecting a ZTM (HHS) call makes every CMS system in the table read 0, and vice versa. With a flat list it isn't obvious that you've selected the other tenant's call.

<img width="539" height="505" alt="Image" src="https://github.com/user-attachments/assets/9da22a83-8fda-4ae0-9944-271f4320f373" />

### Why not merge them
#393 originally planned to consolidate each year's paired calls into one annual call, on the assumption that the two were structurally identical. They are not — the CMS and HHS calls differ, so they cannot be merged. They need to stay as separate calls and instead be displayed in a way that makes the pairing and the tenant clear.

### Proposed change
Group the data-call selector by year: the top level is the year (e.g. 2025), and each year expands to its associated calls (CMS + HHS) as sub-items. MUI Autocomplete supports `groupBy`, so the options can be grouped on a year derived from the call.

- Year headers, with each year's CMS and HHS calls listed under it.
- Clear labeling so it's obvious which entry is the CMS call and which is the HHS call.
- Picking a specific call still scopes the table to that call; the grouping just makes the choice unambiguous so an admin doesn't land on the other tenant's call by accident.

### Components
- `src/views/Title/Title.tsx` — the Datacall Autocomplete.
- Check the other data-call pickers for consistency (`src/components/ScoreDiffModal/ScoreDiffModal.tsx`).

### Acceptance
- The selector groups calls under their year, showing both the CMS and HHS call for each year.
- Each option clearly indicates its tenant/type.
- Ordering follows the deadline-based order (see #393 sort work).

### Notes
Supersedes #393's consolidation scope (merging the paired calls), which is not viable now that the calls are known to differ. Surfaced by the HHS historical load.

danielbowne:
working on this now

PR #493, opened 2026-07-10
feat(pillar-scores): label radar series with actual data-call names

### LINKED ISSUE #468: Radar chart: label series with actual data-call names instead of Current/Previous
### Type
Enhancement (UX)

### Request
On the pillar-scores radar chart, replace the generic series labels "Current" and "Previous" with the actual data-call names (e.g. "FY2025 Q3", "FY24 ZTM"), so it's explicit which data call each series represents.

### Why
Raised during a dev test. "Current" on the radar is the last completed set of scores, which can be confusing while a user is mid-questionnaire on the current in-progress call: they may read "Current" as their live answers rather than the last finalized call. Showing the real data-call labels removes that ambiguity and makes the comparison self-explanatory.

### Where
`src/components/PillarScoresModal/PillarScoresModal.tsx`:
- Radar series names: `name="Current"` (~line 591) and `name="Previous"` (~line 600).
- Tooltip labels: "Current Score" / "Previous Score" (~line 654, 710).
- Summary lines: "Previous: {score}" (~lines 371, 524) and the current counterpart.

The component already fetches `datacalls` and resolves calls by id, so the current/previous call names are available in-component. Likely low churn.

### Acceptance
- Radar series, tooltip, and summary labels show the actual current and previous data-call names instead of "Current"/"Previous".
- Graceful fallback when there is no previous call to compare against (single series, no "Previous" label).
- Long names are handled without breaking the legend/tooltip layout.

### Notes
Relates to the radar work under #332. Feedback from a stakeholder dev test.

costacalvin:
## Summary

Closes #468.

Replaces the generic \"Current\" / \"Previous\" labels throughout the Pillar Scores modal with the real data-call names, adds a user-selectable comparison picker, and fixes how FismaTable resolves each row's data call so the modal always opens anchored to the call the row is displaying.

### Acceptance criteria

| Requirement | Status |
|---|---|
| Radar series, tooltip, summary lines, and table headers show actual data-call names | ✅ |
| Graceful fallback when no previous call exists (single series, no "Previous" label/column) | ✅ |
| Long names handled without breaking legend/tooltip layout | ✅ |
| "Previous" resolved by deadline, not datacallid (per #393 / #467 issue comment) | ✅ |
| Label format matches year-grouped selector styling (e.g. "FY2025 Q3 · CMS") | ✅ |
| Works correctly alongside the multi-call selector (#467) | ✅ |
| Pillar Scores modal opens anchored to the same call the dashboard row is displaying | ✅ |

---

### `PillarScoresModal.tsx`

**Label replacement**
- `name="Current"` / `name="Previous"` on the `<Radar>` series replaced with real data-call names from `getQuarterName`, which appends `· {tenant}` (e.g. `· CMS`, `· HHS`) to match the year-grouped selector format
- System score card, pillar cards, data table column headers, and `aria-label` all use real names
- Tooltip replaced with a custom `content` renderer that wraps long names without overflow

**Deadline-based previous resolution**
- Previous call resolved via `sortDatacallsByDeadline` rather than by datacallid (unreliable after historical re-imports, see #393)
- `previousDatacall` (the deadline-adjacent predecessor call object) and `previousScoreEntry` (its score, which can be `null`) are tracked separately — so the picker shows the predecessor's name even when the system has no submission for it, displaying "No data for {name}" rather than silently falling back to a different call

**Comparison picker**
- "Viewing: **{name}** / Compare with: [dropdown]" bar added at the top of the modal when the system has more than one score
- Dropdown lists all data calls sorted newest-first by deadline, regardless of whether the system has submitted for each
- Three-way state (`undefined` = deadline default, `null` = user picked None, `number` = explicit selection) ensures "None" reliably clears the comparison without silently reverting
- Resets on every modal open or target change — effect depends on `[open, selectedDataCallId]` so switching between two systems that share the same active call correctly resets the picker

**Conditional rendering**
- `comparisonActive` flag distinguishes "comparison selected but no score → show 'No data for {name}'" from "None selected / no predecessor → hide comparison UI"
- `hasComparisonPillarData` guards the radar second series and comparison table column — requires the comparison entry to have pillar detail, preventing a phantom all-zero series when only an aggregate-level score exists

### `FismaTable.tsx` + `aggregateScores.ts` + `Home.tsx`

**Per-row data-call resolution**
- `buildDashboardMaps` now returns `chosenCallMap` — the single call chosen per system for its dashboard row (the one most recently updated by that system)
- `chosenCallMap` is lifted through `Home` → `FismaTable` → `handleOpenPillarScores`, so `selectedDataCallId` passed to the modal always matches the call the table row is showing
- Falls back to newest-by-deadline if no entry exists, then to `activeDataCallId` as a last resort

**Deadline sort tiebreak**
- Questionnaire button `renderCell` was sorting data calls with an inline deadline comparison that omitted the datacallid tiebreak; now uses `sortDatacallsByDeadline` consistently

### Tests

- `aggregateScores.test.ts` — multi-call test extended to assert `chosenCallMap` resolves to the most-recently-updated call; empty-maps test covers the new field
- `aggregateScores.ts` return type (`DashboardMaps`) updated with `chosenCallMap`

---
## UX
Datacall names on Pillar Scores modal
image

Comparison dropdown, sorted by deadline
image

Comparison with datacall that contains no scores for the selected system states "No Data for {datacallname}"
image

None selected on the comparison
image

## Test plan

- [ ] Open Pillar Scores for a system with multiple scores — confirm radar legend, tooltip, summary lines, and table headers all show real names with tenant suffix (e.g. "FY2025 Q3 · CMS")
- [ ] Hover a radar data point — confirm tooltip shows the data-call name, not "Current Score" / "Previous Score"
- [ ] Open for a system with only one score — confirm no Previous series, no comparison column, no "Previous" label, and no "Compare with" picker
- [ ] Use the "Compare with:" dropdown — confirm scores, trend arrows, radar overlay, and table column all update to the selected call
- [ ] Select a call the system has no score for — confirm "No data for {comparison call name}" appears in the score box and each pillar card, radar shows a single series, comparison column is hidden
- [ ] Select "None" — confirm comparison is fully hidden and does not silently revert to the default
- [ ] With a full year selected on the dashboard, open Pillar Scores for a multi-call system — confirm "Viewing:" shows the call the table row is displaying, not the global fallback
- [ ] Open the modal, change the comparison, close, reopen for a different system sharing the same active call — confirm the picker resets to the new system's default
- [ ] Open Pillar Scores for a system where the predecessor call exists but was never submitted — confirm the picker defaults to that predecessor's name (not "None") and pillar cards show "No data for {predecessor call name}"

## Questions for Reviewers

1. Should there be two dropdown selectors in the pillar scores modal (similar to the Questionnaire Page Compare Datacalls Modal)? Currently we only have one which is the comparison datacall and defaults to the "previous" datacall sorted by deadline.

2. Is the dropdown component in line with the design / good as is, or should it be changed to Autocomplete (used in other areas of the app)?

### DIFF
--- a/src/components/PillarScoresModal/PillarScoresModal.tsx
+++ b/src/components/PillarScoresModal/PillarScoresModal.tsx
@@ -125,8 +125,8 @@ const PillarScoresModal: React.FC<PillarScoresModalProps> = ({
 
   // True whenever a comparison is active (default or user-chosen call),
   // false when the user explicitly picked "None" or there is no predecessor
-  // at all. Distinguishes "selected a call with no score → show No previous
-  // data" from "None / no comparison possible → hide entirely".
+  // at all. Distinguishes "selected a call with no score → show 'No data for <name>'"
+  // from "None / no comparison possible → hide entirely".
   const comparisonActive =
     selectedComparisonId !== null &&
     (selectedComparisonId !== undefined || previousDatacall !== null)
@@ -475,7 +475,7 @@ const PillarScoresModal: React.FC<PillarScoresModalProps> = ({
                             mb: 0.3,
                           }}
                         >
-                          {`${comparisonDatacallName}: ${previousSystemScore.toFixed(2)}`}
+                          {`Compared with ${comparisonDatacallName}: ${previousSystemScore.toFixed(2)}`}
                         </Typography>
                         <Typography
                           variant="caption"
@@ -504,7 +504,7 @@ const PillarScoresModal: React.FC<PillarScoresModalProps> = ({
                           fontSize: '0.85rem',
                         }}
                       >
-                        No previous data
+                        No data for {comparisonDatacallName}
                       </Typography>
                     )
                   }
@@ -627,7 +627,7 @@ const PillarScoresModal: React.FC<PillarScoresModalProps> = ({
                             fontSize: '0.75rem',
                           }}
                         >
-                          {`${comparisonDatacallName}: ${previousPillarScore.toFixed(2)}`}
+                          {`Compared with ${comparisonDatacallName}: ${previousPillarScore.toFixed(2)}`}
                         </Typography>
                       )}
 
@@ -641,7 +641,7 @@ const PillarScoresModal: React.FC<PillarScoresModalProps> = ({
                               fontSize: '0.75rem',
                             }}
                           >
-                            No previous data
+                            No data for {comparisonDatacallName}
                           </Typography>
                         )}
 
--- a/src/components/PillarScoresModal/PillarScoresModal.tsx
+++ b/src/components/PillarScoresModal/PillarScoresModal.tsx
@@ -327,16 +327,18 @@ const PillarScoresModal: React.FC<PillarScoresModalProps> = ({
                   <FormControl size="small">
                     <Select
                       value={
-                        selectedComparisonId === undefined
-                          ? previousScoreEntry?.datacallid ?? ''
-                          : selectedComparisonId ?? ''
+                        sortedDatacalls.length === 0
+                          ? ''
+                          : selectedComparisonId === undefined
+                            ? previousScoreEntry?.datacallid ?? ''
+                            : selectedComparisonId ?? ''
                       }
                       onChange={(e) => {
                         const val = e.target.value
                         setSelectedComparisonId(val === '' ? null : Number(val))
                       }}
                       displayEmpty
-                      sx={{ minWidth: 160 }}
+                      sx={{ minWidth: 160, maxWidth: 260 }}
                     >
                       <MenuItem value="">None</MenuItem>
                       {sortedDatacalls
@@ -479,7 +481,11 @@ const PillarScoresModal: React.FC<PillarScoresModalProps> = ({
                         </Typography>
                       </>
                     )
-                  } else if (latestScore.systemscore && !previousSystemScore) {
+                  } else if (
+                    latestScore.systemscore &&
+                    comparisonScoreEntry !== null &&
+                    !previousSystemScore
+                  ) {
                     return (
                       <Typography
                         variant="body2"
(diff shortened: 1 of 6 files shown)


jsos3-cms:
@costacalvin Rehoming this one and running some reviews, be back in a bit.

jsos3-cms:
Thanks for this, @costacalvin — nice work. I've rehomed it to an in-repo branch as #520 so CI and the dev deployment can run against it (fork PRs can't reach the org secrets those workflows need). Your original commit is preserved on that branch.

On your two questions:

1. **One dropdown vs. two selectors** — let's keep the single "Compare with" dropdown. The modal opens from a dashboard row whose data call is already fixed (the row's most-recently-updated call), so that row is effectively the "current" side; a second selector would let the two drift out of sync with the row that was clicked and would duplicate the dashboard's own call selector. The two-selector pattern fits the Questionnaire compare view because neither side is anchored there.

2. **Select vs. Autocomplete** — let's keep the Select. It matches how data calls are picked elsewhere (the dashboard's primary data-call filter is also a Select); Autocomplete is reserved for the larger entity lists (systems, users, OpDivs). If the list grows enough to want search, the on-idiom next step would be a Select with fiscal-year subheaders rather than Autocomplete — open to your take if you feel it's already long enough to warrant it.

One follow-up change I made on #520: the default comparison and the dropdown were being sourced from the full data-call list across tenants, so HHS/OpDiv systems (whose only call is the annual ZTM) defaulted to a CMS quarterly and showed "No data." I scoped both the default and the dropdown to the calls each system actually has a score for. Details are on #520.

Let's continue review on #520.

jsos3-cms:
Merged — this shipped via the rehomed PR #520, squashed to `main` as a2c21ea, carrying your original commit plus the follow-up fixes discussed above. #468 is now closed.

Thanks again, @costacalvin — closing this one in favor of the merged rehome. One pre-existing edge case we spotted along the way is tracked separately in #521.

PR #494, opened 2026-07-10
Group the data-call selector by year and aggregate the dashboard

### LINKED ISSUE #467: Group the data-call selector by year (show paired CMS and HHS calls, don't merge)

danielbowne:
Closes #467.

## Problem

The HHS historical import added a second data call per year (e.g. 2025 has FY2025 Q3 and FY25 ZTM). The flat data-call selector defaulted to a single call, so every system that wasn't in that specific call showed as `0/40 Not updated` on the dashboard, making it look broken.

## What changed

**Year-grouped selector.** The data-call picker now groups calls by fiscal year (parsed from the name). Selecting a year toggles on all of that year's calls; the dashboard aggregates them so each system shows its real status. Per-call checkboxes let you narrow within a year. Defaults to the latest year with data. Each option is tagged with its call type (ZTM / Q#) and deadline as a visual aid — note that tag is derived from the name, not a stored field.

**Dashboard aggregation.** `Home` fetches `/scores/aggregate` and `/scores/progress` for each active call and merges them by system (frontend only, no backend change).

**Multi-call systems.** We found systems that appear in more than one of a year's calls (217 systems in 2025). Handled as:
- Dashboard shows the system's **newest call by deadline**.
- The **questionnaire** action opens the system's own call; if it has more than one, a small picker lets the user choose which to open.
- Read-only / deadline state is decided by the opened call's own deadline.

**Export** targets the call shared by the selected rows; a selection spanning multiple calls is disabled.

## Testing

- New unit tests: year/type parsing and grouping, the score/progress merge (including the multi-call newest-wins case), and the filter logic.
- Full suite green (352), typecheck and lint clean. Verified locally against dev data.

- **Pillar Scores modal** has the same multi-call ambiguity; being handled on a separate branch.
- A single system that belongs to two calls can't be exported yet (the multi-call export gate catches it).
- The default score shown for a multi-call system (currently newest-by-deadline) is still being confirmed with the team.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -304,6 +304,9 @@ export type FismaTableProps = {
   // fismasystemid. Optional so the table degrades to an em-dash column if
   // the progress fetch fails - score display must not depend on it.
   progress?: Record<number, ScoreProgress>
+  // Which active data call(s) each system has scores in, keyed by
+  // fismasystemid, so per-row actions target the system's own call.
+  systemCallMap?: Record<number, number[]>
 }
 
 export type ThemeColor =
--- a/src/utils/datacallGrouping.test.ts
+++ b/src/utils/datacallGrouping.test.ts
@@ -1,0 +1,76 @@
+import { parseDatacallName, groupDatacallsByYear } from './datacallGrouping'
+
+type Row = { datacallid: number; datacall: string; deadline: string }
+
+describe('parseDatacallName', () => {
+  it('parses a 4-digit CMS quarterly call', () => {
+    expect(parseDatacallName('FY2025 Q3')).toEqual({
+      fiscalYear: 2025,
+      tenant: 'CMS',
+    })
+  })
+
+  it('parses a 2-digit HHS ZTM call', () => {
+    expect(parseDatacallName('FY25 ZTM')).toEqual({
+      fiscalYear: 2025,
+      tenant: 'HHS',
+    })
+  })
+
+  it('expands a 2-digit fiscal year to 20##', () => {
+    expect(parseDatacallName('FY23 Q1').fiscalYear).toBe(2023)
+  })
+
+  it('returns Other/null for an unrecognized name', () => {
+    expect(parseDatacallName('Special one-off')).toEqual({
+      fiscalYear: null,
+      tenant: 'Other',
+    })
+  })
+
+  it('tolerates surrounding whitespace', () => {
+    expect(parseDatacallName('  FY24 ZTM  ').fiscalYear).toBe(2024)
+  })
+})
+
+describe('groupDatacallsByYear', () => {
+  const calls: Row[] = [
+    { datacallid: 1, datacall: 'FY23 ZTM', deadline: '2023-09-30' },
+    { datacallid: 2, datacall: 'FY2025 Q3', deadline: '2025-05-07' },
+    { datacallid: 3, datacall: 'FY25 ZTM', deadline: '2025-09-30' },
+    { datacallid: 4, datacall: 'FY24 ZTM', deadline: '2024-09-30' },
+    { datacallid: 5, datacall: 'FY2023 Q4', deadline: '2024-08-31' },
+  ]
+
+  it('groups by fiscal year, newest first', () => {
+    const groups = groupDatacallsByYear(calls)
+    expect(groups.map((g) => g.year)).toEqual([2025, 2024, 2023])
+  })
+
+  it('pairs the CMS and HHS calls under the same year', () => {
+    const y2025 = groupDatacallsByYear(calls).find((g) => g.year === 2025)
+    expect(y2025?.calls.map((c) => c.datacall)).toEqual([
+      'FY25 ZTM', // deadline 2025-09-30, furthest out first
+      'FY2025 Q3', // deadline 2025-05-07
+    ])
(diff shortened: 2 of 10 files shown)


MackOverflow (commented):
Two biggies here: 

🟡 Silent per-call failures reintroduce the exact "shows 0" symptom (observability)
Home.tsx — both fetch loops swallow errors with .catch(() => []):

.then((res) => res.data.data as ScoreAggregate[])
.catch(() => [] as ScoreAggregate[])
The old code logged console.error on failure. Now, if one call's /scores/aggregate fails (network blip, 500), every system in that call silently shows 0/missing with no diagnostic trace — which is precisely the bug this PR set out to eliminate, just triggered by a different cause and now invisible. Recommend logging non-abort errors before returning []:

.catch((e) => {
  if (!controller.signal.aborted) console.error(`scores/aggregate ${id} failed:`, e)
  return [] as ScoreAggregate[]
})
This is the one finding I'd want addressed before merge.

🟡 Controlled Checkbox without onChange — React console warning
Title.tsx (year menu) and FismaTable.tsx (call picker) render <Checkbox checked={checked} /> with no onChange; the toggle is handled by the parent MenuItem's onClick. React will warn "You provided a checked prop to a form field without an onChange handler." Add readOnly (or onChange={() => {}}) to the Checkbox to silence it without double-toggling.

danielbowne:
Good catches, both fixed in f802424. Added console.error on the per-call fetch failures so a dropped call is diagnosable instead of silently zeroing its systems, and marked the menu/filter checkboxes readOnly since the MenuItem/Select owns the toggle. Thanks for the review.

danielbowne:
Pushed a follow-up to how the dashboard resolves a multi-call system. Newest-by-deadline was surfacing a call with no activity yet (the annual ZTM call before anyone worked it), so systems that had completed the year's other call read 0. It now picks, per system, the call they most recently updated (by last-updated timestamp), taking that call's score and progress together, and falls back to newest when a system never updated any call. Deterministic and reflects real status; goes away in 2026 when everyone shares one call.

Separately: some systems (e.g. AGX) show Not Updated despite having full answers, because the imported answers carry no update events and the progress metric counts events, not answers. That's a data-load question (seed the update fields with real submission dates on the final import), not this PR, and I've raised it with the backend/import folks.

danielbowne:
@MackOverflow following up on the dashboard-reads-0 thing.

The multi-call side is fixed in 7e89526: for a system in more than one of a year's calls, the dashboard now shows the call it most recently updated (by last-updated timestamp) rather than the newest by deadline, which was landing on calls with no activity yet.

The remaining piece is data, not UI. Some systems (AGX is the clean example) have full answers but still read Not Updated, because the progress column counts questions updated this cycle via update events, not questions answered, and the imported answers came in without those events. The full fix is to seed the update timestamp on the final data load using the real submission dates. We are planning to seed the dates, so this closes out on the import side.

Field details from what I traced locally, in case it helps the load: the progress metric treats a question as updated when there is a score-update event (an events row with resource public.scores and the scoreid in the payload), and it surfaces lastupdatedat off that. So the import would need to generate those events (or populate whatever lastupdatedat derives from) for the imported scores. Happy to pin down the exact columns/table if that is useful, just say the word.

jsos3-cms (approved):
I feel comfortable bringing this in as is. Good work getting this out. I think it'll be ... finnicky for a while as "time" was different before we brought everyone in a standard through this tool.

PR #495, opened 2026-07-10
Tarratsco/enhancement/add logout button

### LINKED ISSUE #477: Add a logout button to the header for all users (frontend for ztmf#402)
### Feature Type

UI Component

### Workflow Description

ZTMF has no logout button — once signed in, a user has no way to end their session from the UI. This is the frontend slice of CMS-Enterprise/ztmf#402, which adds the backend logout endpoint that clears the `ztmf_session` and ALB OIDC cookies.

Auth in the UI is cookie-only (`withCredentials` axios, no client-side token state), so there is nothing to tear down client-side — the work is the affordance and the call:

- **Where the button goes:** the header user area in `Title.tsx` already has a kebab `IconButton` + MUI `Menu`, but it renders only when `hasAdminRead` is true — non-admin users currently get no menu at all, just the avatar and name. Un-gate the menu so it renders for every logged-in user (the admin items stay conditional inside it) and add a "Log out" `MenuItem` at the bottom. A standalone logout `IconButton` beside the account name is the fallback if un-gating the menu is undesirable.
- **What it does:** call the #402 logout endpoint, then land the user on the sign-in page. Once the cookie is cleared, `authLoader` returns non-200 and `Title.tsx` already renders `LoginPage` — no new routes needed.
- **Cleanup:** the dead `Routes.AUTH_LOGOUT` / `RouteIds.LOGOUT` / `RouteNames.LOGOUT` constants in `src/router/constants.ts` should be either used for this or removed. The unused `VITE_COGNITO_REDIRECT_SIGN_OUT` declaration in `vite-env.d.ts` can go too.

`Title.tsx` has no test file today, so this adds one — follow the mocking pattern in `LoginPage.test.tsx` (and the `renderWithProviders` caveat about the production hash router).

### Acceptance Criteria

- [ ] A logout affordance is visible to **all** logged-in users, not just `hasAdminRead` admins
- [ ] Clicking it calls the CMS-Enterprise/ztmf#402 logout endpoint and lands the user on the sign-in page
- [ ] Signing in again from that page prompts a fresh IdP login (verified on dev)
- [ ] Dead logout route constants in `src/router/constants.ts` are used or removed
- [ ] `Title.test.tsx` added: logout renders for admin and non-admin users, and clicking it triggers the logout flow

### Additional Context

Frontend partner to CMS-Enterprise/ztmf#402 (backend session-clear endpoint). Blocked until #402 is on dev — the button needs a real endpoint to call. The header already distinguishes IdP (Entra/Okta badge), but logout behavior is identical for both: the backend endpoint owns clearing the app and ALB cookies; IdP-side SSO logout is explicitly out of scope in #402.

tarratsco:
## Summary

Closes #477.

Adds a logout affordance to the header for every signed-in user (previously the account menu was gated behind `hasAdminRead`, so non-admins had no way to end their session). Clicking Log out posts to the backend logout endpoint, then forces a full page reload onto `/signin` so the root `authLoader` re-runs against the now-cleared cookie and Title naturally renders LoginPage.

## Backend dependency

Pairs with the backend logout endpoint from https://github.com/CMS-Enterprise/ztmf/pull/408.

## Changes

* **`src/views/Title/Title.tsx`**
  * Removed the `hasAdminRead` gate around the account menu `<IconButton> + <Menu>`. Menu now renders for every user whose session is active (`loaderData.status === 200`). Admin-only menu items (Users, Manage OpDivs, Add Fisma System, Email Users, Create Datacall) stay individually gated.
  * New `handleLogout` async handler:
    * Fires a `notify('Signing out...', 'info')` toast immediately to cover the click → reload window with visible feedback.
    * `POST /auth/logout` with `{ skipAuthHandling: true, timeout: 5000 }`. `skipAuthHandling` short-circuits the centralized 401 interceptor if the session has already expired (avoids a flash of error layout before the reload lands). `timeout: 5000` caps a hung backend so a slow/dead endpoint cannot strand the user.
    * Best-effort: any error is logged and the reload runs anyway, since failing to clear the backend session simply lands the user back on the dashboard.
    * Sets `window.location.hash = Routes.SIGNIN` and calls `window.location.reload()`. The reload is deliberate, because a client-side hash change alone would not re-run the root loader.
  * `IconButton aria-label` renamed from `"more"` → `"Account menu"` to reflect that this is now the everyone-affordance, not admin overflow. `aria-controls` + `<Menu id>` renamed from `"long-menu"` to stay consistent.
* **`src/router/constants.ts`** - removed dead constants: `Routes.LOGOUT`, `Routes.AUTH_LOGOUT`. Never used; superseded by the direct axios call.
* **`src/vite-env.d.ts`** - removed unused `VITE_COGNITO_REDIRECT_SIGN_OUT` declaration. Vestige of a pre-ALB auth model.
* **`src/views/Title/Title.test.tsx`** (new) - 4 tests: admin sees Log out; non-admin sees Log out (and admin-only items are gated out - regression guard); click flow posts with the expected axios config and redirects on success; failed logout still redirects (best-effort).

### Automated

`yarn jest Title.test --no-coverage` - 4/4 passing.

`make check` + `make test` - clean.

### Manual (against deployed dev once ztmf#402 is available)

* [ ] Sign in as an admin (OWNER / HHS_ADMIN). Header shows the account menu icon. Open it → Log out is present at the bottom.
* [ ] Sign in as a non-admin (ISSO / ISSM). Header shows the kebab menu (previously did not). Open it → Log out is present; Users / Manage OpDivs / Add Fisma System / Email Users / Create Datacall menu options are absent.
* [ ] Click Log out → "Signing out..." toast appears briefly → page reloads → lands on `/signin`.
* [ ] Devtools Network: confirm `POST /api/v1/auth/logout` completes with `withCredentials`.
* [ ] Devtools Application → Cookies: confirm `ztmf_session` and the ALB OIDC cookies are gone after logout.
* [ ] Signing in again from `/signin` prompts a fresh IdP login sequence.

## Notes for reviewers

* **Full-page reload after logout is deliberate.** Commented inline in `Title.tsx`: a hash change alone would not re-run the root loader against the cleared cookie. A full reload also guarantees no in-memory session state lingers. Router-based navigation would work too but is more fragile - the reload is the safest way to guarantee clean state.
* **IdP-side SSO logout is out of scope** (per the parent backend ticket). A user who logs out here and then hits the app URL again may get bounced through their existing IdP session without an IdP re-prompt. Documented so the "Sign in again → new IdP prompt" test-plan check is exercised against a fresh browser session (incognito) rather than assuming logout alone forces re-auth.

### DIFF
--- a/src/router/constants.ts
+++ b/src/router/constants.ts
@@ -5,7 +5,6 @@ export enum RouteIds {
   QUESTIONNAIRE = 'questionnaire',
   AUTH = 'auth',
   LOGIN = 'login',
-  LOGOUT = 'logout',
   HOME = 'home',
   DATA = 'data',
   USERS = 'users',
@@ -18,7 +17,6 @@ export enum RouteNames {
   DASHBOARD = 'Dashboard',
   QUESTIONNAIRE = 'questionnaire',
   LOGIN = 'Login',
-  LOGOUT = 'Logout',
   SIGNIN = 'Sign In',
 }
 
@@ -30,7 +28,6 @@ export enum Routes {
   QUESTIONNAIRE = `/${RouteIds.QUESTIONNAIRE}/:fismaacronym/:datacallid?/:pillar?/:function?`,
   AUTH = `/${RouteIds.AUTH}/*`,
   AUTH_LOGIN = `/${RouteIds.AUTH}/${RouteIds.LOGIN}`,
-  AUTH_LOGOUT = `/${RouteIds.AUTH}/${RouteIds.LOGOUT}`,
   SIGNIN = `/${RouteIds.SIGNIN}`,
   SYSTEM_DETAIL = '/systems/:fismasystemid',
   ADMIN_OPDIVS = '/admin/opdivs',
--- a/src/views/Title/Title.test.tsx
+++ b/src/views/Title/Title.test.tsx
@@ -35,6 +35,14 @@ jest.mock('@/views/QuestionnairePage/draftStore', () => ({
   clearOtherUserDrafts: jest.fn(),
 }))
 
+// notify is fired to bridge the gap between click and page reload; mock so
+// the tests can assert the toast without pulling in notistack.
+jest.mock('@/utils/notify', () => ({
+  __esModule: true,
+  notify: jest.fn(),
+  isAuthHandled: jest.fn(),
+}))
+
 // Heavy children the logout affordance does not depend on. Stubbing them keeps
 // the test focused on the header menu and avoids their own asset/network deps.
 jest.mock('@/components/EmailModal/EmailModal', () => ({
@@ -77,6 +85,7 @@ import { useLoaderData, useLocation } from 'react-router-dom'
 import axiosInstance from '@/axiosConfig'
 import { fetchDataCenterEnvironments } from '@/utils/dataCenterEnvironments'
 import { clearOtherUserDrafts } from '@/views/QuestionnairePage/draftStore'
+import { notify } from '@/utils/notify'
 import Title from './Title'
 
 const mockedUseLoaderData = useLoaderData as jest.Mock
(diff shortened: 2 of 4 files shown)


tarratsco:
It's hard to test the logout locally since the Auth token is always present. In this clip Im trying to capture that the logout request is being sent to the BE. In the second attempt I slowed down the connection to be able to stop the page before full reload to be able to have a look at the request in the network tab.


https://github.com/user-attachments/assets/7378e77d-3156-473c-ab9a-b719f115e75f




This will have to be properly tested in the hosted dev environment.

jsos3-cms:
Closed in #497

PR #497, opened 2026-07-10
Enhancement/add logout button

### LINKED ISSUE #477: Add a logout button to the header for all users (frontend for ztmf#402)

a-dipietro:
In-repo version of #495 by @tarratsco, moved so the org-secret CI gates (Snyk,
Infrastructure/Deploy) can run — they can't on a fork PR. Commits pushed unchanged;
authorship preserved.

Closes #477.

## Summary
Adds a logout affordance to the header for every signed-in user (previously the account menu
was gated behind `hasAdminRead`, so non-admins had no way to end their session). Clicking
**Log out** posts to the backend logout endpoint, then forces a full page reload onto
`/signin` so the root `authLoader` re-runs against the now-cleared cookie and `Title` renders
`LoginPage`.

## Backend dependency
Pairs with the backend logout endpoint from CMS-Enterprise/ztmf#408.

## Changes
**`src/views/Title/Title.tsx`**
- Removed the `hasAdminRead` gate around the account menu `<IconButton>` + `<Menu>`. Menu now
  renders for every user whose session is active (`loaderData.status === 200`). Admin-only
  items (Users, Manage OpDivs, Add Fisma System, Email Users, Create Datacall) stay
  individually gated.
- New `handleLogout` async handler: fires a `notify('Signing out...', 'info')` toast
  immediately; `POST /auth/logout` with `{ skipAuthHandling: true, timeout: 5000 }`
  (`skipAuthHandling` short-circuits the centralized 401 interceptor if the session already
  expired; `timeout` caps a hung backend). Best-effort — any error is logged and the reload
  runs anyway. Sets `window.location.hash = Routes.SIGNIN` and calls
  `window.location.reload()` (deliberate: a hash change alone wouldn't re-run the root loader).
- `IconButton` `aria-label` renamed `"more"` → `"Account menu"`; `aria-controls` + `<Menu id>`
  renamed from `"long-menu"` for consistency.

**`src/router/constants.ts`** — removed dead constants `Routes.LOGOUT`, `Routes.AUTH_LOGOUT`
(never used; superseded by the direct axios call).

**`src/vite-env.d.ts`** — removed unused `VITE_COGNITO_REDIRECT_SIGN_OUT` (vestige of a
pre-ALB auth model).

**`src/views/Title/Title.test.tsx`** (new) — 4 tests: admin sees Log out; non-admin sees Log
out and admin-only items are gated out (regression guard); click flow posts with the expected
axios config and redirects on success; failed logout still redirects (best-effort).

## Test plan
Automated: `yarn jest Title.test --no-coverage` — 4/4 passing; `make check` + `make test`
clean. Manual against deployed dev (once the backend logout endpoint is available): admin and
non-admin both see Log out; non-admins do not see admin-only items; click shows the
"Signing out..." toast, reloads, lands on `/signin`; `POST /api/v1/auth/logout` completes
`withCredentials`; session + ALB OIDC cookies cleared afterward.

## Notes for reviewers
Full-page reload after logout is deliberate (a hash change alone wouldn't re-run the root
loader against the cleared cookie, and a reload guarantees no in-memory session state
lingers). IdP-side SSO logout is out of scope per the parent backend ticket — test the
"sign in again → new IdP prompt" check from a fresh/incognito session rather than assuming
logout alone forces re-auth.

Refs #495

### DIFF
--- a/src/router/constants.ts
+++ b/src/router/constants.ts
@@ -5,7 +5,6 @@ export enum RouteIds {
   QUESTIONNAIRE = 'questionnaire',
   AUTH = 'auth',
   LOGIN = 'login',
-  LOGOUT = 'logout',
   HOME = 'home',
   DATA = 'data',
   USERS = 'users',
@@ -18,7 +17,6 @@ export enum RouteNames {
   DASHBOARD = 'Dashboard',
   QUESTIONNAIRE = 'questionnaire',
   LOGIN = 'Login',
-  LOGOUT = 'Logout',
   SIGNIN = 'Sign In',
 }
 
@@ -30,7 +28,6 @@ export enum Routes {
   QUESTIONNAIRE = `/${RouteIds.QUESTIONNAIRE}/:fismaacronym/:datacallid?/:pillar?/:function?`,
   AUTH = `/${RouteIds.AUTH}/*`,
   AUTH_LOGIN = `/${RouteIds.AUTH}/${RouteIds.LOGIN}`,
-  AUTH_LOGOUT = `/${RouteIds.AUTH}/${RouteIds.LOGOUT}`,
   SIGNIN = `/${RouteIds.SIGNIN}`,
   SYSTEM_DETAIL = '/systems/:fismasystemid',
   ADMIN_OPDIVS = '/admin/opdivs',
--- a/src/views/Title/Title.test.tsx
+++ b/src/views/Title/Title.test.tsx
@@ -35,6 +35,14 @@ jest.mock('@/views/QuestionnairePage/draftStore', () => ({
   clearOtherUserDrafts: jest.fn(),
 }))
 
+// notify is fired to bridge the gap between click and page reload; mock so
+// the tests can assert the toast without pulling in notistack.
+jest.mock('@/utils/notify', () => ({
+  __esModule: true,
+  notify: jest.fn(),
+  isAuthHandled: jest.fn(),
+}))
+
 // Heavy children the logout affordance does not depend on. Stubbing them keeps
 // the test focused on the header menu and avoids their own asset/network deps.
 jest.mock('@/components/EmailModal/EmailModal', () => ({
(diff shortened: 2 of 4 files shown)


jsos3-cms (approved):
This works like a charm! Verified in Dev.

:shipit:

PR #503, opened 2026-07-11
Preserve the picked data call across in-survey navigation

### LINKED ISSUE #501: Questionnaire reverts to the latest data call after first in-survey navigation
### Summary

On the dashboard, opening a system's questionnaire for a **specific data call** via the data-call picker works on the first render, but the survey silently reverts to the global/latest data call after the first in-survey navigation (Next, Back, or clicking a function in the sidebar).

### Steps to reproduce

1. From the dashboard, click the Questionnaire action for a system that has scores in more than one open data call (so the picker appears).
2. Pick the non-latest call (e.g. **FY2025 Q3** when **FY25 ZTM** is the latest).
3. The questionnaire opens correctly under FY2025 Q3 (breadcrumb + URL).
4. Click any function in the sidebar, or Next/Back.

**Expected:** the survey stays on FY2025 Q3.
**Actual:** the breadcrumb, URL, and loaded answers revert to the latest call (FY25 ZTM).

### Root cause

The picker passes the chosen call to `QuestionnairePage` through `location.state` (`datacallid`, `datacall`, `deadline`). The page's load effect honors that on first render.

However, the four in-questionnaire `navigate()` calls — Next, Back, the sidebar item click, and the initial canonical-URL redirect — re-set router state to only `{ fismasystemid }`, dropping the data-call fields. On the next render `location.state.datacallid`/`datacall` are `undefined`, so the load effect falls through to the branch that uses the global data-call selector (`selectedDatacall` / `latestDataCallId`) and reverts the survey to the latest call.

### Fix

Persist the resolved data-call context (`datacallid` / `datacall` / `deadline`) in a ref when the load effect resolves it, and re-supply it in the state of every internal `navigate()` call so the chosen call survives click-through.

This is already implemented and verified on the `impl` environment branch (`feature/impl-question-insights`) and should be ported to `main`. The change is isolated to `src/views/QuestionnairePage/QuestionnairePage.tsx`.

danielbowne:
## What

Fixes #501. Opening a system's questionnaire for a specific data call from the dashboard picker works on first render, but the survey silently reverts to the latest data call after the first in-survey navigation (Next, Back, or a sidebar function click).

## Why

The picker passes the chosen call to `QuestionnairePage` via router `location.state` (`datacallid`, `datacall`, `deadline`), and the load effect honors it on first render. But the four in-questionnaire `navigate()` calls (Next, Back, the sidebar item, and the initial canonical-URL redirect) re-set state to only `{ fismasystemid }`, dropping the data-call fields. On the next render `location.state.datacallid`/`datacall` are `undefined`, so the load effect falls through to the global data-call selector and reverts the survey to the latest call.

## Fix

Persist the resolved data-call context (`datacallid` / `datacall` / `deadline`) in a ref when the load effect resolves it, and re-supply it in the state of every internal `navigate()` call so the chosen call survives click-through. Isolated to `src/views/QuestionnairePage/QuestionnairePage.tsx`.

## Testing

- `yarn typecheck` passes.
- Manually verified on the impl environment: pick a non-latest call (FY2025 Q3 with FY25 ZTM latest), open the questionnaire, click through functions — the survey stays on FY2025 Q3 (breadcrumb, URL, and answers).

### DIFF
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -175,6 +175,17 @@ export default function QuestionnarePage() {
   optionsRef.current = options
   const loadingQuestionRef = React.useRef(loadingQuestion)
   loadingQuestionRef.current = loadingQuestion
+  // Persisted data-call context so in-questionnaire navigation (Next/Back/
+  // sidebar) can re-supply it via router state. Without this, the first
+  // click-through drops location.state.datacall* and the load effect reverts to
+  // the global data-call selector — opening a system for a specific picked call
+  // (e.g. FY2025 Q3) would silently fall back to the latest call (FY25) on the
+  // next click.
+  const datacallStateRef = React.useRef<{
+    datacallid: number
+    datacall: string
+    deadline?: string
+  }>({ datacallid: 0, datacall: '' })
   // Bumped when a re-seed changes the answer so the uncontrolled radio ChoiceList
   // (which only reflects defaultChecked on mount) remounts and shows the
   // corrected selection.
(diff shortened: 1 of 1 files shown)


danielbowne:
Closing as a duplicate of #506 — same data call persistence fix for #501. Keeping #506 since it's the leaner version.

PR #506, opened 2026-07-11
Keep the chosen data call through in-survey navigation

### LINKED ISSUE #501: Questionnaire reverts to the latest data call after first in-survey navigation

danielbowne:
Closes #501.

## Problem

Opening a system's questionnaire for a specific data call via the dashboard picker works on first render, but the survey reverts to the global/latest call after the first in-survey navigation (Next, Back, or a sidebar click).

## Root cause

The picker passes the chosen call to QuestionnairePage through `location.state` (`datacallid` / `datacall` / `deadline`), and the load effect honors it on first render. But the four internal `navigate()` calls (Next, Back, sidebar item, and the canonical-URL redirect) re-set router state to only `{ fismasystemid }`, dropping the data-call fields. On the next render those fields are `undefined`, so the load effect falls through to the global-selector branch and reverts to the latest call.

## Fix

Persist the resolved data call in a ref when the load effect resolves it, and spread it into the state of every internal `navigate()` so the choice survives click-through. Isolated to `QuestionnairePage.tsx`.

This mirrors the fix already verified on the impl environment; ported here to main.

## Testing

QuestionnairePage suite (75) and the full suite (346) pass; typecheck and lint clean.

### DIFF
--- a/src/views/QuestionnairePage/QuestionnairePage.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.tsx
@@ -242,6 +242,15 @@ export default function QuestionnarePage() {
   const routeDeadline = location.state?.deadline as string | undefined
   const systemRef = React.useRef(system)
   systemRef.current = system
+  // The in-survey navigate() calls (Next, Back, sidebar, canonical redirect)
+  // reset router state and would drop the chosen data call, reverting the
+  // survey to the latest call (#501). Persist the resolved call here and
+  // re-supply it in every internal navigation so the choice survives.
+  const datacallStateRef = React.useRef<{
+    datacallid?: number
+    datacall?: string
+    deadline?: string
+  }>({})
   const systemInfo = fismaSystems.find((s) => s.fismasystemid === system)
   const systemName = systemInfo?.fismaname ?? fismaacronym ?? ''
   // Resolve the system's raw datacenter environment to its scoring category
@@ -356,6 +365,11 @@ export default function QuestionnarePage() {
             setIsPastDeadline(
               routeDeadline ? new Date() > new Date(routeDeadline) : true
             )
+            datacallStateRef.current = {
+              datacallid: routeDatacallId,
+              datacall: routeDatacall,
+              deadline: routeDeadline,
+            }
           } else {
             const isHistorical =
               selectedDatacall !== null &&
@@ -365,6 +379,11 @@ export default function QuestionnarePage() {
               setDatacall(datacall)
               setIsPastDeadline(true)
               activeDataCallId = selectedDatacall.datacallid
+              datacallStateRef.current = {
+                datacallid: selectedDatacall.datacallid,
+                datacall: selectedDatacall.datacall,
+                deadline: selectedDatacall.deadline,
+              }
             } else {
               datacall = latestDatacall.replaceAll(' ', '_')
               setDatacall(datacall)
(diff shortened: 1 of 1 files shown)


MackOverflow (commented):
@danielbowne Solid, well-scoped fix for the reported bug — good comments and the right ref-based approach (a ref avoids adding a render-triggering effect dependency). One question gates my approval: because the canonical redirect and all four internal navigate() calls now persist datacallid in location.state, once you enter via the dashboard picker the load effect's first branch (routeDatacallId && routeDatacall) stays truthy on every subsequent run, which means a later change via the global data-call selector re-runs the effect but gets ignored — the historical/selector branch is now unreachable on this page. Previously the redirect cleared that state, which is what let the selector take over. If the selector isn't meant to override a dashboard-sourced call on this page, this is a clean approve; otherwise the first branch needs a way to yield to an explicit selector change. (Minor, non-blocking: on deep links the redirect now flips routeDatacallId from undefined to an id, triggering one extra effect run whose fetches are aborted by the existing AbortController — worth a note but not a concern.)

danielbowne:
Thanks Mack, good eye. That behavior is intended for this pass. The reason it has to work this way: persisting the picked call through Next/Back/sidebar is the whole fix for #501, and that persisted state is the same thing the selector used to clear, so we can't both keep the picked call across navigation and let the selector wipe it without extra logic. Since opening a system's questionnaire for a specific call is a deliberate choice, having it stick felt like the right default, and the Title selector stays a dashboard control. Anyone who wants a different call for that system just re-opens via the picker.

If we do want the selector to switch the call mid-survey later, we'd need to tell a real selector change apart from its default value so it doesn't clobber the pick on load. Filed that as a future enhancement: #511. Good to approve on your end?

MackOverflow (approved):
LGMT

PR #514, opened 2026-07-13
ZTMF Insights panel on the questionnaire

danielbowne:
Promotes the ZTMF Insights panel from impl to dev (and, on merge, prod).

## What it does

Alongside each questionnaire question, the panel surfaces the evidence ZTMF maps to that question from the security tooling it ingests — Kion, SecurityHub, Hardenize, CFACTS, ARS — plus prior-year context, and badges the answer options that match a suggested or prior score.

It is **purely informational**: it never changes, gates, or auto-sets the score. The responder still chooses; the panel just puts the evidence next to the answer.

## Invisible when off

The panel renders only for systems whose OpDiv is entitled and only when the insights endpoint returns data. Systems without insights (most OpDivs at first, since rollout is CMS-first) see no change to the questionnaire — no empty panel, no "no data" banner.

## The remediation flag

Includes a build-time flag, `INSIGHTS_SUGGEST_FIX_ENABLED`, that gates the "How to fix" remediation text on a finding. `ui.yml` sets it true **only on the impl build**; dev and prod ship Insights **without** the remediation while that piece is being trialed. This is intentional.

## Safety

The panel is wrapped in an error boundary and renders nothing on any malformed payload, so a bad insight can never blank the questionnaire. The payload is treated as opaque and additive (every field optional, arrays guarded).

## Testing

Full suite green (376 passing, InsightsPanel adds ~30), typecheck and lint clean. This PR deploys to dev for validation before any merge to prod.

### DIFF
--- a/.github/workflows/ui.yml
+++ b/.github/workflows/ui.yml
@@ -58,6 +58,8 @@ jobs:
         # Default off; flip per environment as IdP rollout progresses.
         env:
           VITE_IDP_ENABLED: ${{ inputs.environment == 'dev' && 'true' || 'false' }}
+          # Insights "How to fix" remediation — trialed in impl only; off in dev/prod.
+          VITE_INSIGHTS_SUGGEST_FIX_ENABLED: ${{ inputs.environment == 'impl' && 'true' || 'false' }}
           # Dev-only banner overrides. Sourced from secrets so the feedback form
           # URL and contact address stay out of the repo. Clear these secrets
           # when the testing window ends and the banner reverts to default copy.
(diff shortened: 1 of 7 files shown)


MackOverflow (approved):
looks good, two nits, both non-blocking, but worth considering before we roll to prod. 

- **Hardcoded hex colors throughout**, bypassing the MUI theme (`#f8f9fe`, `#5666b8`, `#b02a37`, …). This is the biggest maintainability concern: no theme tokens means no dark-mode support and manual contrast management. If the rest of the app themes via MUI palette, this panel will drift. Acceptable for a prototype-derived trial component, but I'd file a follow-up to migrate to theme tokens before prod hardening.
- **`MATURITY_LABEL` (1→Traditional … 4→Optimal) is redefined locally.** These maturity labels almost certainly exist elsewhere in the questionnaire codebase. If so, this is a DRY risk — a future label change has two homes. Worth a quick grep to confirm there isn't an existing constant to reuse.

PR #519, opened 2026-07-13
Show non-satisfied ARS controls greyed in insights panel

### LINKED ISSUE #518: InsightsPanel: show non-satisfied ARS controls, greyed out
### Type
Enhancement (UX) — ZTMF Insights

### Observed
On a system's questionnaire insight , the ARS section reads "1 of 4 satisfied" but only the one satisfied control renders as a pill. The three non-satisfied controls are not shown, so the reader can't see which controls are missing.

### Want
Show all of the controls behind the count. Satisfied controls keep their current (green) treatment; the non-satisfied ones render **greyed out / muted** — not red. "Not satisfied" is informational, not a failure to alarm on, consistent with the panel's informational-only framing.

### Frontend (`src/views/QuestionnairePage/InsightsPanel/InsightsPanel.tsx`)
The panel already renders ARS control pills: `ars_satisfied_controls` as green `ControlChip passed`, and `ars_failing_controls` as red `ControlChip passed={false}` (~lines 311-312, 421-441, `ControlChip` ~494). Changes:
1. Render the non-satisfied controls in a muted/grey style rather than red.
2. Keep the graceful fallback: when the non-satisfied IDs aren't in the payload, fall back to today's count-only line ("X of Y satisfied") — no empty state.

### Data dependency (prerequisite — backend / insights pipeline)
The payload rarely carries the non-satisfied control IDs today. Across the current `system_insights` data: **223 rows have `ars_satisfied_controls` but only 13 have `ars_failing_controls`**. So for most systems the non-satisfied control IDs simply aren't in the payload, and the UI has nothing to grey out. The full fix needs the pipeline to emit the non-satisfied ARS control IDs for every system that has ARS counts (satisfied vs total), the same way it emits the satisfied ones. This is a companion backend/insights item.

### Acceptance
- When the payload carries them, all ARS controls behind the count render: satisfied in the current style, non-satisfied greyed out (not red).
- The count line still matches the pills (all `ars_controls_total` controls represented).
- When the non-satisfied IDs are absent, the panel degrades to the current count-only display with no empty state.

danielbowne:
## What

Shows the non-satisfied ARS controls behind the "X of Y satisfied" count in the questionnaire insights panel, greyed out. Before, only the satisfied controls rendered as pills, so a reader seeing "1 of 4 satisfied" (e.g. EPPEC) couldn't tell which three controls were missing.

Now all controls behind the count are visible:
- satisfied → green with a ✓ (unchanged)
- non-satisfied → grey with a ○ (new — informational, not an alarm)
- failing → red with a ✗ (unchanged; rare, Archer-explicit fails)

## Why grey, not red

"Not satisfied" is informational, consistent with the panel's informational-only framing. Most non-satisfied controls are applicable-but-unassessed, not failures — so they get a muted neutral treatment, distinct from the red "failing" state.

## Backend dependency (already landed)

Rides a new payload field `ars_not_satisfied_controls`, emitted by the insights view so that `satisfied + not_satisfied = total`. The field is null on non-ARS questions, so the panel keeps its existing count-only fallback with no empty state.

## Changes

- `types.ts` — add `ars_not_satisfied_controls?: string[] | null` to the insight payload.
- `InsightsPanel.tsx` — render the grey pills between green and red; `ControlChip` goes from a binary satisfied/failing prop to a three-state variant. Reuses the existing `toStringArray` guard, so a malformed payload can't throw.
- tests — cover the three-state render and the grey (non-red) marker; the count-only fallback test stays green.

## Test

System **EPPEC**, Identity pillar, question 1 ("How does the system authenticate user identity?"). Expand the insight details → ARS Controls: "1 of 4 satisfied", IA-01 green, IA-02(01)/IA-02(02)/IA-02(08) grey.

Closes #518

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -452,6 +452,10 @@ export type InsightPayload = {
   // pipeline; may be undefined (before it ships / on non-ARS questions), null,
   // or []. `ars_satisfied_controls` length == `ars_controls_satisfied`.
   ars_satisfied_controls?: string[] | null
+  // Applicable-but-not-satisfied controls (applicable − satisfied). Informational,
+  // rendered greyed — distinct from `ars_failing_controls` (Archer-explicit fails).
+  // satisfied + not_satisfied == ars_controls_total when the pipeline emits them.
+  ars_not_satisfied_controls?: string[] | null
   ars_failing_controls?: string[] | null
 
   findings?: InsightFindings
--- a/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
+++ b/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
@@ -229,12 +229,13 @@ describe('InsightsPanel', () => {
     expect(screen.getByText(/IAM\.10/)).toBeInTheDocument()
   })
 
-  it('lists satisfied and failing ARS control IDs when the pipeline provides them', () => {
+  it('lists satisfied, non-satisfied, and failing ARS control IDs when the pipeline provides them', () => {
     render(
       <InsightsPanel
         payload={{
           ...fullPayload,
           ars_satisfied_controls: ['IA-01', 'IA-02(01)'],
+          ars_not_satisfied_controls: ['IA-02(02)', 'IA-02(08)'],
           ars_failing_controls: ['AC-17'],
         }}
       />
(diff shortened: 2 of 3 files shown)


MackOverflow (approved):
lgtm!

PR #520, opened 2026-07-13
feat(pillar-scores): label radar series with actual data-call names (#468)

### LINKED ISSUE #468: Radar chart: label series with actual data-call names instead of Current/Previous

jsos3-cms:
Rehome of #493 (by @costacalvin) onto an in-repo branch so the CI and dev-deploy workflows — which fork PRs can't trigger, since they can't reach the org secrets — run against these changes. This branch keeps @costacalvin's original commit; the only addition is the follow-up fix described below.

Closes #468.

## What this does

Labels the Pillar Scores radar chart, tooltip, cards, and table with the actual data-call names (with a `· CMS` / `· HHS` tenant suffix) instead of generic "Current"/"Previous". Adds a "Compare with" picker so a reviewer can choose which prior data call to compare against (with a "None" option), resolves the default "previous" call by deadline order rather than by datacall id (per #393), and threads the data call each dashboard row is displaying into the modal so it opens anchored to that call. Guards the comparison series against rendering an all-zero radar when there is nothing comparable.

## Follow-up fix in this branch

The comparison — both the default and the dropdown options — was originally sourced from the full data-call list across all tenants. CMS quarterly calls (`FY## Q#`) and HHS annual ZTM calls (`FY## ZTM`) interleave by deadline, so the nearest-by-deadline call was frequently one the system can never have a score for. This hit HHS/OpDiv systems hardest: their only call is the annual ZTM, so the default comparison landed on a CMS quarterly and showed "No data" every time, with no user action. This branch scopes both the default predecessor and the dropdown to the calls the system actually has a score for, keeping the #393 deadline ordering.

## Design decisions (answering the questions raised on #493)

- **Single comparison dropdown, not two selectors.** The modal opens from a dashboard row whose data call is already fixed (the row's most-recently-updated call), so that row is effectively the "current" side; a second selector would let the two sides drift out of sync with the row that was clicked and would duplicate the dashboard's own call selector. The two-selector pattern fits the Questionnaire compare view because neither side is pre-anchored there.
- **Kept the Select rather than switching to Autocomplete.** It matches how data calls are selected elsewhere — the dashboard's primary data-call filter is also a Select, whereas Autocomplete is reserved for the large entity lists (systems, users, OpDivs). If the list grows enough to want search, the on-idiom next step would be a Select with fiscal-year subheaders rather than Autocomplete. Open to your input if you feel the list is already long enough to warrant it.

## Test plan

- [ ] `yarn typecheck` clean
- [ ] `yarn test` green (aggregateScores, sortDatacallsByDeadline, datacallGrouping)
- [ ] CMS system with 2+ quarterly submissions: default comparison is the prior quarter; dropdown lists only that system's scored calls
- [ ] HHS/OpDiv system: default comparison is the prior ZTM (or "Current" only when there is no prior), never a CMS quarterly "No data"
- [ ] "None" hides the comparison series; picking a scored call with no pillar data shows "No data for {name}"

### DIFF
--- a/src/components/PillarScoresModal/PillarScoresModal.tsx
+++ b/src/components/PillarScoresModal/PillarScoresModal.tsx
@@ -125,8 +125,8 @@ const PillarScoresModal: React.FC<PillarScoresModalProps> = ({
 
   // True whenever a comparison is active (default or user-chosen call),
   // false when the user explicitly picked "None" or there is no predecessor
-  // at all. Distinguishes "selected a call with no score → show No previous
-  // data" from "None / no comparison possible → hide entirely".
+  // at all. Distinguishes "selected a call with no score → show 'No data for <name>'"
+  // from "None / no comparison possible → hide entirely".
   const comparisonActive =
     selectedComparisonId !== null &&
     (selectedComparisonId !== undefined || previousDatacall !== null)
@@ -475,7 +475,7 @@ const PillarScoresModal: React.FC<PillarScoresModalProps> = ({
                             mb: 0.3,
                           }}
                         >
-                          {`${comparisonDatacallName}: ${previousSystemScore.toFixed(2)}`}
+                          {`Compared with ${comparisonDatacallName}: ${previousSystemScore.toFixed(2)}`}
                         </Typography>
                         <Typography
                           variant="caption"
@@ -504,7 +504,7 @@ const PillarScoresModal: React.FC<PillarScoresModalProps> = ({
                           fontSize: '0.85rem',
                         }}
                       >
-                        No previous data
+                        No data for {comparisonDatacallName}
                       </Typography>
                     )
                   }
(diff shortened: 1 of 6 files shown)


MackOverflow (approved):
no nits

PR #523, opened 2026-07-14
fix(questionnaire): make questionnaire deep links self-addressable (#500)

voidspooks:
## Summary

Fixes #500. A questionnaire URL (`/#/questionnaire/:fismaacronym/:datacallid/:pillar/:function`) only worked when reached by clicking through from the dashboard. Pasting, refreshing, or bookmarking one failed — the page could not resolve the system, and any pillar/function in the URL was discarded. This is entirely client-side routing (the API already serves questionnaire/score/insight data by `fismasystemid` on a direct request), so there is no backend change.

`QuestionnairePage` sourced `fismasystemid` only from React Router `location.state` (empty on a cold load) and, after categories loaded, unconditionally redirected to the first pillar/function.

## Changes

All in `src/views/QuestionnairePage/`:

- **Resolve the system** from `:fismaacronym` against the already-loaded systems list when `location.state` is absent. While that list is still loading the page shows a spinner; the "cannot load" warning now appears only once systems are loaded and the acronym is still unmatched (reworded to a not-found message).
- **Resolve the cycle** from the URL's data-call segment via the `datacalls` list (already on context), falling back to the selected/latest call when the segment is absent or unrecognized.
- **Honor `:pillar/:function`** on mount, falling back to `categoriesData[0]` only when they are missing or invalid.
- **Gate the fetch** on the datacall context being ready (`latestDataCallId > 0`) so a cold link cannot query scores with `datacallid=0` before datacalls load.

The deep-link branch only runs when router state is absent, so the dashboard → questionnaire path is unchanged.

## Tests

- `deepLink.ts` — resolution logic extracted as pure functions (`resolveSystemIdByAcronym`, `resolveDatacallBySlug`, `resolveFunctionTarget`, shared `toSlug`), unit-tested in `deepLink.test.ts`.
- `QuestionnairePage.deeplink.test.tsx` — mounts the real page at a cold deep link (no `location.state`) and asserts the system resolves, the cycle is read from the URL (scores queried for the URL's call, not the latest), the deep-linked pillar/function opens (not the first), plus the fallback / not-found / loading paths.

Full frontend suite green locally: `tsc --noEmit`, `eslint` (0 errors), `jest` (394 passed).

Closes #500.

PR #524, opened 2026-07-14
fix(questionnaire): make questionnaire deep links self-addressable (#500)

### LINKED ISSUE #500: Questionnaire deep-links don't load: route relies on in-memory location.state and force-redirects to the first function
## Description

A questionnaire URL (`/#/questionnaire/:fismaacronym/:datacallid/:pillar/:function`) only works when reached by clicking through from the dashboard. Pasting, refreshing, or bookmarking one fails: the page cannot resolve the system, and any pillar/function in the URL is discarded. There are effectively no shareable or bookmarkable questionnaire links.

This is entirely client-side routing. The API serves questionnaire, score, and insight data by `fismasystemid` on a direct request, so no backend change is involved - it is a frontend routing/state problem.

## Root cause

`src/views/QuestionnairePage/QuestionnairePage.tsx`:

1. The system id and data-call come only from React Router `location.state`:
   - `const system = location.state?.fismasystemid` (~L236)
   - `location.state?.datacallid / datacall / deadline` (~L240-242)

   `location.state` is populated during in-app navigation and is empty on a cold load (paste / refresh / bookmark). There is no fallback that resolves the `:fismaacronym` URL param to a `fismasystemid` (e.g. via the systems list the app already loads), so `system` is `undefined` and the page cannot load.

2. After categories load, the page unconditionally redirects to the first category/function, discarding the URL's `:pillar/:function`:
   - `navigate(.../${toSlug(categoriesData[0].name)}/${toSlug(categoriesData[0].steps[0].function.function)}, { replace: true })` (~L446)

   So even when the system does resolve, you land on pillar[0]/function[0], not the deep-linked target.

## Steps to reproduce

1. Open any system's questionnaire from the dashboard and copy the URL (`/#/questionnaire/<acronym>/<datacall>/<pillar>/<function>`).
2. Open that URL in a new tab, or refresh the page.
3. The deep-linked question does not load - the system fails to resolve and/or the page redirects to the first function.

## Expected behavior

A questionnaire URL is self-addressable: it resolves the system from `:fismaacronym`, the cycle from `:datacallid`, and opens the `:pillar/:function` named in the URL.

## Proposed fix

1. On cold load (no `location.state`), resolve `fismasystemid` and the data-call from the URL params by looking them up in the systems / data-calls data the app already fetches, instead of depending on `location.state`.
2. Honor the `:pillar/:function` URL params on mount; fall back to `categoriesData[0]` only when they are absent or invalid.

## Impact

No shareable or bookmarkable links to a system, pillar, or function - including deep-linking someone straight to a specific question or insight for review.

Affected: `src/views/QuestionnairePage/QuestionnairePage.tsx` (~L231-246, ~L446) and the route definition in `src/router/constants.ts`. Present on `main`.

MackOverflow:
> **Note:** In-repo copy of #523 by @voidspooks. Moved from the fork so Snyk
>
> Refs #523

---

## Summary

Fixes #500. A questionnaire URL (`/#/questionnaire/:fismaacronym/:datacallid/:pillar/:function`) only worked when reached by clicking through from the dashboard. Pasting, refreshing, or bookmarking one failed — the page could not resolve the system, and any pillar/function in the URL was discarded. This is entirely client-side routing (the API already serves questionnaire/score/insight data by `fismasystemid` on a direct request), so there is no backend change.

`QuestionnairePage` sourced `fismasystemid` only from React Router `location.state` (empty on a cold load) and, after categories loaded, unconditionally redirected to the first pillar/function.

## Changes

All in `src/views/QuestionnairePage/`:

- **Resolve the system** from `:fismaacronym` against the already-loaded systems list when `location.state` is absent. While that list is still loading the page shows a spinner; the "cannot load" warning now appears only once systems are loaded and the acronym is still unmatched (reworded to a not-found message).
- **Resolve the cycle** from the URL's data-call segment via the `datacalls` list (already on context), falling back to the selected/latest call when the segment is absent or unrecognized.
- **Honor `:pillar/:function`** on mount, falling back to `categoriesData[0]` only when they are missing or invalid.
- **Gate the fetch** on the datacall context being ready (`latestDataCallId > 0`) so a cold link cannot query scores with `datacallid=0` before datacalls load.

The deep-link branch only runs when router state is absent, so the dashboard → questionnaire path is unchanged.

## Tests

- `deepLink.ts` — resolution logic extracted as pure functions (`resolveSystemIdByAcronym`, `resolveDatacallBySlug`, `resolveFunctionTarget`, shared `toSlug`), unit-tested in `deepLink.test.ts`.
- `QuestionnairePage.deeplink.test.tsx` — mounts the real page at a cold deep link (no `location.state`) and asserts the system resolves, the cycle is read from the URL (scores queried for the URL's call, not the latest), the deep-linked pillar/function opens (not the first), plus the fallback / not-found / loading paths.

Full frontend suite green locally: `tsc --noEmit`, `eslint` (0 errors), `jest` (394 passed).

Closes #500.

### DIFF
--- a/src/views/QuestionnairePage/QuestionnairePage.deeplink.test.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.deeplink.test.tsx
@@ -1,0 +1,222 @@
+import { render, screen, waitFor } from '@testing-library/react'
+import { MemoryRouter, Routes, Route } from 'react-router-dom'
+import { Routes as AppRoutes } from '@/router/constants'
+import QuestionnairePage from './QuestionnairePage'
+import type { userData } from '@/types'
+
+// Deep-link integration test for #500: a questionnaire URL reached by paste /
+// refresh / bookmark (no router location.state) must resolve the system from
+// :fismaacronym, the cycle from the datacall segment, and open the named
+// :pillar/:function — not fail to load or snap to the first function.
+
+// config.ts reads import.meta.env, which jest can't parse; only the insight
+// feature flag is reachable from this page, so a minimal default is enough.
+jest.mock('@/utils/config', () => ({
+  __esModule: true,
+  default: { INSIGHTS_SUGGEST_FIX_ENABLED: false },
+}))
+
+jest.mock('@/axiosConfig', () => ({
+  __esModule: true,
+  default: { get: jest.fn(), post: jest.fn(), put: jest.fn() },
+}))
+const mockGet = require('@/axiosConfig').default.get as jest.Mock
+
+// draftStore uses crypto.subtle (unavailable in jsdom); stub to no-ops.
+jest.mock('./draftStore', () => ({
+  saveDraft: jest.fn().mockResolvedValue(true),
+  loadDraft: jest.fn().mockResolvedValue(null),
+  clearDraft: jest.fn().mockResolvedValue(undefined),
+}))
+
+// The page reads its shared state via useContextProp (useOutletContext).
+let mockCtx: Record<string, unknown>
+jest.mock('../Title/Context', () => ({
+  useContextProp: () => mockCtx,
+}))
+
+// Seed shape mirrors GET /fismasystems/:id/questions (empire data, SSD-EX).
+const PILLARS: [string, number, string][] = [
+  ['Identity', 7006, 'Imperial Identity Verification'],
+  ['Devices', 7001, 'Imperial Device Management'],
+  ['Networks', 7003, 'Imperial Network Security'],
+  ['Applications', 7002, 'Fleet Application Security'],
+  ['Data', 7004, 'Fleet Data Protection'],
+  ['CrossCutting', 7005, 'Imperial Cross-Cutting Controls'],
+]
+
+const QUESTIONS = PILLARS.map(([pillar, functionid, fn], i) => ({
+  questionid: 900 + i,
+  question: `Question for ${fn}`,
+  notesprompt: 'Notes',
+  pillar: { pillar },
+  function: {
+    functionid,
+    function: fn,
+    description: `${fn} description`,
+    datacenterenvironment: 'Imperial-Fleet',
+  },
+}))
+
+beforeEach(() => {
+  jest.clearAllMocks()
+  mockCtx = {
+    userInfo: {
+      userid: '1',
+      email: 'grand.moff@deathstar.empire',
+      fullname: 'Grand Moff Tarkin',
+      role: 'OWNER',
+    } as userData,
+    // Latest is call 5; the deep link below names call 4 — so a scores query
+    // for datacallid=4 proves the cycle was resolved from the URL, not defaulted.
+    latestDataCallId: 5,
+    latestDatacall: 'Audit Fields Smoke Cycle',
+    latestDeadline: '2099-12-31T23:59:59Z',
+    selectedDatacall: {
+      datacallid: 5,
+      datacall: 'Audit Fields Smoke Cycle',
+      datecreated: '',
+      deadline: '2099-12-31T23:59:59Z',
+    },
+    datacalls: [
+      {
+        datacallid: 5,
+        datacall: 'Audit Fields Smoke Cycle',
+        datecreated: '',
+        deadline: '2099-12-31T23:59:59Z',
+      },
+      {
+        datacallid: 4,
+        datacall: 'FY2025 Death Star Assessment',
+        datecreated: '',
+        deadline: '2025-03-31T23:59:59Z',
+      },
+    ],
+    activeDatacallIds: [5],
+    fismaSystems: [
+      {
+        fismasystemid: 1002,
+        fismaacronym: 'SSD-EX',
+        fismaname: 'Super Star Destroyer Executor Command Systems',
+        datacenterenvironment: 'Imperial-Fleet',
+      },
+    ],
+    setFismaSystems: jest.fn(),
+    showDecommissioned: false,
+    setShowDecommissioned: jest.fn(),
+    fetchFismaSystems: jest.fn(),
+    datacenterEnvironments: [],
+  }
+
+  mockGet.mockImplementation((url: string) => {
+    if (url.includes('/questions'))
+      return Promise.resolve({ data: { data: QUESTIONS } })
+    if (url.startsWith('scores')) return Promise.resolve({ data: { data: [] } })
+    if (url.includes('/options')) return Promise.resolve({ data: { data: [] } })
+    if (url.includes('insights')) return Promise.resolve({ data: { data: [] } })
+    return Promise.resolve({ data: { data: [] } })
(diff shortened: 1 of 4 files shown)


jsos3-cms:
## Code review

Nice work — the resolution logic is cleanly extracted into pure, well-tested functions, the slug round-trip for pillar/function is consistent, and the click-through path is untouched. Two things to address before merge, plus a couple of smaller suggestions.

### 🟠 Every cold deep-link runs the whole fetch effect twice

`QuestionnairePage.tsx` — the data-fetch effect's dependency array includes `routeDatacallId`/`routeDatacall`/`routeDeadline` (derived from `location.state`, lines 298-300). On a cold load these start `undefined`, so the first run takes the deep-link branch and its `navigate(url, { state: { fismasystemid, ...datacallStateRef.current }, replace: true })` (line ~605) writes `location.state` for the first time — flipping those three deps from `undefined` to real values and re-running the entire effect (which then takes the "opened from dashboard" branch at line ~453).

Reproduced locally with a probe on the deep-link test: `/fismasystems/:id/questions`, `/scores`, and `/functions/:id/options` each fire **twice** for a single cold-load mount, with a second `navigate()` and a late `setLoadingQuestion(true)` (jest flags the second wave with `act(...)` warnings). It converges after two runs and the end state is correct, so it isn't data-corrupting — but it doubles backend load on `/questions` and `/scores` for every paste/refresh/bookmark (the exact path this PR ships) and can flicker the spinner after content has rendered.

The `:pillar`/`:function` params already dodge this by being read through refs so they don't enroll as effect deps; the same guard just wasn't applied to the `location.state`-derived route values, and the effect's own `navigate()` is what mutates that state. Suggested fix: read `routeDatacallId`/`routeDatacall` through refs like the slug params, or gate the deep-link branch behind a "resolved-once" ref so the self-written state doesn't re-trigger the fetch.

### 🟠 Deep-link to a decommissioned system reads as "not found / no access"

`resolveSystemIdByAcronym` only searches the `fismaSystems` list, which the `Title` layout loads without `?decommissioned=true` by default (`showDecommissioned` starts `false`, `Title.tsx:86,94-98`) and replaces on each mount. So a cold link to a real-but-decommissioned system won't resolve and shows the new *"Could not find a system matching '{acronym}'. It may not exist, or you may not have access to it."* — bypassing the page's own pre-existing "No questionnaire is available for this system" handling for decommissioned systems, and telling the user it doesn't exist / they lack access when neither is true. Not a regression (no direct link worked before), but the wording actively misleads. At minimum, soften the copy so it doesn't assert non-existence/no-access; ideally have resolution consider the decommissioned list before concluding not-found.

### 🟡 `resolveDatacallBySlug` is case-sensitive while the other resolvers aren't

`deepLink.ts` — `resolveSystemIdByAcronym` and `toSlug` both lowercase, but `resolveDatacallBySlug` does an exact `dc.datacall.replaceAll(' ', '_') === slug` match. A case-mangled shared/retyped URL silently falls back to the latest/selected cycle rather than surfacing a mismatch — and quietly showing the *wrong* cycle's scores is a worse failure mode than an explicit miss here. Suggest lowercasing both sides, consistent with the other resolvers.

### ⚪ Latent: underscore collision in datacall slugs

Only spaces are encoded (`→ _`), so a datacall name containing a literal underscore would round-trip ambiguously (e.g. `"FY_2025 Q4"` and `"FY 2025_Q4"` both slugify to `FY_2025_Q4`, and `.find()` picks the first). Not reachable under current naming (no literal underscores) — noting for the record, not blocking.

### Confirmed working

- Pillar/function slug round-trip is consistent (shared `toSlug` on both the `navigate()` write and the `resolveFunctionTarget` read).
- Click-through from the dashboard is behaviorally unchanged: it ships `fismasystemid`/`datacallid`/`datacall`/`deadline` in one `navigate()`, so the route-state deps are defined from the first render (the double-run can't occur there) and state takes precedence over acronym resolution.
- No infinite loop — the double-run stabilizes at exactly two executions.

---

Happy to apply these fixes to the in-repo branch directly, or if you'd rather fix them on your side, push to the fork and we'll rehome the additional commits onto this branch — whichever you prefer. Items 1 and 2 are the ones worth doing before merge; 3 and 4 are optional.

voidspooks:
Thanks for the thorough review — all four findings verified and fixed in voidspooks/ztmf-ui@14c57ba on the fork branch (`fix/questionnaire-deeplink-500`), ready to rehome onto this one. All items addressed, including the latent underscore collision:

1. **Double-run** — confirmed with an exact-count regression test (the existing `.some()`-style assertions were call-count-blind, which is how it slipped through). Rather than reading the route-state values through refs — `routeDatacallId` being a dep is what lets a picker-driven datacall switch refetch without a remount (#467/#501), so refs there risked breaking that flow — the fix stops the effect from self-writing the state: `datacallStateRef` is now populated only on the dashboard path, and URL-driven flows leave it empty since the URL's datacall segment already carries the cycle. `ScoreDiffModal` seeds from the resolved `datacallID` so deep links still pin the modal to the URL's call. Exact-count tests cover both the cold and dashboard flows.

   One discovery worth flagging: the component-test harness was plain `<MemoryRouter>`, whose `useNavigate` is keyed on `locationPathname` and changes identity on the canonical redirect — re-running the effect in tests in ways the real app (a `createHashRouter` data router with the stable navigate) never does. The harness now uses `createMemoryRouter` + `RouterProvider`, so the exact-count assertions reflect production behavior.

2. **Decommissioned systems** — when the acronym misses the active list, the page now lazily fetches `fismasystems?decommissioned=true` once (page-local, so the app-wide list is untouched) and retries resolution against it, landing on the existing "No questionnaire is available" path. The not-found warning only renders after both lists have answered.

3. **Case-insensitive datacall matching** — done, consistent with the other resolvers.

4. **Underscore collision** — took this one too: literal underscores are doubled before spaces become underscores, so `FY_2025 Q4` and `FY 2025_Q4` encode distinctly. Matching re-encodes each candidate name rather than decoding the slug, and current names contain no underscores, so existing URLs are byte-identical.

Full suite green locally: `tsc --noEmit`, eslint (0 errors), jest 402 passed — the double-fetch regression test was verified to fail against e99dab2 before the fix.

jsos3-cms:
Verified all four fixes and moved them onto this branch (re-signed onto current main, authorship preserved). The double-run is gone — an exact-count test now asserts a single fetch per cold deep-link (and for the dashboard flow); the decommissioned-system path lands on the no-questionnaire state; datacall matching is case-insensitive; and the underscore encoding round-trips. Full suite green (402), with the regression test confirmed failing against the pre-fix commit. CI is re-running on the updated head, and since the commits are signed this can merge without an admin override once approved and the dev deploy passes. Thanks @voidspooks.

jsos3-cms (approved):

a-dipietro (approved):
Reviewed and approved per slack convo

PR #527, opened 2026-07-14
Fix/issue 23

costacalvin:
> [!WARNING]
> **FOR ZTMF-INSIGHTS REPO ONLY — DO NOT APPROVE OR MERGE**

## Summary

Implements the frontend proposed in `CMS-Enterprise/ztmf-insights#23` to clearly separate:

- ZTMF Insights’ suggested justification
- The ISSO’s response from last year
- The ISSO’s current response

Only the text in **Current response** is submitted.

## Changes

- Adds separate, source-labeled cards for the Insights suggestion and last year’s ISSO response.
- Requires the ISSO to explicitly insert or dismiss carried-forward text before continuing.
- Keeps Next/Complete disabled while prior-response review or the initial Insights lookup is pending.
- Persists an explicit review even when the final response is identical to last year’s text.
- Allows both cards to be dismissed, reopened, and inserted again if their content is removed.
- Uses two visible states: **Added to response** and **Not used**.
- Keeps edited inserted content marked as Added until it is completely removed.
- Disables insert actions when the content is already present or would exceed the 2,000-character limit, with an explanation.
- Keeps the response area a fixed size and makes contextual content scrollable.
- Only shows **Show all** when contextual text is clipped.
- Moves smaller, rounded actions into the card heading row.
- Adds source-specific ARIA labels and tooltips.
- Preserves the original four-row textarea for questions without Insights data.
- Displays an error and disables Next/Complete when an answer changes without a substantive justification update.
- Scopes review state by system, data call, and question to prevent state leaking between questions.

## HHS behavior

For HHS data calls:

- The main Insights panel is hidden.
- Option-level Insights badges are hidden.
- The suggested-justification card is hidden.
- Last year’s ISSO response remains available when provided by the Insights API.

Prior-response labels use `Last year's response — <data call>` or `Last year's ISSO response` when the data-call name is unavailable.

## Acceptance criteria

- [x] Authorship and submitted content are clear at a glance.
- [x] Carried-forward text cannot be submitted without affirmative review.
- [x] Insights suggestions are advisory and easy to insert or dismiss.
- [x] Dismissed or previously inserted content can be reviewed again.
- [x] Clutter is reduced without removing supporting context.
- [x] The existing dimensions and 2,000-character limit are preserved.
- [x] Questions without Insights retain their previous behavior.
- [x] HHS Insights content is conditionally hidden.

## Testing

Coverage includes:

- Insert, dismiss, reopen, reinsert, edit, and remove behavior
- Required prior-response review and identical-text persistence
- Delayed Insights loading
- HHS/CMS conditional rendering
- Duplicate and over-limit insert errors
- Fixed-height and overflow behavior
- Accessibility labels and read-only rendering
- No-Insights textarea fallback
- Review-state reset when switching systems

Validation results:

- Full Jest suite: **406 passed, 3 skipped**
- Focused justification tests: **29 passed**
- ESLint and Prettier: passed

## Notes

HHS prior-response context depends on the Insights API returning `payload.last_score_notes`; otherwise the questionnaire falls back to the original textarea.

### DIFF
--- a/src/views/QuestionnairePage/JustificationField/JustificationField.test.tsx
+++ b/src/views/QuestionnairePage/JustificationField/JustificationField.test.tsx
@@ -25,12 +25,17 @@ const insight: InsightPayload = {
   },
 }
 
-function Harness() {
+function Harness({
+  contextId = 'system-1003:question-1',
+}: {
+  contextId?: string
+}) {
   const initial = insight.last_score_notes ?? ''
   const [value, setValue] = React.useState(initial)
   const [review, setReview] = React.useState<PriorReviewState>('pending')
   return (
     <JustificationField
+      contextId={contextId}
       label="Explain the available authentication options"
       value={value}
       onChange={setValue}
@@ -61,6 +66,19 @@ describe('justification context helpers', () => {
 
   it('does not label the viewed data call as a previous response', () => {
     expect(priorResponseFor(insight, 'FY2025_Q1')).toBeUndefined()
+  })
+
+  it("labels prior content as last year's response", () => {
+    expect(priorResponseFor(insight, 'FY2025 Q2')).toEqual({
+      label: "Last year's response — FY2025 Q1",
+      text: insight.last_score_notes,
+    })
+    expect(
+      priorResponseFor({ last_score_notes: insight.last_score_notes })
+    ).toEqual({
+      label: "Last year's ISSO response",
+      text: insight.last_score_notes,
+    })
   })
 })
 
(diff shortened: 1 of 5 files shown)


jsos3-cms:
Rehomed this to an in-repo branch as #529 (based on `impl`, not `main`) so Analysis CI can run and it can deploy to the impl environment for testing — fork PRs can't reach the org secrets those workflows need. Your two commits are replayed as-is, authorship preserved.

Targeting `impl` is deliberate: this is going to the impl environment first, and the impl↔main divergence will be reconciled separately. Let's continue on #529.

jsos3-cms:
Merged — this shipped via the impl rehome #529, squashed onto `impl` as 03769af, carrying your two commits (authorship preserved). Because it landed on `impl` rather than `main`, it deploys to the impl environment; the impl↔main divergence will be reconciled separately.

Thanks, @costacalvin — closing this in favor of the merged rehome.

PR #528, opened 2026-07-14
Create Datacall ux fixes

tarratsco:
## Summary

Closes #504.

Two UX bugs in the Create Datacall flow:

1. **Silent disabled Create button.** The Create button was disabled with no explanation whenever the datacall name was less than 7 chars or the deadline was empty/incomplete. Users had no signal for what was missing.
2. **New data call not visible until manual page reload.** `submitDatacall` fired a success toast but did not close the modal or refresh the parent's data-call list, so the newly created call was absent from the picker until the user hit reload.

Bonus fixes in the same pass (surfaced during review):

* **Double-submit created duplicate datacalls.** Nothing disabled the Create button during the in-flight POST — a fast double-click or double-Enter fired two POSTs before the modal closed.
* **Invalid calendar dates passed validation.** `Date.parse("02/30/2026")` silently rolls over to Mar 2. The button enabled and the wrong date submitted.
* **`SingleInputDateField` was uncontrolled.** Rendered with hardcoded `value=""` instead of `value={deadline}`.

## Changes

* **`src/views/DatacallModal/DataCallModal.tsx`**
  * `isValidFormat` no longer waits for a 7-char minimum. Any non-empty invalid input reports "Invalid datacall format" immediately.
  * New `handleDatacallBlur` fires "Datacall name is required" when the user tabs out with an empty field. Standard form UX — untouched fields stay quiet on mount, only complain on blur.
  * `validateDeadline` (blur) now fires "Deadline is required" when the field is empty on blur; the existing "Invalid Deadline" branch stays for non-empty invalid.
  * New `validateDeadlineValue` (on-change) mirrors the blur validation once the input reaches 10 chars — partial input stays quiet.
  * New `isValidCalendarDate` helper: verifies MM/DD/YYYY round-trips against a rebuilt `Date` object, so 0d. Used in both the on-change and on-blur validators. Fixes a pre-existing bug in the blur path.
  * Fix `value=""` to `value={eField}` in the on-change and on-blur validators. Fixes a pre-existing bug in the path.
  * New `submitting` state guards against duplicate POSTs from double-Enter. Button label in-flight for visible feedback. Reset in the open/close effect.
  * `submitDatacall` on success calls `onClose()` so the parent refreshes and the modal auto-closes.
  * Drop the now-redundant `validateDeadlineValue` (subsumed by `length < 10`).
* **`src/types.ts`** — add optional `onCreated?: () => void` to `datacallModalProps`. Fired by the parent so it can re-fetch its data-call list.
[...]

PR #530, opened 2026-07-14
fix(datacall-modal): create-datacall UX fixes (#504)

### LINKED ISSUE #504: Create-datacall modal: no feedback when Create is disabled, and new data call needs a manual refresh to appear
### Summary

Two problems in the create-datacall flow (`src/views/DatacallModal/DataCallModal.tsx`), hit while creating a new data call:

1. **The Create button is disabled with no indication why.**
2. **A successfully created data call does not appear until the page is manually refreshed.**

### Problem 1 — silent disabled Create

The Create button is disabled unless the name matches `^FY(\d{2}|\d{4}) (Q[1-4]|ZTM)$` **and** the deadline is a full 10-char `MM/DD/YYYY`. But the inline name error is only set once the input reaches `DATACALL_MIN_LENGTH` (7 chars):

```
function isValidFormat(input: string) {
  if (input.length < DATACALL_MIN_LENGTH) { setDatacallError(''); return }
  ...
}
```

So typing a short/partial name (e.g. `2026`, `FY2026`) clears the error yet leaves Create disabled — the user gets no message explaining what's wrong. The deadline requirement is likewise invisible: `deadlineError` is only set on blur, so while the deadline is incomplete the button is disabled silently.

**Expected:** when Create is disabled, tell the user why — show the format/required-field guidance inline (name pattern `FY<YYYY> Q<1-4>` or `FY<YY> ZTM`, deadline required), rather than a silently-disabled button.

Related minor: the date field is rendered with a hardcoded `value=""` (line ~177) instead of `value={deadline}`, so the field is not a properly controlled input — worth correcting in the same pass.

### Problem 2 — new data call not shown until manual refresh

On success, `submitDatacall` only fires a success toast:

```
await axiosInstance.post(`/datacalls`, {...})
notify('Datacall has successfully been created', 'success', {...})
```

It does not close the modal or re-fetch the data-call list. `Title` fetches `/datacalls` once on mount (`fetchDatacalls` in a `useEffect`), so the newly created call does not appear in the selector/picker until the user manually reloads the page.

**Expected:** after a successful create, close the modal and refresh the data-call list (re-fetch or optimistically add) so the new call shows immediately.

### Files
- `src/views/DatacallModal/DataCallModal.tsx` (validation feedback, date `value`, post-create refresh)
- `src/views/Title/Title.tsx` (owns the data-call list / `fetchDatacalls`)

jsos3-cms:
Rehome of #528 (by @tarratsco) into an in-repo branch so CI and the dev deployment can run — fork PRs can't reach the org secrets those workflows need. tarratsco's two commits are replayed onto current main (authorship preserved); one follow-up fix is added on top (below).

Closes #504.

## Summary

UX fixes for the Create Data Call modal: real calendar-date validation (rejects roll-over dates like `02/30`), a double-submit guard, required-on-blur messaging, and an `onCreated` callback so a newly created call appears in the picker without a manual reload. Also makes the deadline field properly controlled (`value={deadline}` — it was previously frozen to `""`).

## Follow-up fix in this branch

The post-create refetch reused the mount-time `fetchDatacalls`, which resets the active year/selection to the latest year group. Creating a datacall while viewing an older year would snap the selection back to the latest — and since that selection flows through the shared Outlet context, it discarded the user's filter across the dashboard and questionnaire. Gated the reset behind a `resetSelection` flag (default true for the initial mount load; false for the post-create refetch).

## Known follow-ups (not blocking)

- Deadline validation runs on the raw `updatedValue` rather than the design system's zero-padded `formattedValue`, so a single-digit month (e.g. `7/14/2026`) never reaches 10 characters — Create stays disabled while typing, then shows "Invalid Deadline" on blur for a calendar-valid date. Pre-existing pattern; worth a follow-up.
- A code comment references a "Required indicator" (`requirementLabel`) that neither field actually passes.

## Test plan

- [ ] `yarn typecheck`, `eslint`, `jest`
- [ ] Create a datacall while viewing an older fiscal year → selection is preserved (not snapped to the latest year)
- [ ] Create a datacall → the new call appears in the picker without a reload
- [ ] Double-click Create → a single POST fires
- [ ] Invalid calendar date (`02/30`) → "Invalid Deadline"; valid `MM/DD/YYYY` → accepted

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -216,6 +216,9 @@ export type editSystemModalProps = {
 export type datacallModalProps = {
   open: boolean
   onClose: () => void
+  // Fired after a successful POST /datacalls so the parent can re-fetch
+  // its data-call list and the new call appears without a manual reload.
+  onCreated?: () => void
 }
 
 export type ScoreData = {
--- a/src/views/DatacallModal/DataCallModal.tsx
+++ b/src/views/DatacallModal/DataCallModal.tsx
@@ -27,9 +27,12 @@ import { isAuthHandled, notify } from '@/utils/notify'
 // introduced FY23/FY24/FY25 ZTM datacall names.
 const DATACALL_NAME_PATTERN = /^FY(\d{2}|\d{4}) (Q[1-4]|ZTM)$/
 const DATACALL_MAX_LENGTH = 10 // "FY2025 ZTM" = 10 chars; longest valid form
-const DATACALL_MIN_LENGTH = 7 // "FY23 Q1" / "FY23 ZTM" share the floor
 
-export default function DataCallModal({ open, onClose }: datacallModalProps) {
+export default function DataCallModal({
+  open,
+  onClose,
+  onCreated,
+}: datacallModalProps) {
   const [datacall, setDatacall] = React.useState<string>('')
   const [datacallError, setDatacallError] = React.useState<string>('')
   const [deadline, setDeadline] = React.useState<string>('')
@@ -61,8 +64,11 @@ export default function DataCallModal({ open, onClose }: datacallModalProps) {
       </Typography>
     )
   }
+  // Validates on every change so the user is never left with a disabled
+  // Create button and no explanation. Empty input still resets the error
+  // (an untouched field should not scream at the user).
   function isValidFormat(input: string) {
-    if (input.length < DATACALL_MIN_LENGTH) {
+    if (input.length === 0) {
       setDatacallError('')
       return
     }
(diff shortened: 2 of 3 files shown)


MackOverflow (dismissed):
Approve with minor changes. Nothing here is blocking on correctness given the disabled-button backstop. Before merge I'd like to see the contradictory requirementLabel comment removed/fixed, its flagged already, but at the rate we're going, it probably makes sense to remove now.

tarratsco:
@MackOverflow, I removed the comment and pushed 👍

jsos3-cms:
Pulling it in, two shakes

jsos3-cms:
Updated — pulled tarratsco's latest commit in (removes the contradictory comment). The push cleared the prior approval, so I've re-requested your review, @MackOverflow 🙏

MackOverflow (approved):
lgtm!

Issue #37, opened 2024-09-04
fix: Reset the modal to begin at the first question


ATNoblis:
When opening any question modal, reset to always be at the first question. This allows users to know that they'll always start at the beginning of questions.

PR #533, opened 2026-07-14
Show Kion pass/fail checks as an ARS-style block in the insights panel

danielbowne:
## What

The ZTMF Insights panel previously rendered only *failing* Kion checks, so a system that passed everything showed nothing — reading as missing data. This adds the passing side and renders Kion as an ARS-style pass/fail block.

This frontend change is the visible tip of a larger pipeline effort. The panel *couldn't* show pass/fail because the enrichment payload didn't carry the passing side, and the failing side's descriptions were unusable. Both were fixed upstream first; this PR consumes the result.

**Scope:** ~230 lines of panel code here, on top of upstream work spanning the Kion sync lambda (new per-check API call + two metadata columns), the Snowflake enrichment view (two versioned revisions), the findings dictionary (re-seed + weekly-task fix + seed-script fix), the CORE snapshot table (schema-safe rebuild + re-grant), and the local dev sync — plus a source-by-source audit to establish which feeds can even express pass/fail. The three files in this diff are the last mile.

## Frontend (this PR)

- Consumes the new passing-check array `kion_passing` (`InsightFinding` shape + a `level`).
- Header mirrors the ARS Controls line — **"N of M checks passed"** — one chip per check: green ✓ passing, red ✗ failing, labeled by the NIST control, with the check slug + description in a hover.
- **Kion always renders the block** so every question looks consistent. SecurityHub and Hardenize keep the detailed finding rows (their sources are failing-only — see below).
- Source chips now show the maturity **word** (e.g. "Initial") instead of a bare number, which read as a finding count. ARS leads the chip row and the drawer.
- Files: `types.ts`, `InsightsPanel.tsx`, `InsightsPanel.test.tsx`.

## Enabling data-pipeline work (upstream, separate repos)

The panel change required a chain of upstream fixes across the Snowflake enrichment view, the findings dictionary, and the Kion sync lambda.

### 1. Kion finding descriptions — control-statement wall → real per-check text

Kion findings rendered a wall of raw NIST control-statement text (e.g. the full AC-2 a–l enumeration, ~1,600 chars), identical across every check mapped to the same control. Root cause: the Kion sync lambda's model only parsed the check name + controls and dropped the API's per-check description, so the dictionary seed fell back to the control statement. Short controls looked fine; long ones (AC-2) dumped a wall.

- **kion-sync lambda (PR #37):** added a per-check detail call (`GET /compliance/check/{id}`); captured `CHECK_DESCRIPTION` + `CHECK_BODY` (the Cloud Custodian policy) into the check-metadata table — additive, existing columns untouched.
- **Findings dictionary:** re-seeded Kion `DESCRIPTION` from the real per-check text; amended the weekly refresh task to mirror `CHECK_DESCRIPTION` (self-healing), and fixed the one-time seed so a full rebuild can't reintroduce the wall.
- Result: each Kion finding shows its own concise description (e.g. *"Identify IAM users without an MFA device enabled"*) + NIST chip — avg ~60 chars vs ~650, max 116 vs 1,665.

### 2. New `kion_passing` payload field (enrichment view v13)

- Added a `KION_PASSING` array column to the insights view: one object per *passing* Kion check — `{ id, nist_controls, description, level }` — exploded from the per-level passing lists and joined to the findings dictionary for the control + real description.
- **Passing-only by design:** the failing side already ships in `findings.kion`; a unified pass+fail list would duplicate it and risk drift. The two halves are paired in the UI.
- Additive contract: the new key flows into the JSONB payload with no change to any existing field, so nothing downstream breaks.

### 3. Pipeline plumbing

- **CORE snapshot rebuilt** to match the widened view. The daily snapshot task does a positional `INSERT ... SELECT *`, so adding a view column requires rebuilding the snapshot table + re-granting the service role — verified the daily task still runs clean (11,856 rows, no drift).
- **Local dev sync** updated so the new array loads as nested JSON, not a JSON string.
- New view validated in a scratch copy before promotion; `kion_passing` confirmed populated end-to-end (view → CORE → local).

### 4. Source pass/fail audit — why only Kion (and ARS) get the block

| source | pass/fail model | result |
|--------|-----------------|--------|
| **Kion** | evaluates every mapped check, reports pass *and* fail | ✅ `kion_passing` shipped |
| **ARS** | already ships satisfied / not-satisfied controls (the "N of M satisfied" block this mirrors) | ✅ already live |
| **SecurityHub / Hardenize** | failing-only at the source — feeds record findings, never passing controls (0 of ~1,900 SecurityHub rows carry any passing set) | ❌ a `*_passing` array would be permanently empty; intentionally not built |
| **CFACTS** | narrative + score, no enumerable checks | ❌ n/a |

## Screenshot

`Kion: 3 of 4 checks passed` — `[✓ IA-5] [✓ AC-2] [✓ AC-2] [✗ IA-2]`, mirroring the ARS Controls block directly above it.

## Testing

- `InsightsPanel.test.tsx`: 41 tests pass, including 10 new for the pass/fail block (header + counts, repeated NIST tags as distinct checks, pass/fail in the accessible name, pluralization, the empty/absent-array gates, SecurityHub block-vs-FindingRow fallback, and opaque-payload degradation).
- `tsc --noEmit` clean; eslint clean (one pre-existing `react-refresh` warning unrelated to this change).

## Follow-up

- **SecurityHub / Hardenize will *not* get passing arrays** — corrected from earlier assumption. Their feeds are failing-only at the source (a control/remark only appears when it fails; 0 of ~1,900 SecurityHub rows carry any passing set), so a `*_passing` array would be permanently empty. They keep detailed finding rows for failures; their positive/clean state is the generic "evaluated, no failing findings" line, not passing chips. There is nothing to "ship" for them.
- **`kion_passing` uniformity:** the view emits the array wherever Kion has passing checks and `NULL` where it doesn't (all checks failed, or no Kion data for that question). If the panel needs a guaranteed-present array to render the block uniformly, the view can `COALESCE` `NULL` → `[]` — a one-line, additive change on request.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -399,6 +399,9 @@ export type InsightFinding = {
   remediation?: string
   severity?: string
   nist_controls?: string
+  // Maturity tier (1-4) of the check. Present on passing-check entries
+  // (`{source}_passing`); the tier that failed/passed is what drives the score.
+  level?: number | null
   instances?: InsightHardenizeInstance[]
 }
 
@@ -459,6 +462,13 @@ export type InsightPayload = {
   ars_failing_controls?: string[] | null
 
   findings?: InsightFindings
+  // Passing checks per source (counterpart to the failing-only `findings`).
+  // Same InsightFinding shape (id, nist_controls, description, level); severity
+  // omitted (a pass isn't a severity event). Kion is live; sechub/hardenize land
+  // later in the identical shape.
+  kion_passing?: InsightFinding[]
+  sechub_passing?: InsightFinding[]
+  hardenize_passing?: InsightFinding[]
 
   // Additive: the pipeline may add keys at any time.
   [key: string]: unknown
--- a/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
+++ b/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
@@ -55,6 +55,7 @@ describe('InsightsPanel', () => {
   it('renders the ZTMF Insights label and all five source chips', () => {
     render(<InsightsPanel payload={fullPayload} />)
     expect(screen.getByText('ZTMF Insights')).toBeInTheDocument()
+    // Chips render the source name with a trailing colon (e.g. "Kion:").
     for (const source of [
       'Kion',
       'SecurityHub',
@@ -62,7 +63,7 @@ describe('InsightsPanel', () => {
       'CFACTS',
       'ARS',
     ]) {
-      expect(screen.getByText(source)).toBeInTheDocument()
+      expect(screen.getByText(`${source}:`)).toBeInTheDocument()
     }
   })
 
@@ -78,8 +79,9 @@ describe('InsightsPanel', () => {
 
   it('marks a source with no data using a dash instead of a score', () => {
     render(<InsightsPanel payload={fullPayload} />)
-    // Hardenize has has_hardenize_data:false, so its dot renders an en-dash.
-    expect(screen.getByText('–')).toBeInTheDocument()
+    // Hardenize has has_hardenize_data:false, so its badge renders an em-dash
+    // instead of a maturity word.
+    expect(screen.getByText('—')).toBeInTheDocument()
   })
 
   it('shows "No suggestion" when suggested_score is null', () => {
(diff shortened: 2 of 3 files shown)


MackOverflow (dismissed):
Looks good, one nit: 

1. (Medium — a11y) aria-label on a role-less <span> may not be announced. In ControlChip, the label is placed on a Box component="span" with no role and no tabIndex:

<Box component="span" aria-label={ariaLabel || undefined} ...>
Per ARIA, aria-label on a generic/role-less element is not guaranteed to be exposed — NVDA/VoiceOver commonly ignore it. Since the pass/fail distinction lives only in the ✓/✗ glyph, color, and this aria-label, the stated 508 goal ("fold pass/fail into the accessible name") may not actually reach screen-reader users. The getByLabelText test passes because Testing Library reads the DOM attribute directly, not real AT behavior — so the test gives false confidence here. Fix: add role="img" to the chip span (makes aria-label its accessible name), or render visually-hidden text (<Box sx={visuallyHidden}>Passed</Box>).

MackOverflow (approved):
lgtm

PR #534, opened 2026-07-14
ci: nightly main→impl sync PR + manual impl deploy

jsos3-cms:
Adds a nightly automation that opens a PR merging `main` into `impl` for human review, plus a manual impl deploy for optional pre-merge validation. No existing workflows change.

## What

- **`nightly-impl-sync.yml`** — scheduled workflow that, at **true midnight Pacific (= 3am Eastern)**, opens a PR merging `main` into `impl` on a dedicated branch (`automation/sync-main-to-impl`). A human reviews the diff, resolves any conflicts on that branch, and merges — which pushes to `impl` and triggers the existing `orchestration-impl` deploy.
  - DST-aware: GitHub cron is UTC-only and doesn't adjust for DST, so both `08:00` UTC (PST) and `07:00` UTC (PDT) are registered and a guard job runs the sync only at true local midnight.
  - Skips if a prior sync PR is still open (won't clobber in-progress conflict resolution) or if `impl` already matches `main`.
  - Uses the built-in `GITHUB_TOKEN`, so PR-level analysis does not auto-run on the sync PR; analysis + deploy run on merge (push to `impl`).
- **`manual-deploy-impl.yml`** — mirrors `manual-deploy-dev.yml` (`workflow_dispatch` → `analysis` + `ui` with `environment: impl`). Optional pre-merge smoke test of a candidate branch against the impl environment: `gh workflow run "Manual Deploy to IMPL" --ref automation/sync-main-to-impl`.

## Why

Keeps `impl` periodically reconciled with `main` through a reviewable, conflict-safe PR rather than an unattended auto-merge — a human handles the merge, conflicts, and the impl update.

## Notes

- Takes effect after this merges (schedules run from the default branch); opening this PR does not fire the cron.
- The first nightly run will open a sizable sync PR — `impl` is currently ~58 commits behind `main` — likely with conflicts. That's the intended human-review path.

## Test plan

- [ ] After merge: `gh workflow run "Nightly sync main -> impl"` opens a `main → impl` PR on `automation/sync-main-to-impl`
- [ ] Re-running while that PR is open is a no-op (skips)
- [ ] `gh workflow run "Manual Deploy to IMPL" --ref automation/sync-main-to-impl` deploys the candidate to impl

### DIFF
--- a/.github/workflows/manual-deploy-impl.yml
+++ b/.github/workflows/manual-deploy-impl.yml
@@ -1,0 +1,25 @@
+name: Manual Deploy to IMPL
+
+# Deploy a chosen branch to the IMPL environment on demand.
+#   - Primary use: smoke-test the nightly main -> impl sync candidate BEFORE
+#     merging it. Run:
+#       gh workflow run "Manual Deploy to IMPL" --ref automation/sync-main-to-impl
+#   - The deploy AT merge is automatic (orchestration-impl, push to impl). This
+#     workflow is the optional pre-merge preview against the shared impl env.
+on:
+  workflow_dispatch:
+
+jobs:
+  analysis:
+    name: Analysis
+    uses: ./.github/workflows/analysis.yml
+    secrets: inherit
+
+  ui:
+    name: ui
+    needs:
+      - analysis
+    uses: ./.github/workflows/ui.yml
+    with:
+      environment: impl
+    secrets: inherit
--- a/.github/workflows/nightly-impl-sync.yml
+++ b/.github/workflows/nightly-impl-sync.yml
@@ -1,0 +1,115 @@
+name: Nightly sync main -> impl
+
+# Opens a PR merging `main` into `impl` each night so a human can review the
+# diff, resolve any conflicts on the sync branch, and merge it (merging pushes
+# to `impl`, which triggers the impl deploy via orchestration-impl).
+#
+# To smoke-test the candidate in the impl environment BEFORE merging, run the
+# "Manual Deploy to IMPL" workflow with --ref automation/sync-main-to-impl.
+#
+# GitHub's cron is UTC and does NOT adjust for DST. Midnight America/Los_Angeles
+# (== 3am America/New_York) is 08:00 UTC under PST and 07:00 UTC under PDT, so
+# both times are registered and the guard job runs the sync only at true local
+# midnight -- the off-season cron proceeds no further.
+
+on:
+  schedule:
+    - cron: '0 8 * * *' # 00:00 PST / 01:00 PDT
+    - cron: '0 7 * * *' # 00:00 PDT / 23:00 PST
+  workflow_dispatch: {} # allow an on-demand sync
+
+permissions:
+  contents: write # reset + push the sync branch
+  pull-requests: write # open the sync PR
+
+concurrency:
+  group: nightly-impl-sync
+  cancel-in-progress: false
+
+env:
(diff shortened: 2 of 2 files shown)


danielbowne:
love this @jsos3-cms should we do one for backend repo too?

danielbowne (commented):
Like the direction, a reviewable sync PR beats an unattended auto-merge. One thing worth leveling on before it goes in, plus a small safety tweak.

**The nightly merge will phantom-conflict, and the mechanism fights the "main wins" policy**

We squash-merge into both `main` and `impl`, so the branches don't share real ancestry, their merge-base is back at #363 (May 22). Resetting the sync branch to pure `main` and opening it against `impl` makes git 3-way diff across that stale base, so everything already promoted to both sides comes back as a conflict. I ran it:

```
git merge-base origin/main origin/impl  ->  f2a4802 (#363, 2026-05-22)
main ahead 58, impl ahead 16
git merge-tree origin/impl origin/main  ->  42 conflicted files
```

Most are `add/add` (InsightsPanel, aggregateScores, PillarScoresModal, etc.): both branches "added" the same file independently because of the squash history, even though it's the same change already on both.

If our policy is "main always wins, impl gets overwritten unless we cherry-pick" (which I think is where we've landed), then a merge PR that forces someone to hand-resolve 42 conflicts every night, where the answer is always "take main," is the wrong tool. Same call, 42 times, forever, and it never shrinks because merging a no-shared-ancestry branch just spawns another one next night.

Two ways to make it deterministic and zero-touch, both matching that policy:
- `git merge -X theirs origin/main` on the sync branch: main wins every overlap, impl-only files that don't conflict stick around, no hand-resolution.
- Or just reset `impl` to `main` outright if we truly want main to win everything and cherry-pick to be the only survival path.

Either way the PR carries a clean deterministic result instead of a conflict pile, and it has to land as a merge commit (not squash) or the divergence never closes.

**Small safety thing:** `git push --force` (line ~95) resets the branch to main unconditionally. If anyone ever pushes to `automation/sync-main-to-impl` (a cherry-pick, a tweak) between runs, the next run wipes it. `--force-with-lease` plus a quick check for non-main commits is cheap insurance.

Minor: both crons fire at :00, the most delayed scheduler slot, so a >1hr delay can make the guard miss local midnight and skip a night. Manual dispatch covers it.

DST guard logic and the permissions/security scoping both look solid.

jsos3-cms:
Thanks @danielbowne — great catch on the phantom conflicts. Pushed an update:

- Sync branch now builds by merging `main` into an impl-based branch with **`-X theirs`**, so main deterministically wins every overlap and impl-only files that don't conflict survive — no more 42-conflict pile. The PR body now calls out that it must land as a **merge commit, not squash**, so shared ancestry closes the gap instead of re-conflicting nightly.
- Force-push safety: **`--force-with-lease`** plus a guard that skips if the sync-branch tip was hand-edited (non-automation author), so a cherry-pick/tweak won't get wiped.
- Moved the crons off the contended `:00` slot to `:20`.

On a backend version — **yes, definitely want one for `ztmf` too.** Plan is to get this one fully settled here first, then port it over rather than maintaining two in parallel while it's still changing.

danielbowne (approved):
LGTM. The impl-based `-X theirs` merge is exactly right for the main-wins policy, deterministic with nothing to hand-resolve, and the merge-commit note keeps the divergence from coming back. force-with-lease plus the non-automation-author guard covers the branch. Nice work on the modify/delete abort too.

PR #535, opened 2026-07-15
Target maturity isolated save

### LINKED ISSUE #525: Target Maturity Level needs its own save button, decoupled from page-level required-field validation
## Description

On the assessment/edit page, the **Target Maturity Level** field (and its required justification) shares a single page-level save with every other field on the page. Because save is gated on *all* required fields being complete, an ISSO cannot save a change to Target Maturity Level (or its justification) whenever anything else on the page is still blank.

This makes changing the target maturity level unnecessarily coupled to unrelated fields. ISSOs should be able to freely adjust the Target Maturity Level and its justification on its own, without being blocked by incomplete fields elsewhere.

The Target Maturity Level box needs its **own save action**, logically separate from the rest of the page, that validates only that field plus its required justification.

## Steps to Reproduce

1. Open a record/assessment page that contains the Target Maturity Level field.
2. Leave one or more of the *other* required fields on the page blank.
3. Change the Target Maturity Level value and fill in its required justification.
4. Attempt to save.

## Expected Behavior

- The Target Maturity Level field and its required justification can be saved on their own via a dedicated save action.
- Saving that field validates only the target maturity value and its justification — not unrelated required fields elsewhere on the page.
- ISSOs can freely change the target maturity level regardless of the completeness of the rest of the page.

## Actual Behavior

- Save is blocked whenever any required field anywhere on the page is blank.
- A valid, justified change to Target Maturity Level cannot be persisted until every other required field is also complete.
- ISSOs are blocked from freely updating the target maturity level.

## Acceptance Criteria

- [ ] The Target Maturity Level box has its own save button, decoupled from the page-level save.
- [ ] That save persists the Target Maturity Level value and its required justification independently.
- [ ] Validation on that save is scoped to the target maturity value and its justification only.
- [ ] Blank required fields elsewhere on the page do not block saving a Target Maturity Level change.
- [ ] The justification remains required for a Target Maturity Level change (its own validation is preserved).

## Affected Component

Target Maturity Level field / assessment edit form (React SPA)

## Workaround

Complete every other required field on the page before saving a Target Maturity Level change — the behavior this issue asks to remove.

Refs CMS-Enterprise/ztmf#398, CMS-Enterprise/ztmf#405

tarratsco:
## Summary

Closes #525.
Refs CMS-Enterprise/ztmf#398, CMS-Enterprise/ztmf#405

Target maturity now has its own Save button and its own edit flow, independent of the page-level Edit button. An assigned ISSO can update the target level even when other required fields on the system page are blank; those fields are validated by the page-level PUT and were previously blocking every ISSO save.

## Root cause

Target maturity lived inside the page-level Edit/Save flow. Entering Edit snapshotted the whole system into `editedSystem`, and Save PUT the full `/fismasystems/:id` body, which the backend validates for required fields (`issoemail`, `datacallcontact`, etc.). ISSOs typically inherit systems with legacy blanks in those fields, so their target-maturity change couldn't get through: the page's `isFormValid` check disabled Save, and even if it hadn't, the backend would 400 on the missing required fields. The problem wasn't really an ISSO permission gap, it was a validation-scope mismatch - the page save was doing a full-system validate for a two-field change.

## Fix

Extract target maturity into a self-managing card:

- `TargetMaturityCard` owns its own view/edit/save state. Save fires `PUT /fismasystems/:id/target-maturity`, which validates only the two fields it writes. Justification stays required.
- Page-level Edit becomes admin-only. Target maturity was the only field ISSOs could write there, so `SystemDetailHeader.canEdit` is now `isAdmin`.
- The card's Edit button is gated on `!isEditing` so an admin can't run both flows at once. Saving the card mid-page-edit would fire the `[isEditing, system]` useEffect and re-seed `editedSystem`, wiping any in-progress page-form edits.
- Save guards a 200 with no body: the endpoint contract is to echo the updated record, so a missing payload now surfaces an error toast instead of a silent success that diverges from the server.

`SystemDetailPage`'s target-maturity state, useEffect seeding, and `hasUnsavedChanges` / `isFormValid` target checks are all removed. The card writes back into `fismaSystems` via a new `handleTargetMaturitySaved` callback.

## Known tradeoff

Opening the card's Edit, then clicking the page Edit button, causes the card to unmount and drop any typed justification with no confirm. The reverse order is safe (card's Edit is hidden while page is in edit mode). This is a minor edge case left for a follow-up if it surfaces.

## Testing

- 15 tests in `TargetMaturityCard.test.tsx` cover view mode, editing, seed-on-Edit, save success, save with server error, save with an empty 200 body, cancel-clean vs cancel-dirty with ConfirmDialog, tier option filtering, and justification validation (required, over-limit, trim before send).
- `make check` and `make test` clean.
- Manual: verified the ISSO save unblocks; admin dual-flow is gated; card cancel-with-dirty prompts before discarding.

### DIFF
--- a/src/views/SystemDetailPage/SystemDetailEditView.tsx
+++ b/src/views/SystemDetailPage/SystemDetailEditView.tsx
@@ -64,7 +64,7 @@ interface SystemDetailEditViewProps {
   // SystemDetailPage (which reads it from the outlet context).
   datacenterEnvironments: DataCenterEnvironment[]
   // Rendered in the right column between Data Lake Export and Organization,
-  // matching the read view's placement (ztmf#398).
+  // matching the read view's placement.
   targetMaturitySlot?: React.ReactNode
 }
 
--- a/src/views/SystemDetailPage/SystemDetailPage.tsx
+++ b/src/views/SystemDetailPage/SystemDetailPage.tsx
@@ -23,10 +23,6 @@ import SystemDetailHeader from './SystemDetailHeader'
 import SystemDetailReadView from './SystemDetailReadView'
 import SystemDetailEditView from './SystemDetailEditView'
 import TargetMaturityCard from './TargetMaturityCard'
-import {
-  DEFAULT_TARGET_TIER,
-  TARGET_JUSTIFICATION_MAX,
-} from './targetMaturityConfig'
 import { EXTENDED_METADATA_KEYS } from './fieldConfig'
 import CfactsRecordCard from './CfactsRecordCard'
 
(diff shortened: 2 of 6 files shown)


tarratsco:
ISSO editing maturity level:

https://github.com/user-attachments/assets/886adff7-5d80-4419-83da-1b880c8d6bfd

Maturity lvl edit is canceled when the system edit btn is clicked (Admin)


https://github.com/user-attachments/assets/a79600ce-9c67-4f21-954f-e3ad3e53a4a9

stu-aq:
Rehomed to #538 so the Snyk/deploy CI gates run (they cannot on a fork PR). Commits pushed unchanged, authorship preserved. This PR can be closed in favor of #538. Thanks @tarratsco!

tarratsco:
Covered in https://github.com/CMS-Enterprise/ztmf-ui/pull/538

PR #542, opened 2026-07-15
fix(dashboard): Data Call Progress "Complete" chip for past calls, not "0/40 Not updated"

### LINKED ISSUE #537: Dashboard "Not updated / 0-of-40" is a current-cycle metric, wrongly shown for past data calls
## Symptom

On the Home dashboard, selecting a **past** year in the top selector (e.g. `2024 - FY24 ZTM`):
- The `Data Call Progress` column shows `0/40  Not updated` (orange) for systems that were in fact **complete** in that call.
- The `Zero Trust Score` shown doesn't always correspond to the intended call for that year.

**Business impact (per Mackenzie):** people reviewing an old, completed data call see "Not updated / 0/40" and think the data is missing. Retro/historical systems should read as complete when they are complete.

## Root cause (frontend only — backend is correct)

The progress chip is a **this-cycle** metric. `src/views/FismaTable/progressColumn.tsx:37,48,52` derives everything from `entry.questionsupdated` — questions **touched this cycle** — and labels `questionsupdated > 0 ? "Updated" : "Not updated"` with an `N/40` fraction. Per ztmf#299, answers carried over from a prior call don't count as "updated" until re-saved.

For a **past** data call, nobody has touched anything this cycle, so `questionsupdated = 0` for every system → `0/40 Not updated`, regardless of whether the call was completed at the time. "Updated this cycle" has no meaning once a call is closed.

Aggregation context: `src/views/Home/aggregateScores.ts` `buildDashboardMaps` (l.69-89) picks, per system, the call it **most recently updated** (`chosenCallMap`, also feeding the Pillar Scores modal since #520) and pulls score + progress from it. That "best call per system" pick is fine and should stay — the bug is the progress chip's current-cycle framing bleeding onto past calls.

## Selector reality

The selector is **year-level**: a system can have more than one call in a year, so selecting `2025` activates *all* its calls and "best call per system" decides which to show. There is no single-call selection today. From **2026** onward there's one call per year for all OpDivs, so this ambiguity largely disappears — this is mainly a historical / transitional bug, but retro viewing needs to be correct now.

## Fix direction

Make the `0/40` count and the "Not updated" chip apply **only to the current / active data call**. For a past call, show the system's completed state instead of the orange laggard chip.

- Pass through whether the row's displayed call is the current/active one (context already has `latestDataCallId`). If the displayed call is not current, the progress column should not render "Not updated / 0/40".
- The score column keeps showing the best-call score for the selected year.

## Progress display for a past call (decided)

Per @MackOverflow: for a past call, show a **neutral "Complete" chip when the system has a score for that call** (a score means it was completed). Not the orange "Not updated" laggard chip. `ScoreProgress` has no "total answered" field, so the presence of a score is the completion signal.

## Acceptance criteria

- Select a past year (e.g. `FY24 ZTM`): systems with a score for that call show a neutral **Complete** chip, not `0/40 Not updated`.
- The orange "Not updated" / `0/40` chip appears **only** when viewing the current/active data call and the system genuinely hasn't updated it this cycle.
- The Zero Trust Score column continues to reflect the call each system participated in for the selected year.
- 2026's single-call-per-year setup renders correctly (no regression).

## Test coverage to update

- `src/views/FismaTable/progressColumn.test.tsx` — past-call **Complete** chip vs current-call Updated/Not-updated.
- `src/views/FismaTable/dashboardFilters.test.ts` — the "laggard / Not updated" filter is current-call-only.
- `src/views/Home/aggregateScores.test.ts` — best-call selection unchanged; add past-call assertions if the fix touches the maps.

Surfaced on HHS (FDA) historical data but affects any fiscal year with more than one data call.

voidspooks:
Closes #537

## Problem

On the Home dashboard, selecting a **past** year (e.g. `FY24 ZTM`) showed `0/40 Not updated` (orange) for systems that were in fact **complete** in that call — making completed historical data look missing.

## Root cause (frontend only)

The Data Call Progress chip is a **this-cycle** metric: it derives from `entry.questionsupdated`, the count of functions with a genuine edit event this cycle (ztmf#299 — answers carried over from a prior call don't count until re-saved). A past call has no edits this cycle, so `questionsupdated = 0` for those systems → `0/40 Not updated`, regardless of whether the call was completed. "Updated this cycle" has no meaning once a call is closed.

## Fix

Gate the current-cycle framing to the **current/active** data call. A row's displayed call is current when `chosenCallMap[fismasystemid] === latestDataCallId` (both already available — `chosenCallMap` from `buildDashboardMaps`, `latestDataCallId` from context).

- **Past call with a score** → neutral **Complete** chip. Per the issue decision, a score is the completion signal (`ScoreProgress` has no total-answered field).
- **Current call** → unchanged `N/40` fraction + Updated/Not-updated chip.
- The orange "Not updated" chip and the "Not updated only" filter are now **current-call-only**, so a past year surfaces no false laggards.
- The `0/0` **N/A** case and the Zero Trust Score column are unchanged.

Nothing is hardcoded to a specific call: `latestDataCallId` is derived dynamically (furthest-out deadline), so this stays correct for every future cycle, including the 2026+ single-call-per-year setup.

## Changes

- `progressColumn.tsx` — `ProgressCell` takes `isCurrentCall` / `hasScore`; renders the Complete chip for a scored past call (defaults preserve current behavior for existing callers).
- `progressHelpers.ts` — `progressTooltip` gains a `{ completed }` option so its fallback ("Data call complete") doesn't contradict the Complete chip; a real timestamp still wins.
- `dashboardFilters.ts` — `isNotUpdated` / `applyDashboardFilters` take an `isCurrentCall` signal; the not-updated facet only matches current-call laggards.
- `FismaTable.tsx` — derives the per-row `isRowCurrentCall` predicate and wires it into the column and the filter.

## Acceptance criteria

- [x] Past year: systems with a score show a neutral **Complete** chip, not `0/40 Not updated`.
- [x] Orange "Not updated" / `0/40` appears **only** on the current/active call.
- [x] Zero Trust Score column continues to reflect each system's call for the selected year.
- [x] 2026 single-call-per-year renders correctly (no regression).

## Testing

- `progressColumn.test.tsx` — past-call Complete chip vs current-call Updated/Not-updated; past-call no-score em-dash; tooltip completed cases.
- `dashboardFilters.test.ts` — the laggard filter is current-call-only.
- `aggregateScores.test.ts` — untouched; the fix doesn't alter the maps (best-call selection unchanged).
- Local: `tsc --noEmit`, `eslint` (0 errors), `jest` 420 passed. Pre-commit + pre-push hooks green.

> Note: Snyk / secret-dependent CI checks will show red on this fork PR (no repo secrets) — expected; maintainers re-run after pulling.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -268,6 +268,11 @@ export type SystemScoreEntry = {
 export type ScoreProgress = {
   fismasystemid: number
   questionsexpected: number
+  // Distinct applicable questions with an answer at all, any status (ztmf#437).
+  // The completion signal a closed call needs: imported/carried answers are
+  // answered but never "updated this cycle". Optional so the UI degrades to the
+  // prior score-presence proxy until the backend field is deployed.
+  questionsanswered?: number
   questionsupdated: number
   lastupdatedat?: string | null
   updatedsincestart: boolean
--- a/src/views/FismaTable/FismaTable.tsx
+++ b/src/views/FismaTable/FismaTable.tsx
@@ -26,7 +26,7 @@ import {
   Button,
 } from '@mui/material'
 import CloseIcon from '@mui/icons-material/Close'
-import { useState, useEffect, useMemo } from 'react'
+import { useState, useEffect, useMemo, useCallback } from 'react'
 import FormControlLabel from '@mui/material/FormControlLabel'
 import Switch from '@mui/material/Switch'
 import FileDownloadSharpIcon from '@mui/icons-material/FileDownloadSharp'
@@ -415,6 +415,21 @@ export default function FismaTable({
     datacenterEnvironments,
   } = useContextProp()
   const activeDataCallId = selectedDatacall?.datacallid ?? latestDataCallId
+  // Whether the call a given row is displaying (chosen by most-recently-updated
+  // in buildDashboardMaps) is the current/active one — the call with the
+  // furthest-out deadline (latestDataCallId). The Data Call Progress column's
+  // current-cycle framing (the "0/40 Not updated" laggard chip and the
+  // not-updated filter) only makes sense for that call; a past call shows a
+  // neutral Complete chip instead (ztmf#537). Rows without a chosen call, or
+  // before latestDataCallId has loaded, keep the current-cycle rendering.
+  const isRowCurrentCall = useCallback(
+    (fismasystemid: number): boolean => {
+      const chosen = chosenCallMap[fismasystemid]
+      if (!latestDataCallId || chosen == null) return true
+      return chosen === latestDataCallId
+    },
+    [chosenCallMap, latestDataCallId]
+  )
   const hasSystemDetailAccess = hasSystemAccess(userInfo)
   const [open, setOpen] = useState<boolean>(false)
   const [selectedRow, setSelectedRow] = useState<FismaSystemType | null>(null)
@@ -532,8 +547,14 @@ export default function FismaTable({
 
   const filteredRows = useMemo(
     () =>
-      applyDashboardFilters(fismaSystems, progress ?? {}, categoryMap, filters),
-    [fismaSystems, progress, categoryMap, filters]
+      applyDashboardFilters(
+        fismaSystems,
+        progress ?? {},
+        categoryMap,
+        filters,
+        isRowCurrentCall
+      ),
+    [fismaSystems, progress, categoryMap, filters, isRowCurrentCall]
   )
   const [pillarScoresModal, setPillarScoresModal] = useState<{
     open: boolean
@@ -715,7 +736,11 @@ export default function FismaTable({
       valueGetter: (value) =>
         progressSortValue(progress?.[value.row.fismasystemid]),
       renderCell: (params) => (
-        <ProgressCell entry={progress?.[params.row.fismasystemid]} />
+        <ProgressCell
+          entry={progress?.[params.row.fismasystemid]}
+          isCurrentCall={isRowCurrentCall(params.row.fismasystemid)}
+          hasScore={Boolean(scores[params.row.fismasystemid])}
+        />
       ),
     },
     {
--- a/src/views/FismaTable/FismaTable.tsx
+++ b/src/views/FismaTable/FismaTable.tsx
@@ -415,20 +415,32 @@ export default function FismaTable({
     datacenterEnvironments,
   } = useContextProp()
   const activeDataCallId = selectedDatacall?.datacallid ?? latestDataCallId
+  // "Latest by deadline" is not the same as "still open". Once the newest
+  // call's deadline has passed there is no active cycle at all, so nothing is
+  // "current" - every row must render past-call (Complete/Incomplete) rather
+  // than the "0/40 Not updated" laggard framing. This mirrors the
+  // Current-while-open / Latest-once-closed distinction the call picker already
+  // makes (ztmf#393). Without this gate the newest *closed* call stays labeled
+  // current and shows 0/40 forever (ztmf-ui#542).
+  const latestDeadlinePassed = useMemo(() => {
+    if (!latestDataCallId) return false
+    const latest = datacalls.find((d) => d.datacallid === latestDataCallId)
+    return latest ? new Date() > new Date(latest.deadline) : false
+  }, [datacalls, latestDataCallId])
   // Whether the call a given row is displaying (chosen by most-recently-updated
-  // in buildDashboardMaps) is the current/active one — the call with the
-  // furthest-out deadline (latestDataCallId). The Data Call Progress column's
+  // in buildDashboardMaps) is the current/active one: the latest-by-deadline
+  // call AND that call is still open. The Data Call Progress column's
   // current-cycle framing (the "0/40 Not updated" laggard chip and the
-  // not-updated filter) only makes sense for that call; a past call shows a
-  // neutral Complete chip instead (ztmf#537). Rows without a chosen call, or
-  // before latestDataCallId has loaded, keep the current-cycle rendering.
+  // not-updated filter) only makes sense then; a past/closed call shows a
+  // neutral Complete/Incomplete chip instead (ztmf#537). Rows without a chosen
+  // call, or before latestDataCallId has loaded, keep the current rendering.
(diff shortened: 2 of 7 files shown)


jsos3-cms:
Rehoming this one.

jsos3-cms:
Thanks @voidspooks — the direction is right (a closed call shouldn't read as an overdue "0/40"), but there are a couple of correctness issues to sort before this comes off draft.

### 🔴 Blocker — "Complete" is shown for *partially* scored past calls, not just finished ones

`src/views/FismaTable/progressColumn.tsx:66-76` renders the past-call chip on `hasScore` alone — `Boolean(scores[fismasystemid])`, i.e. "does the system have any `ScoreAggregate` row for the chosen call." But a `ScoreAggregate` row is created as soon as a system answers **any** question for a call (backend `scores.go`: `scored_pairs = SELECT DISTINCT fismasystemid, datacallid FROM scores`); unanswered pillars just default to "Not Assessed." So a score row does **not** mean the call was completed.

Failure case: System X was 10/40 in FY26 Q1 (now closed) and never touched FY26 Q2 (current). `buildDashboardMaps` picks Q1 as X's chosen call, so `isCurrentCall = false` and `hasScore = true` → the cell renders **"Complete"**, indistinguishable from a system that actually finished 40/40. That masks genuinely incomplete historical work — the inverse of the bug this PR fixes.

The completion signal is already in scope: the `entry` (`ScoreProgress`) carries `questionsupdated`/`questionsexpected`. Gate on actual completion — e.g. `progressSortValue(entry) === 1` (or `questionsupdated >= questionsexpected`) — and render something honest (a fraction, or a distinct "Incomplete") when a past-call system wasn't finished.

### 🟠 Major — the column sort key contradicts the chip

`src/views/FismaTable/FismaTable.tsx:736-737` sorts by `progressSortValue(progress?.[id])` unconditionally — it never consults `isRowCurrentCall`. `progressSortValue` returns `-1` (top of the "most urgent" order) whenever `questionsupdated <= 0`. So a past-call row rendered as "Complete" jumps to the very top when you sort the column, ahead of real current-call laggards — contradicting its own chip. `dashboardFilters.ts` (`isNotUpdated`) was correctly updated for the current/past distinction; the sort key needs the same treatment.

### 🟡 Test gap

`progressColumn.test.tsx`'s past-call tests only use the `0/41` entry. There's no test with a *partial* entry (e.g. `10/41`) + `isCurrentCall=false` + `hasScore=true` — the exact case behind the blocker. Adding one would have caught it.

### 🟡 Worth confirming (not necessarily this PR)

"Current" here is `latestDataCallId` — the single furthest-out-deadline call — not "hasn't closed yet"; it's never compared against today. If two calls in the active year group can be open simultaneously (e.g. Q3 still open while Q4 is already scheduled with a later deadline), Q3 would be treated as "past" and get the Complete-masking branch above. If that can happen in practice, "current" may need an actual deadline-vs-now check rather than an identity match on one precomputed id.

Happy to re-review once the completion gating + sort key are addressed.

danielbowne:
Two things, one for this PR and one that's on us on the data side, so they don't get conflated.

**Needs a change here:** the Complete chip works for calls older than the latest, but not for the newest closed call. `isRowCurrentCall` treats a row as current when its call === `latestDataCallId`, and `latestDataCallId` is just latest-by-deadline with no check that the deadline is still in the future. Right now no data call is open (newest deadline is 2025-09-30), so the newest call is closed but still gets classified as current and renders `0/40 Not updated` instead of Complete. It's the same distinction ScoreDiffModal already makes per #393 ("Current while open, Latest once the deadline has passed"). Adding a `deadline > now` gate to `isRowCurrentCall` makes the newest closed call render Complete like the older ones. Verified locally with this branch running: FY25 still shows 0/40 until that gate is in. That gate fully covers the frontend side of this.

**Not a frontend concern, just a heads-up:** imported historical rows show a blank last-updated because they were loaded outside the app and never got an edit event. We'll reseed the HHS dev history to add import attribution (a service account plus backdated import events) so those rows show who and when they were loaded instead of blank. This is provenance only and does not affect this PR's logic - a closed call renders Complete from score presence regardless of the per-answer status, so nothing about the seed changes what the chip shows. Flagging it only so a blank last-updated on old HHS rows doesn't read as a bug here.

danielbowne:
Reworked to address the review (we're driving this one now to get it landed alongside the backend):

**🔴 Blocker (partial past calls shown Complete)** — fixed. Completion now gates on `questionsanswered` (the new field from ztmf#437), not `hasScore`. A fully-answered past call reads **Complete**; a partially-answered one shows an honest `answered/total` + **Incomplete** chip. Your 10/41 case now renders `10/41 Incomplete`, not Complete. Until the backend field is deployed the cell falls back to the prior score-presence proxy, so this is safe to merge ahead of the backend but only fully correct once ztmf#437 is live.

**🟠 Sort key contradicts chip** — fixed. `progressSortValue` now takes `isCurrentCall`; past-call rows sort by completion (answered/total), never the current-cycle `-1` laggard rank, so they no longer jump above real current-call laggards.

**🟡 Test gap** — added: partial-past → Incomplete, fully-answered → Complete, and a past-call sort test.

**🟡 Your "worth confirming" (current = latestDataCallId, never vs today)** — done. `isRowCurrentCall` now also requires the latest call to still be open (`deadline > now`). This was the real reason the newest *closed* call stayed at 0/40; with no open call, every row renders past-call. Same Current-while-open / Latest-once-closed distinction the call picker already makes (#393). Covers your two-simultaneous-open-calls variant too.

Full suite green (500 passing), tsc clean. Deploy order: **ztmf#437 first** (ships `questionsanswered`), then this. Happy for your re-review @jsos3-cms.

jsos3-cms:
Rehomed to #567 so the Snyk/Deploy CI gates run — they can't on a fork PR. The commits (including @danielbowne's fixes) were carried over unchanged with authorship preserved, and review is requested there from @MackOverflow and @danielbowne. Keeping this PR open as a draft for now so it can't accidentally merge; it'll be closed once #567 is merged. Thanks @voidspooks and @danielbowne!

jsos3-cms:
#567 merged to main — closing this fork rehome as planned. Thanks @voidspooks and @danielbowne!

PR #543, opened 2026-07-15
fix(nightly-impl-sync): drop brittle midnight guard; open PR via PAT

jsos3-cms:
Fixes the nightly `main` → `impl` sync, which hasn't produced a PR since #534 merged. Two bugs, found while investigating why no sync PR appeared overnight:

## Bugs fixed

**1. Brittle DST guard skipped every night.** The guard required the local hour to be exactly `00`. GitHub delayed the `:20` crons ~40 min, so the guard step executed at 01:00 / 02:00 PT and skipped both scheduled runs (logs: *"Pacific local hour is 01, not midnight — Skipping"*). **Fix:** drop the exact-hour guard. Both DST crons now run; the existing "skip if a sync PR is already open" step dedupes them to one PR/night, robust to scheduler delay.

**2. The org blocks Actions-created PRs.** CMS-Enterprise prohibits GitHub Actions from opening pull requests, so `gh pr create` with `GITHUB_TOKEN` returns `409 Conflict`. A GitHub App would work but installing one on the org repos needs org-owner approval (declined); the admins recommend a PAT. **Fix:** open the PR and push the sync branch with a fine-grained `SYNC_PAT` secret instead.

## ⚠️ Required before this works: provision `SYNC_PAT`

Add a repo secret **`SYNC_PAT`** = a fine-grained personal access token scoped to this repo, permissions **Contents: read/write** + **Pull requests: read/write**. Prefer a shared machine/service account over a personal token, and set a rotation reminder — fine-grained PATs expire (max ~1 year), and when it lapses the "Open PR" step 401s and no PR is opened. Until the secret exists, the workflow runs but does not open a PR.

## Test plan

- [ ] Add the `SYNC_PAT` secret
- [ ] `gh workflow run "Nightly sync main -> impl"` opens a `main → impl` PR on `automation/sync-main-to-impl`
- [ ] Re-run while that PR is open → no-op (skips)
- [ ] Merge the sync PR **as a merge commit** → impl deploy runs (orchestration-impl)

### DIFF
--- a/.github/workflows/nightly-impl-sync.yml
+++ b/.github/workflows/nightly-impl-sync.yml
@@ -7,30 +7,37 @@ name: Nightly sync main -> impl
 # WHY -X theirs (not a plain merge): main and impl are both squash-merged, so
 # they share no real ancestry -- their merge-base is stale (back at #363). A
 # plain 3-way merge across that base surfaces dozens of phantom add/add
-# conflicts for changes already on both sides. We merge `main` into an
-# impl-based branch with `-X theirs`, so main deterministically wins every
-# overlap, impl-only files that don't conflict are preserved, and there is
-# nothing to hand-resolve. The resulting merge commit also re-establishes
-# shared ancestry, so the divergence closes instead of recurring -- which only
-# holds if the PR is merged as a MERGE COMMIT, not squashed.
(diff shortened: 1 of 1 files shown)


MackOverflow (approved):
lgtm! no nits!

PR #544, opened 2026-07-15
use deadline sort as fallback

### LINKED ISSUE #521: PillarScoresModal: fall back to deadline order, not highest datacallid
### Type

Chore

### Description

Follow-up hardening surfaced during the Pillar Scores comparison work (#468 / #520).

In `src/components/PillarScoresModal/PillarScoresModal.tsx`, `latestScore` resolves to the score matching the modal's `selectedDataCallId`, and falls back to the score with the highest `datacallid` when that id is not present in the fetched `scores`:

```ts
const latestScore =
  scores.find((s) => s.datacallid === selectedDataCallId) ??
  scores.reduce((latest, current) =>
    current.datacallid > latest.datacallid ? current : latest
  )
```

Selecting the fallback by highest `datacallid` is the ordering #393 warns against: a re-imported historical data call can carry a higher id than the true current call by deadline. The fallback is low-probability — it only triggers when `selectedDataCallId` is absent from `scores` (for example, a cached-scores vs. chosen-call divergence) — so this is not urgent, but it is inconsistent with the deadline-based ordering the rest of the modal now uses.

**Proposed fix:** resolve the fallback with `sortDatacallsByDeadline` (or the equivalent deadline comparison) instead of the max-`datacallid` reduce, so "current" is chosen consistently across the component.

Refs #468, #520.

costacalvin:
## Summary

Closes #521. Follow-up hardening to the Pillar Scores comparison work (#468 / #520).

`latestScore` in `PillarScoresModal` resolves to the score matching `selectedDataCallId`, with a fallback when that id is absent from the fetched scores. Previously the fallback selected the entry with the highest `datacallid`:

```ts
scores.reduce((latest, current) =>
  current.datacallid > latest.datacallid ? current : latest
)
```

As flagged in #393, a re-imported historical data call can carry a higher `datacallid` than the true current call by deadline, making this ordering unreliable. The fallback was inconsistent with how every other part of the modal (`scoredDatacalls`, `olderScoredDatacalls`, the comparison selector) already determines ordering.

A reviewer also flagged a load-order issue: the new fallback depends on `scoredDatacalls`, which required `datacalls` to be populated by the modal's own async fetch. On the first open (before `/datacalls` resolved), `scoredDatacalls` was `[]` and the fallback produced `null` — showing "No Score Data Available" for systems that have data. A failed fetch would permanently break the fallback for the session.

## Changes

**Fallback ordering fix (`PillarScoresModal.tsx`)**
- Moved the `scoredDatacalls` useMemo above `latestScore` so it's in scope as a fallback source
- Replaced `reduce(max datacallid)` with `scores.find(s => s.datacallid === scoredDatacalls[0]?.datacallid)` — `scoredDatacalls[0]` is the deadline-latest call this system has a score for, using the same `sortDatacallsByDeadline` ordering used throughout the rest of the modal
- No new utilities or files — `sortDatacallsByDeadline` was already imported and `scoredDatacalls` was already computed

**Load-order fix (`PillarScoresModal.tsx`)**
- Replaced the modal's independent `/datacalls` fetch (37-line `useEffect` with `AbortController`, module-level `datacallsCache`, and local `useState`) with `const { datacalls } = useContextProp()`
- `datacalls` is now sourced from the Title outlet context — the same data `FismaTable` already consumes, populated before the user can open the modal
- Removed the `axiosInstance` import — no remaining network calls in this component
- Stale comments referencing "async datacalls fetch" updated to reflect the context-based source

**Tests (`PillarScoresModal.test.tsx` — first test file for this component)**
- 7 tests covering: dialog closed state, title rendering, empty scores, primary path, fallback path (#521 regression), deadline-vs-max-id ordering (#393 regression), and load-order guard (primary path resolves immediately when `datacalls` context is `[]`)
- Added `global.ResizeObserver` stub to `config/jest/setupTests.ts` for Recharts' `ResponsiveContainer`

## Test plan

- [ ] Open the Pillar Scores modal for a system with scores across multiple data calls — confirm the "Current" label and overall score reflect the expected deadline-latest call
- [ ] Confirm the comparison selector defaults to the correct predecessor call
- [ ] Confirm no visible regression in the happy path (modal opens, scores display correctly, trend arrows render)
- [ ] Confirm "No Score Data Available" renders correctly for a system with no scores
- [ ] `npx jest --no-coverage` — all 422 tests pass

### DIFF
--- a/config/jest/setupTests.ts
+++ b/config/jest/setupTests.ts
@@ -3,6 +3,14 @@ import 'jest-fetch-mock'
 import { TextEncoder, TextDecoder } from 'util'
 import { webcrypto } from 'crypto'
 
+// jsdom does not implement ResizeObserver; stub it for components that use it
+// (e.g. Recharts ResponsiveContainer).
+global.ResizeObserver = class {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+} as unknown as typeof ResizeObserver
+
 // Polyfill for React Router 7 which requires TextEncoder/TextDecoder
 global.TextEncoder = TextEncoder
 global.TextDecoder = TextDecoder as typeof global.TextDecoder
--- a/src/components/PillarScoresModal/PillarScoresModal.test.tsx
+++ b/src/components/PillarScoresModal/PillarScoresModal.test.tsx
@@ -146,7 +146,19 @@ describe('PillarScoresModal', () => {
     expect(
       screen.queryByText('No Score Data Available')
     ).not.toBeInTheDocument()
-    // 4.50 appears in the overall score box, pillar score, and data table — all correct
-    expect(screen.getAllByText('4.50').length).toBeGreaterThan(0)
+    // queryAllByText returns [] on no match (unlike getAllByText which throws),
+    // so this gives a legible failure if the wrong score or no score is shown.
+    expect(screen.queryAllByText('4.50')).not.toHaveLength(0)
+  })
+
+  // Load-order: primary path (selectedDataCallId in scores) must resolve
+  // immediately even when datacalls context is still empty.
+  it('shows score data via primary path when datacalls context is empty', () => {
+    mockUseContextProp.mockReturnValue({ datacalls: [] })
+    renderModal({ selectedDataCallId: 5 })
+    expect(
+      screen.queryByText('No Score Data Available')
+    ).not.toBeInTheDocument()
+    expect(screen.getByText('3.75')).toBeInTheDocument()
   })
 })
--- a/src/components/PillarScoresModal/PillarScoresModal.test.tsx
+++ b/src/components/PillarScoresModal/PillarScoresModal.test.tsx
@@ -1,0 +1,152 @@
+import { render, screen } from '@testing-library/react'
+import PillarScoresModal from './PillarScoresModal'
+import type { ScoreAggregate, datacall } from '@/types'
+
+jest.mock('@/views/Title/Context', () => ({
+  useContextProp: jest.fn(),
+}))
+
+const mockUseContextProp = require('@/views/Title/Context')
+  .useContextProp as jest.Mock
+
+// ---------------------------------------------------------------------------
+// Fixtures
+// ---------------------------------------------------------------------------
+
+const DATACALLS: datacall[] = [
+  {
+    datacallid: 5,
+    datacall: 'FY2025 Q4',
+    datecreated: '2025-01-01T00:00:00Z',
+    deadline: '2099-12-31T00:00:00Z',
+  },
+  {
+    datacallid: 4,
+    datacall: 'FY2024 Q4',
+    datecreated: '2024-01-01T00:00:00Z',
+    deadline: '2024-12-31T00:00:00Z',
+  },
+]
+
+const PILLAR_SCORES = [{ pillarid: 1, pillar: 'Identity', score: 3.5 }]
+
+const SCORES: ScoreAggregate[] = [
+  {
+    datacallid: 5,
+    fismasystemid: 1001,
+    systemscore: 3.75,
+    pillarscores: PILLAR_SCORES,
+  },
+  {
+    datacallid: 4,
+    fismasystemid: 1001,
+    systemscore: 3.0,
+    pillarscores: PILLAR_SCORES,
+  },
+]
+
+const DEFAULT_PROPS = {
+  open: true,
+  onClose: jest.fn(),
+  systemName: 'Test System',
+  systemAcronym: 'TS',
+  scores: SCORES,
+  selectedDataCallId: 5,
+}
+
+function renderModal(props: Partial<typeof DEFAULT_PROPS> = {}) {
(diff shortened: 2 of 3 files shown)


jsos3-cms:
Thanks @costacalvin — the direction is right: deadline-order is the correct fallback and fixes the highest-`datacallid` pitfall from #393 (#521). One load-order issue to guard before this comes off draft.

### 🔴 The new fallback can show "No Score Data Available" for a system that has data

The new fallback resolves through `scoredDatacalls`:
```ts
scores.find((s) => s.datacallid === selectedDataCallId)
  ?? scores.find((s) => s.datacallid === scoredDatacalls[0]?.datacallid)
  ?? null
```
`scoredDatacalls` is derived from the modal's **own component-local `datacalls` state**, which starts `[]` and is only populated by the async `useEffect` that fires when the modal opens. The old fallback (`scores.reduce(highest datacallid)`) needed only the already-loaded `scores` prop and was correct immediately; the new one needs `datacalls` to have loaded first.

The fallback path is commonly hit, not a rare race: `chosenCallMap` only has an entry for a system scored in an *active* call, so any system without a score in the current call falls through to `activeDataCallId` in `handleOpenPillarScores`, which often isn't one of that system's scored calls — exactly the case this fallback exists for. On the **first** modal open in a session (before `/datacalls` resolves), `scoredDatacalls` is `[]`, so `scoredDatacalls[0]?.datacallid` is `undefined`, both `.find`s miss, `latestScore` is `null`, and the modal renders the **"No Score Data Available"** empty state for a system that actually has scores. And because the fetch's `catch` only `console.error`s with no retry, a failed `/datacalls` leaves `datacalls` empty for the **rest of the session** — permanently breaking the fallback for affected systems, not just transiently.

Suggested fixes (any one):
- **Reuse the app-wide `datacalls`** already loaded in `Title.tsx`'s Outlet context rather than the modal's independent fetch. This is the documented React Router pattern — a child reads parent-provided data via `useOutletContext` instead of re-fetching it — and `FismaTable` already consumes that context (and imports `sortDatacallsByDeadline`). Since that data is resolved before a user can open the modal, the load-order dependency disappears entirely, and it drops a redundant `/datacalls` fetch. (Cleanest option.)
- Or keep a **`scores`-only interim fallback** (the previous highest-id reduce) while `datacalls.length === 0`, switching to the deadline-based one once `datacalls` has loaded — so the UI never regresses to "no data" for a system that has data.
- Or render a **distinct loading state** while `datacalls` is in flight. React's data-fetching guidance is explicit about separating an "is loading" state from the loaded/empty result (track loading and render `Loading…` until `datacalls` resolves), so a pending fetch never renders as "No Score Data Available."

### 🟡 Worth adding

- A component test for this exact case: `scores` populated, `selectedDataCallId` not in `scores`, `datacalls` empty → assert the modal does **not** show "No Score Data Available." (`PillarScoresModal` has no test file yet, so this would be the first.)
- Minor note (not blocking): `scoredDatacalls` only treats a scored call as a fallback candidate if it also appears in `/datacalls`; an orphaned score row for a call archived from `/datacalls` would be skipped. Edge case, flagging for awareness.

The rest of the change is clean — it reuses the shared `sortDatacallsByDeadline` util (no ad-hoc sort), `tsc`/eslint are green, and the comparison/dropdown logic is untouched. Happy to re-review once the load-order case is guarded.

costacalvin:
Thanks @jsos3-cms  — I implemented **Option 1** (reuse the app-wide `datacalls` from the Title outlet context) as the other options introduced other blocking bugs.

**What changed:**

- Replaced `const [datacalls, setDatacalls] = useState<DataCall[]>([])` with `const { datacalls } = useContextProp()`, which wraps `useOutletContext<ContextType>()` — the same hook `FismaTable` already uses
- Removed the entire 37-line `useEffect` fetch block (AbortController, cache check, axios call, error handling) and the module-level `datacallsCache` object
- Removed the `axiosInstance` import — no remaining network calls in this component

Since `Title.tsx` populates `datacalls` in the outlet context before `FismaTable` renders and before a user can open the modal, the load-order dependency disappears entirely. The transient-empty case (and the permanent-empty case on fetch failure) are no longer possible.

**Tests added** (`PillarScoresModal.test.tsx` — first test file for this component):

1. Dialog not rendered when `open={false}`
2. System name/acronym in title
3. "No Score Data Available" when `scores` is empty
4. Primary path — `selectedDataCallId` matches a score directly
5. Fallback path — `selectedDataCallId` not in `scores` → deadline-latest scored call shown, not "No Score Data Available" *(the exact case you flagged)*
6. #393 regression guard — furthest deadline wins over highest `datacallid`
7. Load-order guard — `datacalls: []` with `selectedDataCallId` in `scores` → primary path resolves immediately without needing `datacalls`

jsos3-cms:
Rehomed to #549 (in-repo) so CI runs against it — that PR closes #521, with your commits replayed as-is (authorship preserved). Nice turnaround on the context fix, @costacalvin — the load-order regression is resolved and the tests cover it. Let's finish up on #549.

jsos3-cms:
#549 merged, closing

Issue #545, opened 2026-07-15
Design: align main→impl sync strategy (merge-commit vs squash) — or move to Environments


jsos3-cms:
## Context

We promote through long-lived environment branches: `main` → `impl` → `prod`, all squash-merged. Because both sides are squash-merged, `main` and `impl` share no real git ancestry (their merge-base is stale), so keeping `impl` reconciled with `main` needs a periodic sync. We've added a nightly automated sync PR in each repo:

- **ztmf-ui:** #543 (fixes the guard/token bugs in the merged #534 workflow)
- **ztmf:** CMS-Enterprise/ztmf#436

Both build the sync branch as `impl` + `git merge -X theirs origin/main` — `main` deterministically wins every overlap, impl-only non-conflicting files survive, nothing to hand-resolve — and open a reviewable PR into `impl`.

## The design choice

The two repos currently **merge the sync differently**, because their `impl` protections differ:

- **ztmf-ui:** `impl` has no ruleset → the sync merges as a **merge commit**, which re-establishes shared ancestry and *closes* the divergence.
- **ztmf:** **squash-only** (`allow_merge_commit: false`; `impl` requires linear history) → the sync **squash-merges**. A scratch-worktree test confirmed `-X theirs` + squash is zero-touch and settles (a no-change night opens no PR; no balloon), but squash records no ancestry, so the gap stays open — content-synced, not ancestry-closed.

Should we bring these into alignment, and toward what?

- **A — Align ztmf-ui *down* to squash.** One uniform model + review instruction across both repos; matches ztmf's policy. Cost: ztmf-ui gives up ancestry-closing (cosmetically larger nightly diffs, but it settles).
- **B — Keep the per-repo difference** (documented). Each repo does the best its policy allows: ztmf-ui closes the gap, ztmf stays content-synced. No churn on the ztmf-ui workflow.
- **C — Align *up* to merge-commit everywhere.** Requires ztmf to enable merge commits — a branch-policy change.
- **D — Retire the `impl`/`prod` branches; promote off `main` via GitHub Environments + required-reviewer deployment protection rules.** Deploys the same commit to successive environments (GitHub's recommended multi-stage promotion pattern) — eliminates the branch divergence *and* the nightly sync entirely. Bigger effort.

## Constraints worth noting

[...]

PR #522, opened 2026-07-14
Rename CfactsRecordCard to SystemEnrichmentCard

MackOverflow:
> **Note:** In-repo copy of #499 by @tarratsco. Moved from the fork so Snyk and the deploy CI gates can run — those jobs require org secrets unavailable to fork PRs. Commits and authorship are unchanged.
>
> Refs #499

---

## Summary

Rename `CfactsRecordCard` → `SystemEnrichmentCard` for naming consistency with the endpoint it already reads from (`/systemenrichment`). No behavior change, no data-source change, and no user-visible impact beyond internal code identifiers.

The card migrated to the new endpoint in `#367` but kept its old file, component, and type names during the migration window while both surfaces coexisted. With the BE having retired the old `/cfactssystems` read surface, the naming can finally follow.

## Changes

| Concern | Before | After |
| --- | --- | --- |
| File | `CfactsRecordCard.tsx` | `SystemEnrichmentCard.tsx` |
| Test file | `CfactsRecordCard.test.tsx` | `SystemEnrichmentCard.test.tsx` |
| Component | `CfactsRecordCard` | `SystemEnrichmentCard` |
| Props interface | `CfactsRecordCardProps` | `SystemEnrichmentCardProps` |
| Type in `src/types.ts` | `CfactsSystemType` | `SystemEnrichmentType` |
| State var in card | `[cfacts, setCfacts]` | `[enrichment, setEnrichment]` |
| Field accesses | `cfacts.<field>` (~20 sites) | `enrichment.<field>` |
| Importer in `SystemDetailPage.tsx` | `CfactsRecordCard` | `SystemEnrichmentCard` |
| Invariants file-path assertion (`authHandling.invariants.test.ts`) | old path | new path |

Files renamed via `git mv` so history is preserved; the file rename shows as `R` in `git status`, not delete + create.

## Intentionally NOT changed

* **`Last modified in CFACTS: ...`** user-visible label at line 286. This refers to the actual upstream CMS source system ("CFACTS" is a real CMS system name), not our internal naming convention. Correct as-is.
* **`cfacts_systems` DB table** — different scope, different repo. The ticket's gate is the read surface (controller/endpoint), not the underlying sync-target table.

### Automated

* `yarn jest SystemEnrichment --no-coverage` — 6/6 passing. The behavioral tests (200/403/404/500 responses, ATO date formatting) are unchanged; they exercise the endpoint and the shape, both of which were untouched.
* `make check` clean.
[...]

PR #548, opened 2026-07-15
Refresh impl from main (60 commits) — catch-up, JustificationField deferred

danielbowne:
Catches impl back up to main after the nightly sync lapsed (impl had fallen 60 commits behind). This is a plain merge of main into impl, not a rebase or cherry-pick.

## Resolution
- All 14 conflicts resolved to main's side — impl carried only older-main version skew (insights panel, data-call persistence, pillar radar, deep links, dashboard filters, etc. all come from main).
- Reconciled the CfactsRecordCard -> SystemEnrichmentCard rename (#522): removed the stale orphan files.

## JustificationField deferred (on purpose)
The one genuinely impl-only feature, JustificationField, is not re-wired here. Its component files stay in the tree (they compile and their standalone test passes), but the QuestionnairePage wiring + integration tests are held for a focused follow-up PR. That wiring hooks into the answer save/load flow, so it deserves its own review and a manual click-through rather than being buried inside a 60-commit sync. After this lands, impl = main plus those dormant files only.

## Verification
tsc --noEmit clean, yarn lint no errors, full jest suite passes (435).

## Merge method
Correction: the repo is squash-only (merge commits are disabled), so the earlier "merge commit" note here was wrong. A plain squash lands the content but leaves the main↔impl ancestry gap open. Rather than pay that on every sync, we're going with a one-time reset (see thread): rebuild `impl` as `main` plus its current impl-only delta, so `main` becomes a real ancestor and the divergence closes, after which merge commits can be enabled for clean future syncs. Holding this PR until the reset sequencing is settled so impl is not deployed twice. The whole-chain fix (impl to prod) is tracked in #545.


MackOverflow:
Merge method — tie this PR to the outcome of #545. As written, the workflow's own reasoning implies #548 needs a merge commit, contradicting its body's "squash-merge" line. Get that reconciled so whoever merges doesn't guess wrong.
File a tracking issue for the JustificationField re-wire + restoring its integration coverage, so the dormant code and the deleted 419-line test don't get forgotten.

Otherwise G2G

jsos3-cms:
Dug into how this branch merges into `impl` and how the repo's merge settings interact with the sync model, since it bears on the merge-method note in the description. **Bottom line: I'm okay approving this PR as-is — a squash-merge (with a trimmed body) is a correct, working outcome. There's also a cleaner one-time improvement worth weighing. Both are right; most of the *other* ways to land this are wrong, so the pitfalls are listed below too. The merge-time call is yours.**

## What I checked, and what each showed

- **Branch ↔ impl ancestry.** The tip is a merge commit (`impl` × `main`), so `impl` is in its ancestry and it fast-forwards cleanly. The PR is currently `MERGEABLE` / `CLEAN`.
- **Merge → deploy.** `orchestration-impl` triggers on push to `impl`, so merging redeploys the impl environment. The deployed tree is current `main` (all 60 catch-up commits) plus the dormant `JustificationField` files — a full catch-up deploy; worth a pre-merge smoke test via **Manual Deploy to IMPL** against the branch.
- **Repo merge settings.** `allow_squash_merge=true`, `allow_merge_commit=false`, `allow_rebase_merge=false` — squash-only.
- **Squash message default.** `COMMIT_MESSAGES`: the squash body concatenates every commit on the branch, so trim it to the PR summary before confirming.

## The pitfalls (the wrong answers)

- **Merge commit** to close ancestry, as the description intends — not available; merge commits are disabled on the repo.
- **Plain merge or rebase onto `main`.** `main` and `impl` are squash-diverged (stale merge-base at #363); a naive 3-way merge, or rebasing this branch onto `main`, drops the merge-commit form and resurfaces all 14 phantom add/add conflicts. Keep the branch in its current merge-commit shape.
- **Squash without trimming the body**, or **rebase-merge** (even if it were enabled — new SHAs, no ancestry gain, and messy across a 60-commit base).

## Two correct answers

**A — Merge as-is (I'm okay approving this).** Squash-merge with the body trimmed to the PR summary. Content lands correctly and impl deploys. The only cost is that squash doesn't close the `main`↔`impl` ancestry gap, so the next sync again pays the `-X theirs` phantom-conflict cost — which the nightly automation already absorbs. Perfectly fine to ship.

**B — One-time reset (cleaner, optional).** Since `impl` is a single environment with no active users: rebuild `impl` as `main` + its current impl-only delta and reset it there. impl then contains main's commits, so `main` becomes an ancestor and the divergence closes for good (the old tip stays recoverable, so it's low-risk). Build from impl's **current tree** on top of main — not by cherry-picking the old commits, which re-add ~700 lines of stale `QuestionnairePage` wiring — yielding `main` + the dormant `JustificationField` files (matching today's impl and this PR's deferral). Afterward impl has a real merge-base with main, so enabling merge commits lets `main → impl` merge normally from then on (clean, no force-push).

Both A and B are correct; B just trades a one-time destructive reset for permanently closing the gap. Your call at merge time, @danielbowne — you own this PR and the deferral choice.

Regardless of path, this only addresses `main ↔ impl`. `impl ↔ prod` is the same divergence and can't be force-reset if prod has users, so the whole-chain fix remains the Environments direction in #545 (promote one build through `dev → impl → prod` as environment targets, retire the long-lived branches).

danielbowne:
Agree, let's go with B. Since impl has no users, taking the one-time reset to actually close the main/impl ancestry gap is worth it over paying the phantom-conflict cost on every sync.

Good catch on the merge settings too. The description was leaning on a merge-commit close, but the repo is squash-only so that option is off the table. I'll fix the description to match.

Plan for B, matching what you laid out: rebuild impl as main plus its current impl-only delta, built from impl's current tree, not by cherry-picking the old commits (that would drag back the ~700 lines of stale QuestionnairePage wiring). That lands main plus the dormant JustificationField files, same as this PR's deferral. Then we can enable merge commits so main to impl merges clean from there on.

Agreed the whole-chain fix (impl to prod too) is the Environments direction in #545, that is the bigger move.

Want to pair on the reset, or are you good to run it? Either way I will hold this PR until we settle the exact sequencing so we don't deploy impl twice.

jsos3-cms:
After discussion with the team; we've decided to go the option B route.

PR #549, opened 2026-07-15
fix(pillar-scores): deadline-order fallback via shared datacalls context (#521)

### LINKED ISSUE #521: PillarScoresModal: fall back to deadline order, not highest datacallid

jsos3-cms:
Rehome of #544 (by @costacalvin) into an in-repo branch so CI runs against it — fork PRs can't reach the org secrets. costacalvin's commits are replayed onto current `main` with authorship preserved.

Closes #521.

## Summary

Makes `PillarScoresModal`'s `latestScore` fallback **deadline-based** instead of highest-`datacallid` (the #393 pitfall), and sources `datacalls` from the app-wide Outlet context (`useContextProp`) that `Title`/`FismaTable` already load — rather than the modal's own async fetch — so the fallback can no longer render "No Score Data Available" for a system with real scores during load. Adds `PillarScoresModal.test.tsx` (the modal's first tests): the #521 fallback-ordering case, the #393 max-id guard, and an empty-context guard.

## Verification

Cherry-picked cleanly onto current `main` (which now carries the #543 workflow fix — unrelated file, no overlap). `yarn typecheck` clean; the modal suite is 7/7 green; reviewed and verified against the load-order regression that the earlier draft had.

## Known follow-up (non-blocking)

One narrow residual: if the deadline fallback is triggered **and** the `datacalls` context is still empty (only in the brief window right after `Title` mounts, before its `/datacalls` fetch resolves), a system with historical scores could still show "No Score Data Available." It's bounded by the same app-boot precondition as the rest of the dashboard (`latestDataCallId`, `activeDatacallIds`, etc.), so it's a follow-up test/guard, not a blocker.

## Test plan

- [ ] `yarn typecheck`, `yarn test` (modal 7/7)
- [ ] Open Pillar Scores for a system whose only scored call isn't the active one → shows the deadline-latest scored call, not "No Score Data Available"

### DIFF
--- a/config/jest/setupTests.ts
+++ b/config/jest/setupTests.ts
@@ -3,6 +3,14 @@ import 'jest-fetch-mock'
 import { TextEncoder, TextDecoder } from 'util'
 import { webcrypto } from 'crypto'
 
+// jsdom does not implement ResizeObserver; stub it for components that use it
+// (e.g. Recharts ResponsiveContainer).
+global.ResizeObserver = class {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+} as unknown as typeof ResizeObserver
+
 // Polyfill for React Router 7 which requires TextEncoder/TextDecoder
 global.TextEncoder = TextEncoder
 global.TextDecoder = TextDecoder as typeof global.TextDecoder
--- a/src/components/PillarScoresModal/PillarScoresModal.test.tsx
+++ b/src/components/PillarScoresModal/PillarScoresModal.test.tsx
@@ -146,7 +146,19 @@ describe('PillarScoresModal', () => {
     expect(
       screen.queryByText('No Score Data Available')
     ).not.toBeInTheDocument()
-    // 4.50 appears in the overall score box, pillar score, and data table — all correct
-    expect(screen.getAllByText('4.50').length).toBeGreaterThan(0)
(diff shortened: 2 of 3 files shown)


danielbowne (approved):
LGTM

PR #550, opened 2026-07-16
Highlight FIPS maturity baseline on questionnaire answer options

### LINKED ISSUE #547: Questionnaire: highlight maturity options above a system's FIPS baseline (warn, still selectable)
Highlight maturity answer options that are **above a system's FIPS baseline** in the questionnaire — a warning cue on options a Low/Moderate system doesn't need to reach, while keeping them selectable (Option 2, decided in ztmf-misc#137).

## Data (already in the insights payload)

Two fields per (system, question) row, on the ZTMF Insights payload the questionnaire already reads:

- `fips_impact_level`: `"Low" | "Moderate" | "High" | null`
- `fips_ceiling`: `1 | 2 | 3 | 4` (Low→2, Moderate→3, High/HVA→4)

## Rule

`option.score > fips_ceiling` → render the option with **above-baseline** warn styling (still selectable). Score 1 is always at/below the ceiling (floor). Example — a Low system (ceiling 2): options 3 (Advanced) and 4 (Optimal) are above baseline; 1 and 2 are at/below.

## Text bindings

- badge: `FIPS {fips_impact_level.toUpperCase()}` (e.g. "FIPS LOW")
- strip: `Baseline for this system is {maturityLabel(fips_ceiling)} (Score {fips_ceiling})`
- divider: `above {fips_impact_level} baseline`
- `maturityLabel`: 1 Traditional, 2 Initial, 3 Advanced, 4 Optimal (existing map)

## Fallback

No FIPS on file → `fips_impact_level = null`, `fips_ceiling = 4` → warns on nothing, never breaks the UI. 247 CMS systems populated today.

## Acceptance criteria

- Options with `score > fips_ceiling` render the above-baseline warn cue and stay selectable.
- The FIPS badge + baseline strip + "above X baseline" divider render per the reference mock (`eppec-auth-v3.html` Scenario E).
- Null/ceiling-4 systems show no warning and render normally.
- **508:** the above-baseline state is conveyed non-visually (text/aria), not color alone.
- Tests cover the rule (`score > ceiling`), the fallback, and the text bindings.

Decision + origin: ztmf-misc#137 (Option 2 — allow selection, warn). Data source auto-upgrades CFACTS→record with no frontend change (same field names).

danielbowne:
Closes #547.

## What

Surfaces each system's FIPS-derived maturity **baseline** in the questionnaire's ZTMF Insights flow. A system's FIPS impact level sets a ceiling — **Low → Initial, Moderate → Advanced, High → Optimal** — and the answer options are marked against it:

- Options **above** the baseline: gold "above baseline" styling, still selectable (optional, not required). Selecting one reveals a divider, an `ABOVE X BASELINE` badge, and a short justification notice.
- The option **at** the ceiling: a dotted "**LOW/MODERATE BASELINE**" box — "where you should be."
- A **FIPS strip** at the top of the insights panel states the baseline in plain language.

Decision origin: ztmf-misc#137 (Option 2 — warn, don't block).

## Data

Two additive fields on the insight payload: `fips_impact_level` (Low/Moderate/High/null) and `fips_ceiling` (1–4). Because they ride on the insight payload, the whole treatment is **gated on insight data** exactly like the rest of the insights UI — no data, nothing renders. A system with no FIPS on file (`fips_ceiling` null → treated as 4) warns on nothing.

## Implementation

- Replaces the CMSDS `ChoiceList` for answer options with a native-radio `<fieldset>` (`QuestionRadioGroup`) so the per-option baseline treatment can render; preserves the existing selection, save, read-only, and option-insight-badge behavior.
- FIPS logic is isolated in `fipsBaseline.ts` (pure, unit-tested); opaque-payload reads are guarded (a malformed level suppresses the strip rather than throwing).

## Accessibility (508)

Native radios in a labeled fieldset; above-baseline state carries a screen-reader cue independent of color; tooltips paired with `aria-label`; an always-mounted `aria-live` region for the notice. This PR also clears the **pre-existing** 508 findings on the questionnaire page: greyed-chip contrast, sidebar list semantics (`ListItem`), `h1`/`h2` heading order, and the notes-textarea label. Axe on the page is clean except two intentionally-left app-shell items (the logo contrast and a header `region`).

## Testing

- tsc + eslint clean; **161** questionnaire tests pass, incl. new coverage for the baseline logic (`fipsBaseline.test.ts`), the radio group (`QuestionRadioGroup.test.tsx`), and the strip (`InsightsPanel.test.tsx`).
- Verified end-to-end against real dev data (a Low-FIPS system) via headless click-through.

Opened as a regular PR (not draft) so it deploys to **dev** for review/approval before merge.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -476,6 +476,15 @@ export type InsightPayload = {
   sechub_passing?: InsightFinding[]
   hardenize_passing?: InsightFinding[]
 
+  // FIPS maturity baseline (ztmf-ui#547). `fips_ceiling` is the highest maturity
+  // tier this system's FIPS impact level needs (Low→2, Moderate→3, High/HVA→4);
+  // an answer option scoring above it is "above baseline" — warned but still
+  // selectable. `fips_impact_level` drives the "FIPS LOW" badge / divider text.
+  // Fallback for a system with no FIPS on file: level null + ceiling 4, so the
+  // rule (score > ceiling) warns on nothing and the UI never breaks.
+  fips_impact_level?: 'Low' | 'Moderate' | 'High' | null
+  fips_ceiling?: number | null
+
   // Additive: the pipeline may add keys at any time.
   [key: string]: unknown
 }
--- a/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
+++ b/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
@@ -717,3 +717,53 @@ describe('FeedCheckBlock (feed pass/fail checks)', () => {
     expect(screen.getByText('ZTMF Insights')).toBeInTheDocument()
   })
 })
+
+describe('FIPS baseline strip', () => {
+  it('shows the FIPS badge + baseline sentence for a system with headroom (Low → Initial)', () => {
+    render(
+      <InsightsPanel
+        payload={{
+          suggested_score: 2,
+          fips_impact_level: 'Low',
+          fips_ceiling: 2,
+        }}
+      />
+    )
+    expect(screen.getByText('FIPS LOW')).toBeInTheDocument()
+    expect(screen.getByText(/Baseline for this system is/)).toBeInTheDocument()
+    expect(screen.getByText('Initial')).toBeInTheDocument()
+  })
+
+  it('hides the strip when there is no headroom (High / ceiling 4)', () => {
+    render(
+      <InsightsPanel
+        payload={{
+          suggested_score: 4,
+          fips_impact_level: 'High',
+          fips_ceiling: 4,
+        }}
+      />
+    )
+    expect(screen.queryByText(/FIPS/)).not.toBeInTheDocument()
+    expect(
+      screen.queryByText(/Baseline for this system is/)
+    ).not.toBeInTheDocument()
+  })
+
+  it('does not throw or render the strip for a malformed (non-string) impact level', () => {
+    render(
+      <InsightsPanel
+        payload={{
+          suggested_score: 2,
+          fips_impact_level: 3 as unknown as 'Low',
+          fips_ceiling: 2,
+        }}
+      />
+    )
+    // Panel still renders; strip is suppressed rather than crashing.
+    expect(screen.getByText('ZTMF Insights')).toBeInTheDocument()
+    expect(
+      screen.queryByText(/Baseline for this system is/)
+    ).not.toBeInTheDocument()
+  })
+})
--- a/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.tsx
+++ b/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.tsx
@@ -10,8 +10,8 @@ import {
   fipsBadgeText,
   showFipsStrip,
   baselineCeiling,
+  asImpactLevel,
   FIPS_GOLD,
-  type FipsImpactLevel,
 } from '../fipsBaseline'
 
 type Props = {
@@ -392,7 +392,7 @@ function InsightsPanelInner({ payload }: Props) {
           panel → no strip). Only shown when there's headroom above the
           baseline (ceiling < 4). */}
       {showFipsStrip(
-        (payload.fips_impact_level ?? null) as FipsImpactLevel | null,
+        asImpactLevel(payload.fips_impact_level),
         payload.fips_ceiling
       ) && (
         <Box
@@ -415,6 +415,7 @@ function InsightsPanelInner({ payload }: Props) {
           >
             <Box
               component="span"
+              tabIndex={0}
               aria-label={`FIPS impact level: ${payload.fips_impact_level}`}
               sx={{
                 fontSize: 11,
@@ -429,9 +430,7 @@ function InsightsPanelInner({ payload }: Props) {
                 cursor: 'help',
               }}
             >
-              {fipsBadgeText(
-                (payload.fips_impact_level ?? null) as FipsImpactLevel | null
-              )}
+              {fipsBadgeText(asImpactLevel(payload.fips_impact_level))}
             </Box>
           </Tooltip>
           <Box
--- a/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.tsx
+++ b/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.tsx
@@ -6,6 +6,13 @@ import Collapse from '@mui/material/Collapse'
 import Link from '@mui/material/Link'
 import type { InsightPayload, InsightFinding } from '@/types'
 import CONFIG from '@/utils/config'
+import {
+  fipsBadgeText,
+  showFipsStrip,
+  baselineCeiling,
+  FIPS_GOLD,
+  type FipsImpactLevel,
+} from '../fipsBaseline'
 
 type Props = {
   payload: InsightPayload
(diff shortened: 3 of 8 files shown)


MackOverflow (dismissed):
solid build out, one semi-major nit: 

1. Gating asymmetry on a malformed impact level (low severity, but the PR claims to defend against it).
showFipsStrip guards with typeof level === 'string', but QuestionRadioGroup computes hasBaseline from level != null:

// QuestionRadioGroup.tsx
const level = (insight?.fips_impact_level ?? null) as FipsImpactLevel | null
const hasBaseline = level != null && anyAbove
If the opaque payload delivers a truthy non-string (e.g. numeric 3), the strip is correctly suppressed but the per-option treatment still renders, producing garbage copy like ▲ above 3 baseline and an aria-label of above the 3 baseline. Since InsightsPanel.test.tsx explicitly tests the malformed-level path for the strip, the radio group deserves the same guard. Route level through a shared asImpactLevel(x): FipsImpactLevel | null helper (returns null unless it's one of the three literals) so both surfaces suppress together — and add the matching test.

jsos3-cms:
Took a close read of the FIPS baseline treatment — nice build-out. The core mapping (Low→Initial, Moderate→Advanced, High→Optimal), the null-ceiling "warn on nothing" default, and the ChoiceList→native-radio migration (selection / save / draft / read-only wiring all preserved) check out. A few items worth addressing before merge, plus some 508 polish, grouped by priority.

## Two opaque-payload fields need runtime validation

Both new insight fields are read straight off the opaque payload and trusted by their declared type; neither is validated at runtime, and each has a distinct failure mode. (@MackOverflow already flagged the first half of this for the level field — folding it in here.)

**`fips_ceiling` — an out-of-range value inverts the fail-safe (I'd fix this one before merge).** `fipsBaseline.ts:46-58`: `baselineCeiling` is `typeof fipsCeiling === 'number' ? fipsCeiling : 4` with no bounds check, and `isAboveBaseline(score, ceiling)` is `score > baselineCeiling(...)`. The `null`/`undefined` path correctly defaults to 4 ("warn on nothing"), but a numeric `0` doesn't take that path: `isAboveBaseline(1, 0)` is `1 > 0` → `true`, so Traditional (score 1) — which the line-53 comment says is "always the floor, never above" — renders as an above-baseline warning, along with every other option. If the backend ever emits `fips_ceiling: 0` for "no FIPS on file" (a zero-value serialized in place of `null`), the treatment shows the exact opposite of its intent. Suggest clamping to the valid range: `typeof x === 'number' && x >= 1 && x <= 4 ? x : 4`, with a test at `0` / `-1` / `NaN`.

**`fips_impact_level` — a non-string renders garbage copy (building on @MackOverflow's note).** The strip guards with `typeof level === 'string'` (`fipsBaseline.ts:41`), but `QuestionRadioGroup.tsx:58` casts `(insight?.fips_impact_level ?? null) as FipsImpactLevel | null` and then gates on `level != null` (`:64`). An `as` cast is compile-time only — no runtime check — so a truthy non-string (e.g. numeric `3`) passes the `!= null` gate and gets interpolated into the option copy and aria-labels (`:140, 199, 210, 228, 240, 258`): "▲ above 3 baseline", aria-label "above the 3 baseline". The malformed-level case is already tested for the strip (`InsightsPanel.test.tsx:758`), so the radio group deserves the same guard. A shared `asImpactLevel(x): FipsImpactLevel | null` (returns null unless `x` is one of the three literals), routed through both surfaces, makes them suppress together — plus a matching radio-group test.

## Accessibility (508)

**Tooltip triggers aren't keyboard-focusable.** `QuestionRadioGroup.tsx:202-231` (baseline chip) and `:232-261` (above-baseline chip), plus the strip/badge tooltips in `InsightsPanel.tsx`, are plain `<span>` triggers with no `tabIndex` inside an MUI `Tooltip`. A keyboard-only sighted user (a motor-impairment case, not just screen-reader users) can't focus them, so the extra explanatory tooltip text — e.g. "worth documenting what's driving it" — is reachable by mouse hover only. axe doesn't flag this, which is why the page scanned clean. Suggest `tabIndex={0}` on the trigger (or wrapping a natively focusable element).

**Duplicated accessible name on the radio options.** `QuestionRadioGroup.tsx:145-267`: several `aria-label`-bearing spans (the SR-only "(above the Low baseline)" cue, the two tooltip chips, and `OptionInsightBadges`) all sit inside the option `<label>`, so they fold into the radio's accessible name and announce the baseline relationship twice, worded slightly differently each time. The `OptionInsightBadges` nesting predates this PR, but it adds another source or two. Suggest `aria-hidden` on the redundant spans (the visible badge text already carries the meaning), or moving them to `aria-describedby`.

## Consistency + tests

**`hasBaseline` couples the per-option marker to the question's own option set.** `QuestionRadioGroup.tsx:59-64, 109`: `hasBaseline = level != null && anyAbove`, so the "where you should be" baseline box only renders when some *other* option in that question also scores above the ceiling. That holds only because every question currently spans scores 1–4; if a question's options don't reach 4, the radio group drops the baseline marker while the strip (`showFipsStrip`, gated on `ceiling < 4`) still shows the baseline for the same system — the two surfaces disagree. Gating on `level != null` alone (matching the strip) removes the hidden assumption.

**Test coverage:** `fipsBaseline.test.ts` / `QuestionRadioGroup.test.tsx` cover `null` and valid 2–4 ceilings, but not a malformed / out-of-range one — the case behind the first item above.

## Minor / notes

- Insight data loads on its own effect (`QuestionnairePage.tsx:340-372`) and can arrive after the radio group has mounted; the `aria-live` region announces the selection notice but not the *appearance* of baseline styling on options a screen-reader user may already have reviewed. Consistent with the gated design — flagging for QA awareness, not a bug.
- `effectiveSelected`'s `defaultChecked` fallback (`QuestionRadioGroup.tsx:69-72`) looks unreachable in practice — every caller that sets `defaultChecked` also sets `selectedValue` in the same batch. Harmless; worth a comment or trimming.
- Copy nit: "above **the** Low baseline" (SR cue) vs "above Low baseline" (badge); tightening the wording also reduces the double-announcement above.

Rest of the change is clean — `tsc`/eslint green, the mapping and migration are solid, and the unrelated 508 fixes (heading order, notes-textarea label, list semantics, contrast) are self-contained. Happy to re-review once the two payload guards and the tooltip focus are in.

danielbowne:
Thanks Jono, good catches. Pushed a08b696 with all of it:

- **`fips_ceiling` clamp** — fixed before merge. `baselineCeiling` now takes 1-4 only, anything else falls back to 4, so a stray `0` can't invert the floor. Tests at 0/-1/5/NaN.
- **`asImpactLevel` guard** — added the shared narrow-to-literal helper, routed both the strip and the radio group through it so a non-string level suppresses both together. Matching radio-group test added.
- **`hasBaseline`** — now gates on `level != null && ceiling < 4`, same as the strip, so the two surfaces can't disagree.
- **Tooltip focus + double-announce** — `tabIndex={0}` on the baseline/above/FIPS chips, and I drop the sr-only cue on the selected row where the focusable chip already carries the name.

One thing for the record: the page was tested and passing 100% on Lighthouse accessibility before this, so the automated bar was clean. The tooltip keyboard-focus and the double-announce are the kind of thing the scanners don't flag, so genuinely appreciate you reading it by hand. Those are the two that mattered.

Left the two you flagged as notes-only (the post-mount styling announce, the `defaultChecked` fallback) as is, added a comment on the latter.

@MackOverflow when you get a sec, could you re-review? Changes are in since your approval.

MackOverflow (approved):

danielbowne:
Thanks Mack — that one crossed with the push. It's already in: a08b696 adds the shared `asImpactLevel(x)` guard and routes both the strip and the radio group through it, plus `hasBaseline` now gates on `level != null && ceiling < 4` (matching the strip) and there's a malformed-level radio-group test. Landed in the squash merge, so main's covered.

PR #554, opened 2026-07-16
Cover the three effect paths from #481

### LINKED ISSUE #487: Add component (RTL) test coverage for QuestionnairePage effect paths
### Type

Task

### Description

`src/views/QuestionnairePage/QuestionnairePage.tsx` has no component-level test file. Its behavior lives in several interacting effects that recent fixes touched, and those paths are currently covered only indirectly (extracted pure-helper unit tests in `saveGuard.ts`/`scoreSelection.ts`), not by rendering the component.

Stand up a React Testing Library harness for the page (mock `axiosInstance`, `useContextProp`, and the router) and add coverage for the effect paths that have none:

- **Scores fetch 403 (auth-handled) still commits the questions batch.** When the scores request fails on an auth-handled path (FORBIDDEN_ORIGIN or a plain controller-level 403 that keeps the page mounted), the sidebar/URL and `questions`/`datacallID`/`questionId` must update together — no half-updated view, no stuck spinner. Reachable by switching the active datacall as well as the system.
- **Read-only draft eviction.** When a session becomes read-only (deadline passes or read-only admin), a lingering draft for the current question is evicted, and an autosave that was already in flight when `isReadOnly` flipped cannot resurrect it.
- **Out-of-band scores re-seed.** Saving a question then navigating back before the post-save scores GET resolves re-seeds the answer from the refreshed scores map — a newly-answered question is not shown as unanswered, and no duplicate score is posted. In-progress unsaved edits are never overwritten.

Context: follow-up from the review of #481 (encrypted-drafts hardening, closes #475). The fixes for the three paths above landed there; this issue tracks the missing rendered-component coverage.

### Acceptance Criteria

- A `QuestionnairePage` test file renders the component with mocked data/router/context.
- Tests cover the three effect paths above (403-batch-commit, read-only eviction, out-of-band re-seed).
- Tests run in the existing Jest + jsdom setup and pass in CI.

tarratsco:
## Summary

Closes #487.

Adds a rendered-component test file for `QuestionnairePage` locking down the three effect paths hardened by the encrypted-drafts work (#481). Before this, those paths were only covered indirectly through the pure-helper unit tests in `saveGuard.ts` and `scoreSelection.ts` - a rendered-component regression would have slipped through.

## Coverage

Seven tests across the three ticket paths plus a companion mount-seed assertion.

### Path 1 - scores 403 (auth-handled) still commits the questions batch

* `scores fetch 403 (auth-handled) still commits questions and opens the target function` - cold system-mount trigger. Rejects `/scores` with an `__authHandled`-tagged error, asserts `functions/7006/options` fires (only reachable after the L672–675 batch committed `setQuestionId`), spinner clears, and no "try again" toast appears.
* `scores 403 (auth-handled) on a datacall switch also commits the questions batch` - datacall-switch trigger. Same effect path via a live context swap; asserts a second scores GET, the second batch's `fetchOptions` fires, and the toast stays silent across both triggers.

### Path 2 - read-only sessions

* `read-only session evicts the current-question draft on mount` - asserts `clearDraft(uid, sys, questionId, datacallID)` and no `loadDraft`.
* `editable to read-only flip disarms an in-flight autosave` - mounts editable, types into the notes textarea, advances the debounce timer past 1s, and captures the `isCurrent` callback the real `saveDraft` was invoked with. Flips context to read-only via the subscribable store, waits for the read-only branch of `fetchOptions` to run, and asserts `isCurrent()` returns `false` (Ref++ at the eviction call-site disarms the in-flight write). Uses `jest.useFakeTimers` with `userEvent.setup({ advanceTimers })`.

### Path 3 - out-of-band scores refresh

* `out-of-band scores refresh re-seeds the answer after save-and-back` - mounts Q1 unanswered, drives Baseline → Next → Back with a deferred `/scores` promise, simulating a laggy network where the second `questionScores` GET stays in flight during the back navigation. Resolves the deferred scores; asserts the L901 `[questionScores, questionId]` effect re-seeds Baseline as checked and that `axios.post` did not fire a duplicate score creation.
* `out-of-band scores refresh does not overwrite an unsaved in-progress edit` - same save-and-back setup, but after landing back on Q1 the user picks Advanced before the laggy scores GET resolves with Baseline saved. `shouldReseedAnswer`'s unsaved-edits guard fires and Advanced stays checked.
* `the questionId effect reads live scores at mount` - companion assertion that the ref-in-deps replacement from #481 seeds correctly and fires exactly once (no self-triggered rerun).

## Harness note

React Router memoizes its `DataRoutes` wrapper internally, so `rerender()` with the same router prop does not propagate to the mocked `useContextProp`. Fixed by swapping to `useSyncExternalStore` so `setMockCtx(...)` forces re-renders through the memo. This unblocks all three context-swap tests (1b, 2b, and the read-only flip in 2b).

The three post-render `setMockCtx` calls are wrapped inside a single `await act(async () => { ... })` block to keep React quiet about state updates outside `act`.

## Testing

* `yarn jest src/views/QuestionnairePage/QuestionnairePage.test.tsx` - 7 passing.
* `make check` and `make test` - clean.
* The one lingering `console.error` in the run is React Testing Library's own "`ReactDOMTestUtils.act` deprecated" warning, which is unrelated to these tests.

### DIFF
--- a/src/views/QuestionnairePage/QuestionnairePage.test.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.test.tsx
@@ -1,0 +1,665 @@
+import { render, screen, waitFor, act } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { createMemoryRouter, RouterProvider } from 'react-router-dom'
+import { Routes as AppRoutes } from '@/router/constants'
+import type { userData } from '@/types'
+
+// Rendered-component coverage for three QuestionnairePage effect paths from
+// the encrypted-drafts hardening (#481, closes #475):
+//   1. Scores fetch 403 (auth-handled) still commits the questions batch.
+//   2. Read-only sessions evict a lingering draft; an in-flight save from
+//      before the flip cannot resurrect it.
+//   3. An out-of-band scores refresh re-seeds the current answer without
+//      overwriting in-progress edits or posting a duplicate score.
+
+jest.mock('@/utils/config', () => ({
+  __esModule: true,
+  default: { INSIGHTS_SUGGEST_FIX_ENABLED: false },
+}))
+
+jest.mock('@/axiosConfig', () => ({
+  __esModule: true,
+  default: { get: jest.fn(), post: jest.fn(), put: jest.fn() },
+}))
+const axios = require('@/axiosConfig').default as {
+  get: jest.Mock
+  post: jest.Mock
+  put: jest.Mock
+}
+
+const saveDraftMock = jest
+  .fn<
+    Promise<boolean>,
+    [string, number, number, number, unknown, () => boolean]
+  >()
+  .mockResolvedValue(true)
+const clearDraftMock = jest
+  .fn<Promise<void>, unknown[]>()
+  .mockResolvedValue(undefined)
+const loadDraftMock = jest
+  .fn<Promise<null>, unknown[]>()
+  .mockResolvedValue(null)
+
+jest.mock('./draftStore', () => ({
+  saveDraft: (...args: unknown[]) =>
+    saveDraftMock(...(args as Parameters<typeof saveDraftMock>)),
+  loadDraft: (...args: unknown[]) => loadDraftMock(...args),
+  clearDraft: (...args: unknown[]) => clearDraftMock(...args),
+}))
+
+const notifyMock = jest.fn()
+jest.mock('@/utils/notify', () => {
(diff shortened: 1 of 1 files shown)


MackOverflow:
Moving this to #555 — fork PRs can't access the org secrets required for Snyk and the deploy CI gates. Your commits and authorship carry over unchanged. Thanks @tarratsco!

PR #557, opened 2026-07-16
Add an opdiv field to the system details page

### LINKED ISSUE #458: System detail: show the OpDiv at the top of the Organization section
### Feature Type
UI Component

### Workflow Description
On the system detail page, the "Organization" section shows CMS-oriented hierarchy fields (Group Acronym, Group Name, Division Name) but never states which OpDiv the system belongs to. Now that ZTMF is multi-OpDiv (HHS plus every OpDiv), you cannot tell whose system you are looking at from this view.

Lead the Organization section with the OpDiv. Show the OpDiv name (resolved from the system's `opdiv_id`) as the top item, for every OpDiv, so it is immediately clear the record is a CMS system, an HHS/OS system, a CDC system, etc. The existing Group/Division fields stay below it as the finer CMS-hierarchy detail (blank/omitted for OpDivs that do not use them).

Frontend-only: the system already carries `opdiv_id`, and the OpDiv list (`GET /api/v1/opdivs`, already used by OpDivAdmin / the user table) resolves `opdiv_id` to a name/code. No backend change.

### Acceptance Criteria
- [ ] The system detail Organization section shows the OpDiv name (from `opdiv_id`) as its top/lead item, in both read and edit views
- [ ] Works for all OpDivs (not CMS-specific), resolved via the existing opdivs endpoint
- [ ] Existing Group Acronym / Group Name / Division Name fields remain, below the OpDiv
- [ ] Graceful when the sub-hierarchy fields are null (common for non-CMS systems)

### Additional Context
Files: `src/views/SystemDetailPage/fieldConfig.ts` (Organization section), `SystemDetailReadView.tsx`, `SystemDetailEditView.tsx`. Part of the HHS multi-OpDiv expansion; sibling to the ISSO-name and HHS-fields work landing under ztmf#394.

MackOverflow:
> **Note:** In-repo copy of #556 by @costacalvin. Moved from the fork so Snyk
>
> Refs #556

---

## Summary

Closes #458.

Adds the OpDiv as the lead field in the Organization section of the system detail page, in both read and edit views. Previously the section only showed CMS-hierarchy fields (Group Acronym, Group Name, Division Name) with no indication of which OpDiv the system belonged to — making it impossible to tell at a glance whether you were looking at a CMS, HHS/OS, CDC, or other OpDiv system.

Estimated Time Taken: 1 hour

## Changes

- **Read view**: `FieldDisplay` showing the resolved OpDiv name above the existing Group/Division fields. Renders `—` when no OpDiv is set.
- **Edit view**: Disabled `TextField` showing the resolved OpDiv name — display-only, matching the `isso_name` treatment, because the backend makes `opdiv_id` immutable on `PUT /fismasystems/{id}` (assignment is managed separately).
- **OpDiv resolution**: `fetchOpDivs(true)` (includes inactive) so systems assigned to a since-deactivated OpDiv still display their name rather than showing blank.
- **PUT body cleanup**: Removed the pre-existing `opdiv_id` field from the system detail `PUT` payload — confirmed the backend's UPDATE path ignores it entirely (comment in `fismasystems.go`: *"Update path of Save() is already immutable on opdiv_id"*).
- **Type safety**: `opdivName` prop is `string | null` (required, not optional) in both `SystemDetailReadViewProps` and `SystemDetailEditViewProps`.

## Acceptance Criteria

- [x] The system detail Organization section shows the OpDiv name (from `opdiv_id`) as its top/lead item, in both read and edit views
- [x] Works for all OpDivs (not CMS-specific), resolved via the existing opdivs endpoint
- [x] Existing Group Acronym / Group Name / Division Name fields remain, below the OpDiv
- [x] Graceful when the sub-hierarchy fields are null (common for non-CMS systems)

## UI
OpDiv field added to Organization section of the system details page
image

Disabled field on Edit, matching ISSO Name field.
image

## Possible Follow-up

1. **Hoist `opdivs` into the shared Outlet context** (same pattern as `datacenterEnvironments` in #459): `fetchOpDivs` is now independently called in four places — `FismaTable`, `EditSystemModal`, `UserTable`, and `SystemDetailPage` — with no shared cache. Each page mount fires a fresh `GET /opdivs`. The fix is to fetch once in `Title.tsx` and distribute via `useContextProp()`, eliminating the redundant calls.

### DIFF
--- a/src/views/SystemDetailPage/SystemDetailEditView.tsx
+++ b/src/views/SystemDetailPage/SystemDetailEditView.tsx
@@ -518,6 +518,8 @@ export default function SystemDetailEditView(props: SystemDetailEditViewProps) {
             sx={{ pb: 0 }}
           />
           <CardContent>
+            {/* id must not match renderEditField's `edit-${key}` pattern — if opdiv_id
+                is ever added to fieldConfig it would produce a duplicate DOM id. */}
             <TextField
               id="edit-opdiv_id"
               label="OpDiv"
--- a/src/views/SystemDetailPage/SystemDetailEditView.tsx
+++ b/src/views/SystemDetailPage/SystemDetailEditView.tsx
@@ -66,6 +66,7 @@ interface SystemDetailEditViewProps {
   // Rendered in the right column between Data Lake Export and Organization,
   // matching the read view's placement.
   targetMaturitySlot?: React.ReactNode
+  opdivName: string | null
 }
 
 function renderEditField(
@@ -517,6 +518,16 @@ export default function SystemDetailEditView(props: SystemDetailEditViewProps) {
             sx={{ pb: 0 }}
           />
           <CardContent>
+            <TextField
+              id="edit-opdiv_id"
+              label="OpDiv"
+              variant="standard"
+              fullWidth
+              disabled
+              margin="normal"
+              InputLabelProps={{ shrink: true, sx: { marginTop: 0 } }}
+              value={props.opdivName ?? ''}
+            />
             {orgFields.map((field) => renderEditField(field, props))}
           </CardContent>
         </Card>
--- a/src/views/SystemDetailPage/SystemDetailPage.tsx
+++ b/src/views/SystemDetailPage/SystemDetailPage.tsx
@@ -3,7 +3,13 @@ import { useParams } from 'react-router-dom'
 import { Box, CircularProgress, Divider, Typography } from '@mui/material'
 import _ from 'lodash'
 
-import { FismaSystemType, FormValidType, FormValidHelperText } from '@/types'
+import {
+  FismaSystemType,
+  FormValidType,
+  FormValidHelperText,
+  OpDiv,
+} from '@/types'
+import { fetchOpDivs } from '@/utils/opdivs'
 import { useContextProp } from '@/views/Title/Context'
 import axiosInstance from '@/axiosConfig'
 import {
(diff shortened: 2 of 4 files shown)


danielbowne:
I got this one.

danielbowne (approved):
LGTM. Pulled the branch and verified locally and tsc clean, SystemDetailPage suite 24/24, the 3 new read-view tests pass.

Checked the one thing with any blast radius, the `opdiv_id` removal from the PUT body: it's safe. The field is rendered non-editable, so `editedSystem.opdiv_id` was always equal to `system.opdiv_id` and sending it was a no-op, and the backend's Save path is immutable on it anyway. Omitting it changes nothing. The optimistic cache write keeps the unchanged value too, so the resolved OpDiv name stays correct after a save.

Rest is clean: fetch effect has proper abort + auth-handled + error handling, `includeInactive: true` is the right call so a system on a deactivated OpDiv still resolves its name, and the null paths are covered both views (`—` in read, empty disabled field in edit).

Two non-blocking notes, author's call:
- The edit-view disabled OpDiv field and the PUT-body change aren't tested (read view is). A single render assertion on the edit field would close it.
- On an `/opdivs` failure the page shows an error toast for what's otherwise a working page (OpDiv is one read-only field that already falls back to `—`). Could degrade to silent. Minor.

Filed the redundant-fetch follow-up you flagged as #558 (the hoist into Outlet context — it's actually 5 call sites now). Good to merge.

PR #568, opened 2026-07-17
Fix: Source the Assign Systems picker from active systems only

### LINKED ISSUE #532: Decommissioned systems in the user assignment picker
> **Migrated from [CMS-Enterprise/ztmf-misc#218](https://github.com/CMS-Enterprise/ztmf-misc/issues/218)** — originally reported by @eschwein-usds. This is a frontend defect, so it's tracked here in `ztmf-ui`.
>
> _Note: the screenshot below is referenced from the original internal issue's asset store; it renders for authenticated CMS-Enterprise members. See the migration comment on the source issue re: re-attaching for public visibility._

### Description

If the user has selected "Show decommissioned systems" in the main dashboard, only the decommissioned systems appear in the picker

### Steps to Reproduce

1. Go to the main dashboard
2. Select "Show Decommissioned"
3. Go to the menu, select "Edit Users"
4. Select the User ASsignment picker
5. only 8 systems show up

### Expected Behavior

1. The user assignment picker should only show the active systems, regardless of what the user selected on the main dashboard.

### Actual Behavior

The user assignment picker will show only decommissioned systems if selected on the main dashboard.

### Environment

any

### Acceptance Criteria

### Severity

None

### Affected Component

_No response_

### Screenshots

_No response_

### Logs / Error Messages

```shell

```

### Workaround

_No response_

tarratsco:
## Summary

Closes #532.

The user assignment picker (Users page → row action → Assign Fisma Systems) was showing decommissioned systems as its option pool whenever the dashboard's Show Decommissioned toggle was on. Once the toggle poisoned the shared context, an admin could only assign a user to decommissioned systems until they reloaded the app. This PR decouples the picker from that context state so its options stay on the active systems list regardless of what the dashboard is showing.

## Root cause

`Title.tsx` maintains a single `fismaSystems` array in outlet context. The dashboard's Show Decommissioned toggle swaps the endpoint underneath (`/fismasystems` vs `/fismasystems?decommissioned=true`) and replaces that array with whichever set the backend returns. `UserTable` was building its `fismaSystemsMap` from that same context array - so when the toggle was on, the map only carried decommissioned entries, and the `AssignSystemModal` picker's option pool inherited the poison.

## Fix

- `UserTable.tsx` no longer reads `fismaSystems` from `useContextProp()`.
- Removed the map-building block from the existing `/users` load effect and dropped `fismaSystems` from its deps.
- Added a dedicated `useEffect` that fetches `GET /fismasystems` (unparameterized, so active-only) whenever `canRead` flips true, and calls `setFismaSystemsMap(buildFismaSystemsMap(systems))`.
- Extracted `buildFismaSystemsMap(systems)` as an exported pure helper so the fetch → map → picker chain can be unit tested outside of render test uses.

## Behavioral notes

- The picker's option pool is no longer bound to dashboard state - toggling ShowDecommissioned no longer affects the Users page in any way.
- ISSO / non-admin users still respect the existing access restrictions via the call (gated by `canRead`).
- Error, abort, and malformed-response paths are unchanged; existing catch swallow + `signal.aborted` guard behavior remains intact.
- Known pre-existing edge case (not touched here, flagged in review): if a user is assigned to a system that later gets decommissioned, it will be absent from the active map and renders in the picker as a chip with an empty label. Documented in the modal's blank-chip test so a future fallback change can target it intentionally. (Need to create a ticket)

## Testing

Two new test files, twelve tests overall. Both `make check` and `make test` run clean.

`src/views/UserTable/UserTable.test.tsx` (6 tests):
- Fetches `/fismasystems` (active-only) even if the active context is `showDecommissioned: true` and carrying a decommissioned-only array - the regression assertion.
- ISSO / non-admin session issues no fetch (canRead gate).
- Fetch rejection is swallowed cleanly (console.error path, no uncaught rejection).
- Unmount during an in-flight fetch correctly triggers the `AbortSignal` given to axios, resolving the cleanup function upon unmounting, asserting the signal flipped to aborted, then rejecting the pending promise. Confirms the catch block's `signal.aborted` guard prevents logging a `console.error`. Litmus-tested by temporarily removing the guard and observing the test go red.
- Malformed response (`data: null` or missing fields) safely defaults without a crash or broken map build.
- Three pure-function tests for `buildFismaSystemsMap`: base mapping, subsystem name concatenation, and handling of null/undefined/empty arrays.

`src/views/AssignSystemModal/AssignSystemModal.test.tsx` (6 tests):
- Given the active `fismaSystemsMap`, the component surfaces every entry as a selectable option.
- A poisoned decommissioned entry cannot inject options via any path other than the map itself.
- Assigned system present in the map renders a normal chip (labeled path).
- Assigned system id absent from the map renders a chip with an empty label - locks in the pre-existing edge case intentionally for any future fallback changes.

Manual verification: hard reloaded the local server, turned Show Decommissioned on, opened a user's Assign Systems modal, and confirmed the full active list appears in the picker rather than only the decommissioned records.

### DIFF
--- a/src/views/AssignSystemModal/AssignSystemModal.test.tsx
+++ b/src/views/AssignSystemModal/AssignSystemModal.test.tsx
@@ -1,0 +1,151 @@
+// Picker-side coverage for #532. The parent UserTable builds fismaSystemMap
+// from the active /fismasystems response (see UserTable.test.tsx); this
+// file verifies the picker actually surfaces the map's entries as options
+// and doesn't smuggle in the poisoned decommissioned entries any other way.
+
+jest.mock('@/router/router', () => ({
+  __esModule: true,
+  default: { navigate: jest.fn() },
+}))
+
+jest.mock('@/axiosConfig', () => {
+  const axios = require('axios').default
+  const { handleAuthError } = require('@/utils/authInterceptor')
+  const instance = axios.create({ baseURL: '/api/v1/' })
+  instance.interceptors.response.use(
+    (response: unknown) => response,
+    handleAuthError
+  )
+  return { __esModule: true, default: instance }
+})
+
+import { screen, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import MockAdapter from 'axios-mock-adapter'
+import axiosInstance from '@/axiosConfig'
+import AssignSystemModal from './AssignSystemModal'
+import { renderWithProviders } from '@/test-utils/renderWithProviders'
+
+const mock = new MockAdapter(axiosInstance)
+
+const USER_ID = '22222222-2222-2222-2222-222222222222'
+
+const ACTIVE_MAP = {
+  1001: { acronym: 'DS-1', name: 'Death Star' },
+  1101: { acronym: 'ISD-CHI', name: 'Star Destroyer Chimaera' },
+}
+
+function renderModal(
+  overrides: Partial<React.ComponentProps<typeof AssignSystemModal>> = {}
+) {
+  return renderWithProviders(
+    <AssignSystemModal
+      open={true}
+      handleClose={() => {}}
+      userid={USER_ID}
+      userName="Admiral Piett"
+      fismaSystemMap={ACTIVE_MAP}
+      {...overrides}
+    />
+  )
+}
+
+beforeEach(() => {
+  mock.reset()
+})
+
+test('renders the active systems from the map as picker options', async () => {
+  const user = userEvent.setup()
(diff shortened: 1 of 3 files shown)


tarratsco:
Before:

https://github.com/user-attachments/assets/f903284a-2af8-4900-9146-b1987a220149

After:

https://github.com/user-attachments/assets/a5fa0e9e-a9b9-4db2-a16d-c85e9932fa43

MackOverflow:
Moving this to #570 — fork PRs can't access the org secrets required for Snyk and the deploy CI gates, so those jobs get skipped here. Your commits and authorship carry over unchanged. Thanks @tarratsco!

PR #567, opened 2026-07-17
fix(dashboard): Data Call Progress "Complete" chip for past calls, not "0/40 Not updated"

jsos3-cms:
This is the in-repo version of #542 by @voidspooks, with follow-up fixes from @danielbowne, moved onto an in-repo branch so the org-secret CI gates (Snyk, Infrastructure/Deploy) run and it can deploy to dev on green. The commits are carried over unchanged with authorship preserved.

@danielbowne's refinement (`b579da7`) additionally gates the past-call **Complete** chip on answered counts — a past call with zero or partial answers no longer shows Complete — and adds a closed-late gate so a call stops being treated as "current" once its own deadline has passed.

Closes #537. Refs #542.

---

## Problem

On the Home dashboard, selecting a **past** year (e.g. `FY24 ZTM`) showed `0/40 Not updated` (orange) for systems that were in fact **complete** in that call — making completed historical data look missing.

## Root cause (frontend only)

The Data Call Progress chip is a **this-cycle** metric: it derives from `entry.questionsupdated`, the count of functions with a genuine edit event this cycle (ztmf#299 — answers carried over from a prior call don't count until re-saved). A past call has no edits this cycle, so `questionsupdated = 0` for those systems → `0/40 Not updated`, regardless of whether the call was completed. "Updated this cycle" has no meaning once a call is closed.

## Fix

Gate the current-cycle framing to the **current/active** data call. A row's displayed call is current when `chosenCallMap[fismasystemid] === latestDataCallId` (both already available — `chosenCallMap` from `buildDashboardMaps`, `latestDataCallId` from context).

- **Past call, complete** → neutral **Complete** chip (gated on answered counts per the follow-up fix, with a score as the fallback completion signal until the backend answered-count field lands).
- **Current call** → unchanged `N/40` fraction + Updated/Not-updated chip.
- The orange "Not updated" chip and the "Not updated only" filter are now **current-call-only**, so a past year surfaces no false laggards.
- The `0/0` **N/A** case and the Zero Trust Score column are unchanged.

Nothing is hardcoded to a specific call: `latestDataCallId` is derived dynamically (furthest-out deadline), so this stays correct for every future cycle, including the 2026+ single-call-per-year setup.

## Changes

- `progressColumn.tsx` — `ProgressCell` takes `isCurrentCall` / `hasScore`; renders the Complete chip for a completed past call (defaults preserve current behavior for existing callers).
[...]

PR #569, opened 2026-07-17
fix(dashboard): "Show Decommissioned" toggle enables and clears via "Clear filters" (#566)

### LINKED ISSUE #566: fix(dashboard): "Show Decommissioned" toggle does not enable "Clear filters" button
## Summary

## Root cause (frontend)

## Fix direction (pick one)

## Acceptance criteria

## Repro

Found in prod.

danielbowne:
Closes #566.

## Problem

On the Home dashboard filter bar, the **Clear filters** button lit up for an Environment selection or the **Not updated only** switch, but **not** for the **Show Decommissioned** switch — and Clear filters wouldn't reset Show Decommissioned either.

## Cause

`Show Decommissioned` is the odd one out. Environment and Not-updated-only are part of the client-side `DashboardFilterState` object, which is what `hasNoActiveFilters(filters)` (the button's `disabled` source) reads. But Show Decommissioned lives in the Title outlet context, because it gates a server refetch — `Title.tsx` runs `fetchFismaSystems(showDecommissioned)` when it flips, since decommissioned rows are pulled from the API (`?decommissioned=true`), not filtered client-side. So the button never saw it, and `EMPTY_DASHBOARD_FILTERS` never touched it.

## Fix

Treat Show Decommissioned as a de-facto active filter for the button, while leaving it in context where the refetch needs it:

- `disabled={hasNoActiveFilters(filters) && !showDecommissioned}`
- Clear filters also calls `setShowDecommissioned(false)` alongside `onFiltersChange(EMPTY_DASHBOARD_FILTERS)`

Clearing while decommissioned is on drops those rows the correct way — the toggle flip re-runs the fetch without them. When it's already off, the `setShowDecommissioned(false)` is an identical-value set that React short-circuits, so there's no spurious refetch.

## Tests

Exported `QuickSearchToolbar` and added `QuickSearchToolbar.test.tsx`:
- Clear filters disabled when nothing is active
- enabled when only Show Decommissioned is on
- enabled for a normal filter (regression guard)
- Clear filters resets both the filter model and Show Decommissioned
- flipping the switch drives the context setter

## Verified

- `npx tsc --noEmit` clean
- full jest suite green (494 pass)
- lint clean (the one remaining warning is the pre-existing InsightsPanel react-refresh notice)

- A Title-level integration test asserting the clear-triggered refetch drops decommissioned rows (this component's unit tests confirm the setter is called; the refetch itself lives in the Title effect).
- Pre-existing: Clear filters doesn't reset the DataGrid free-text quick-filter — separate from this bug.

### DIFF
--- a/src/views/FismaTable/FismaTable.tsx
+++ b/src/views/FismaTable/FismaTable.tsx
@@ -208,7 +208,7 @@ export function CustomFooterSaveComponent(
   )
 }
 
-function QuickSearchToolbar(props: {
+export function QuickSearchToolbar(props: {
   filters?: DashboardFilterState
   onFiltersChange?: (next: DashboardFilterState) => void
   envOptions?: string[]
@@ -382,8 +382,17 @@ function QuickSearchToolbar(props: {
         <Button
           size="small"
           startIcon={<CloseIcon />}
-          onClick={() => onFiltersChange(EMPTY_DASHBOARD_FILTERS)}
-          disabled={hasNoActiveFilters(filters)}
+          onClick={() => {
+            onFiltersChange(EMPTY_DASHBOARD_FILTERS)
+            // Show Decommissioned lives in Title context (it gates a refetch),
+            // not in the client-side filter model — so clear it separately or it
+            // would survive "Clear filters".
+            setShowDecommissioned(false)
+          }}
+          // ...and it's an active filter for the button's own enabled state:
+          // without this, toggling only Show Decommissioned left Clear filters
+          // greyed out (#566).
+          disabled={hasNoActiveFilters(filters) && !showDecommissioned}
           sx={{ color: '#004297', textTransform: 'none', mr: 2, flexShrink: 0 }}
         >
           Clear filters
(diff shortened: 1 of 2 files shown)


MackOverflow (approved):
Tightly scoped, correctly diagnosed, well-commented, and adequately tested for a unit-level fix. No nits!

jsos3-cms (approved):
Agree with @MackOverflow's approval — tightly scoped and correctly diagnosed. Verified the same locally on top of the just-merged #567: they both touch `FismaTable.tsx`, but different functions (this is `QuickSearchToolbar`; #567 was the progress column), and #569 cherry-picks onto current `main` with no conflict. `tsc --noEmit` clean, eslint clean, full Jest suite green (507 pass), and the new `QuickSearchToolbar` tests cover the real bug (button enabled when only Show Decommissioned is on) plus the reset-both and normal-filter regression cases.

Confirmed the "no spurious refetch" note too: `setShowDecommissioned(false)` when it's already off is a no-op state update, so React bails out and the `Title.tsx` fetch effect doesn't re-fire.

Filed the two items from the Follow-ups section as separate issues so they don't get lost: #572 (Title-level integration test asserting the clear-triggered decommissioned refetch drops those rows) and #573 (pre-existing: Clear filters doesn't reset the DataGrid free-text quick-filter).

PR #576, opened 2026-07-17
Clear filters resets the quick-filter (#573) + Title refetch integration test (#572)

### LINKED ISSUE #566: fix(dashboard): "Show Decommissioned" toggle does not enable "Clear filters" button

### LINKED ISSUE #572: test(dashboard): Title-level integration test for the Clear-filters decommissioned refetch
## Summary

Follow-up to #569 (which closed #566). #569 made "Clear filters" enable for and reset the **Show Decommissioned** toggle. Its unit tests (`QuickSearchToolbar.test.tsx`) confirm the context setter `setShowDecommissioned(false)` is called when Clear filters runs — but the actual server refetch that *drops* the decommissioned rows lives in the `Title.tsx` effect (`fetchFismaSystems(showDecommissioned)`, keyed on the `showDecommissioned` dependency), which the unit test doesn't exercise.

## What's missing

There is no Title-level (integration) test asserting the end-to-end behavior: with Show Decommissioned on, clicking Clear filters re-runs the fetch against the unparameterized `GET /fismasystems` (i.e. **without** `?decommissioned=true`) so the decommissioned rows actually leave the grid.

## Task

Add a `Title`-level rendered/integration test that:
- Enables Show Decommissioned (asserting the fetch runs with `?decommissioned=true` and decommissioned rows are present).
- Triggers Clear filters.
- Asserts the refetch fires against the unparameterized endpoint and the decommissioned rows are dropped.

## Notes

- Purely test coverage; no production code change expected.
- Complements the existing `QuickSearchToolbar` unit tests, which already lock in that the setter is called.

### LINKED ISSUE #573: fix(dashboard): Clear filters doesn't reset the DataGrid free-text quick-filter
## Problem

On the Home dashboard filter bar, **Clear filters** resets the client-side `DashboardFilterState` (Environment, "Not updated only") and — after #569 — the **Show Decommissioned** toggle. But it does **not** clear the DataGrid's built-in **free-text quick-filter** (the search box). After clicking Clear filters, a previously typed search term persists and keeps filtering the grid.

## Steps to reproduce

1. On the Home dashboard, type a term into the DataGrid quick-search box (the grid filters to matching rows).
2. Optionally set an Environment / Not-updated-only / Show Decommissioned filter as well.
3. Click **Clear filters**.
4. Observe: the Environment / Not-updated-only / Show Decommissioned selections reset, but the quick-search text and its filtering remain.

## Expected

Clear filters also clears the free-text quick-filter, so the grid returns to a fully unfiltered state.

## Notes

- **Pre-existing** — surfaced while fixing #566 / #569, but separate from that bug (that one was about the Show Decommissioned toggle).
- The quick-filter lives in the MUI DataGrid's own filter model, not in `DashboardFilterState`, so clearing it needs to reset the grid's `quickFilterValues` (via `apiRef`/the filter model), not just `EMPTY_DASHBOARD_FILTERS`.

costacalvin:
Closes #572
Closes #573

Follow-ups to #569 (which closed #566).

## Summary

Two spin-offs from #569, handled together:

- **#573 (bug):** "Clear filters" reset the client-side `DashboardFilterState` (Environment, Not-updated-only) and the Show Decommissioned toggle, but **not** the DataGrid's built-in free-text quick-filter (search box). A typed search term survived Clear and kept filtering the grid. This PR makes Clear also reset the quick-filter, and enables the button whenever a search term is active so it's actually reachable.
- **#572 (test only):** #569's unit tests lock in that Clear calls `setShowDecommissioned(false)`, but nothing exercised the actual server refetch in `Title.tsx` that *drops* the decommissioned rows. This PR adds a Title-level integration test for that end-to-end path. No production change.

### #573 — production fix
`src/views/FismaTable/FismaTable.tsx` (`QuickSearchToolbar`):
- Read the grid's quick-filter reactively via `useGridApiContext()` + `useGridSelector(apiRef, gridQuickFilterValuesSelector)` — the repo's idiomatic pattern for reaching the grid API from a slot (cf. `UserTable/EditOpDivCell`).
- **Clear filters `onClick`** now also calls `apiRef.current.setQuickFilterValues([])`.
- **Clear filters `disabled`** now also counts the quick-filter (`!hasQuickFilter`), so the button is usable when only a search term is typed — otherwise it stayed greyed out and couldn't clear the term.

Since the grid has `disableColumnFilter`, the quick-filter is the only grid-level filter, so Clear now returns the grid to a fully unfiltered state.

### #573 — tests
`src/views/FismaTable/QuickSearchToolbar.test.tsx` (extends #569's file):
- Extended the `@mui/x-data-grid` mock to stub `useGridApiContext` / `useGridSelector` / `gridQuickFilterValuesSelector` (defaults empty so all 5 existing #569 tests still pass).
- Added: Clear enabled when only a quick-filter is active; Clear stays disabled when nothing is active; Clear calls `setQuickFilterValues([])` alongside the other resets.

### #572 — test only
`src/views/Title/TitleDecommissioned.test.tsx` (new):
- Renders the **real** `Title` so its `showDecommissioned`-keyed refetch effect runs, with a lightweight Outlet-harness that drives `setShowDecommissioned` through the real context (standing in for the deep `Home → FismaTable → DataGrid` subtree, to avoid DataGrid brittleness).
- Asserts the untested path: mount → `GET /fismasystems`; enable → `GET /fismasystems?decommissioned=true` (decommissioned row appears); clear → refetch against the **unparameterized** `/fismasystems` and the decommissioned row is dropped.

**Integration seam (transparency):** the harness's "clear" calls `setShowDecommissioned(false)` directly rather than clicking the literal Clear button (which lives inside the DataGrid we intentionally don't render). The chain is complete because #569's unit tests already prove *Clear → `setShowDecommissioned(false)`*, and this test proves *`setShowDecommissioned(false)` → unparameterized refetch → rows dropped*. The refetch is keyed solely on `showDecommissioned`, so this is the exact call the real button makes.

### #573
- [x] "Clear filters" also clears the free-text quick-filter.
- [x] After Clear, the grid returns to a fully unfiltered state (Environment / Not-updated-only / Show Decommissioned / quick-filter all reset).
- [x] "Clear filters" is clickable when only a search term is active (previously greyed out).
- [x] No regression to #566 behavior (Show Decommissioned still enables/resets via Clear).
- [x] Test coverage for the quick-filter clear + enable paths.

### #572
- [x] Enabling Show Decommissioned asserts the fetch runs with `?decommissioned=true` and decommissioned rows are present.
- [x] Triggering Clear filters re-runs the fetch against the unparameterized `GET /fismasystems` (no `?decommissioned=true`).
- [x] Decommissioned rows are dropped from the grid after Clear.
- [x] Purely test coverage — no production code change.

## UI

Content in the quick search box now enables the clear filter button:
image

"Clear Filter" button now clears content in the quick search box:
image

## Test plan

**Automated**
- `npx jest src/views/FismaTable/QuickSearchToolbar.test.tsx` — 8 pass (5 existing #569 + 3 new #573).
- `npx jest src/views/Title` — 5 pass (existing Title suite + new #572 integration test).
- `npx jest` — full suite green (498 passed, 3 pre-existing skips).
- `npx tsc --noEmit` and `npx eslint` — clean on all changed files.

**Manual (#573)** — verified locally:
1. Home dashboard → type a term in the search box (grid filters). ✅
2. Click **Clear filters** → the search box text clears and the grid returns to unfiltered. ✅
3. With only a search term set, **Clear filters** is enabled (not greyed out). ✅

## Notes
- #573 is pre-existing — surfaced while fixing #566/#569, but separate from the Show Decommissioned bug.
- Stacked on #569; if #569 merges first, this PR will be retargeted to `main`.


jsos3-cms:
Picking this up.

jsos3-cms:
Rehomed to #577 so the Snyk/Deploy CI gates run — they can't on a fork PR. The commits were carried over unchanged with authorship preserved, and since #569 has merged this rehome is just the two follow-up commits. Review is requested there from @danielbowne and @MackOverflow. Keeping this open as a draft for now so it can't accidentally merge; it'll be closed once #577 is merged. Thanks @costacalvin!

jsos3-cms:
Closed by #577

PR #577, opened 2026-07-17
Clear filters resets the quick-filter (#573) + Title refetch integration test (#572)

### LINKED ISSUE #566: fix(dashboard): "Show Decommissioned" toggle does not enable "Clear filters" button

### LINKED ISSUE #572: test(dashboard): Title-level integration test for the Clear-filters decommissioned refetch

### LINKED ISSUE #573: fix(dashboard): Clear filters doesn't reset the DataGrid free-text quick-filter

jsos3-cms:
This is the in-repo version of #576 by @costacalvin, moved onto an in-repo branch so the org-secret CI gates (Snyk, Infrastructure/Deploy) run and it can deploy to dev on green. The commits are carried over unchanged with @costacalvin's authorship preserved. It sits on top of the just-merged #569, so it contains only the two follow-up commits.

Closes #572
Closes #573

Follow-ups to #569 (which closed #566).

## Summary

Two spin-offs from #569, handled together:

- **#573 (bug):** "Clear filters" reset the client-side `DashboardFilterState` (Environment, Not-updated-only) and the Show Decommissioned toggle, but **not** the DataGrid's built-in free-text quick-filter (search box). A typed search term survived Clear and kept filtering the grid. This change makes Clear also reset the quick-filter, and enables the button whenever a search term is active so it's actually reachable.
- **#572 (test only):** #569's unit tests lock in that Clear calls `setShowDecommissioned(false)`, but nothing exercised the actual server refetch in `Title.tsx` that *drops* the decommissioned rows. This adds a Title-level integration test for that end-to-end path. No production change.

## Changes

**`src/views/FismaTable/FismaTable.tsx` (`QuickSearchToolbar`) — #573 fix:**
- Reads the grid's quick-filter reactively via `useGridApiContext()` + `useGridSelector(apiRef, gridQuickFilterValuesSelector)` — the repo's idiomatic pattern for reaching the grid API from a slot (cf. `UserTable/EditOpDivCell`).
- Clear-filters `onClick` now also calls `apiRef.current.setQuickFilterValues([])`.
- Clear-filters `disabled` now also counts the quick-filter (`!hasQuickFilter`), so the button is usable when only a search term is typed.
- Since the grid has `disableColumnFilter`, the quick-filter is the only grid-level filter, so Clear now returns the grid to a fully unfiltered state.

**`src/views/FismaTable/QuickSearchToolbar.test.tsx` — #573 tests:**
- Extends the `@mui/x-data-grid` mock to stub `useGridApiContext`/`useGridSelector`/`gridQuickFilterValuesSelector` (defaults empty so the existing tests still pass).
- Adds: Clear enabled when only a quick-filter is active; Clear disabled when nothing is active; Clear calls `setQuickFilterValues([])` alongside the other resets.

**`src/views/Title/TitleDecommissioned.test.tsx` (new) — #572 test:**
- Renders the real `Title` so its `showDecommissioned`-keyed refetch effect runs, driving `setShowDecommissioned` through the real context.
- Asserts: mount → `GET /fismasystems`; enable → `GET /fismasystems?decommissioned=true` (decommissioned row appears); clear → refetch against the unparameterized `/fismasystems` and the decommissioned row is dropped.

## Acceptance criteria

- [x] "Clear filters" also clears the free-text quick-filter; the grid returns to a fully unfiltered state.
- [x] "Clear filters" is clickable when only a search term is active.
- [x] No regression to #566 behavior (Show Decommissioned still enables/resets via Clear).
- [x] Enabling Show Decommissioned asserts the fetch runs with `?decommissioned=true` and decommissioned rows are present; clearing drops them.

Original fork PR #576 (by @costacalvin): https://github.com/CMS-Enterprise/ztmf-ui/pull/576


MackOverflow (approved):
No real blockers here, looks good!

PR #581, opened 2026-07-20
Refetch active systems when the Assign Systems modal opens

### LINKED ISSUE #574: Assign Systems picker goes stale within a session (no refetch on modal open)
### Feature Type

General Feature

## Recommended labels

## Summary

The Assign Systems picker on the Users page fetches `GET /fismasystems` once when the Users page mounts and never refreshes for the rest of the session. If an admin adds a new FISMA system anywhere in the app while staying on `/users`, that system does not appear in the picker's options until the admin navigates away from the Users page and back. It's not data loss (self-heals on nav), but the moment-of-need behavior is wrong for a picker whose whole job is to show current assignable systems.

## Steps to reproduce

1. Log in as an admin (`OWNER` or `HHS_ADMIN`).
2. Navigate to Users.
3. In a second tab (or via the app's Add Fisma System flow if it's on-page), create a new FISMA system.
4. Return to the Users tab without reloading.
5. Open the Assign Systems modal for any user (e.g. Admiral Piett in the empire seed).
6. Look for the new system in the picker options.

**Expected**: the new system appears in the options.

**Actual**: it doesn't. Navigating away from `/users` and back makes it appear.

## Root cause

In `src/views/UserTable/UserTable.tsx`, the dedicated `/fismasystems` fetch effect (added by #532) keys on `[canRead]` only. No dep changes when the modal opens or when the global systems list mutates, so once the map is built at mount, nothing re-runs the fetch.

Fast-follow from the review of #532. Not a regression from that fix — the pre-#532 code was similarly limited but happened to refresh in some flows via context propagation from `Title.fetchFismaSystems()`. #532 decoupled the picker from context (the correct move), which surfaced this gap.

## Suggested fix

Refetch when the modal opens. Options:

- **Option A (smallest change)**: add an `openModal` boolean or a `modalOpenCount` counter to the UserTable state and include it in the fetch effect's dep array so the fetch fires every time the modal opens.
- **Option B (cleanest colocation)**: move the fetch into `AssignSystemModal`'s existing `[open, userid]` effect so it's part of the modal's own load cycle. Duplicates fetches when multiple users are opened in sequence but keeps the request close to its consumer.

## Out of scope

- Broader realtime sync (WebSockets, SSE). This ticket only closes the "same-session, same-page" staleness window.
- Refreshing the systems list on other views (dashboard, system detail). Those have their own refresh strategies.

## Related

### Acceptance Criteria

- [ ]  Adding a new FISMA system while staying on the Users page makes the new system appear in the picker on the next modal open.
- [ ]  No spurious additional fetches on unrelated interactions.
- [ ]  Existing #532 regression tests still pass (fetch is unparameterized, decommissioned entries filtered, `canRead` gate, unmount-abort guard).

### Additional Context

_No response_

tarratsco:
## Summary

Closes #574.

The Assign Systems picker on the Users page fetched `GET /fismasystems` once at mount and never refreshed. A FISMA system created elsewhere in the session was invisible in the picker until the admin navigated away from `/users` and back - self-heals on nav, but wrong for a picker whose whole job is to show currently assignable systems. This PR fires the fetch when the modal opens so the picker's option pool always reflects the current server state.

## Root cause

The `/fismasystems` fetch effect added in #532 keyed on `[canRead]` only. No dep changes when the modal opens or when the global systems list mutates, so once the map was built at mount, nothing re-ran the fetch. Fast-follow from the review of #532 - not a regression from that fix; #532 correctly decoupled the picker from `context.fismaSystems` and surfaced this pre-existing gap.

## Fix
- `UserTable.tsx` - extracted the `loadActiveSystems` `useCallback` keyed on `canRead`. The mount effect calls it once; `handleOpenModal` calls it directly on each open. An `activeSystemsCtrlRef` (`useRef<AbortController | null>`) holds the latest in-flight controller so a rapid reopen aborts the earlier request rather than racing it. Effect cleanup on unmount aborts via the same ref.
- Removed: intermediate `assignModalOpenTick` state (state-as-signal antipattern).
- Added: `useCallback` and `useRef`.

## Behavioral notes

- Picker's option pool now refreshes every time the modal opens. Adding a system elsewhere in the session shows up in the picker on the next open, no page reload or nav-away/back required.
- No spurious refetches: the fetch is triggered only by mount / `canRead` flip / modal open. Closing the modal does nothing. Other Users-page interactions (Show Deleted toggle, sort, filter) are unaffected.
- Bonus: rapid modal reopen no longer risks a stale earlier response clobbering a later one - the AbortController ref cancels the previous request the moment a new one starts. Visible under DevTools throttling as the earlier request marked `(canceled)`.
- All #532 regression behavior preserved: fetch is still unparameterized (active-only), `canRead` still gates, the catch guard on `signal.aborted` still suppresses spurious console.error on unmount.

## Testing

`UserTable.test.tsx` gets one new test; the existing 8 are unchanged and still pass.

- New: `clicking the Assign Systems row action refetches /fismasystems` - renders UserTable, waits for initial fetch (count = 1), clicks the row's Assign Systems button, asserts count becomes 2. Because MUI DataGrid virtualizes rows and won't render them under jsdom (no layout / no measured height), the test stubs `@mui/x-data-grid`'s `DataGrid` with a minimal row renderer and `GridActionsCellItem` with a plain `<button>` - the actions column is the only surface the test needs, and the DataGrid API is otherwise passed through.
- **Litmus verified**: temporarily removing the `void loadActiveSystems()` call from `handleOpenModal` flipped the new test red. Restored, back to green. So the assertion genuinely protects the fix - a future refactor that drops the direct call will fail this test.
- Existing tests still pass:
  - Fetches `/fismasystems` unparameterized even with `showDecommissioned: true` in context (the #532 regression).
  - `canRead=false` (non-admin) fires zero `/fismasystems` calls.
  - Fetch error is swallowed cleanly (`console.error`, no uncaught rejection).
  - Unmount during an in-flight fetch: the effect cleanup aborts via the ref; the catch guard suppresses `console.error`. Same litmus test on the guard line.
  - Malformed response (`data: null`) doesn't crash the map build.
  - Three `buildFismaSystemsMap` unit tests (base mapping, subsystem concatenation, null/undefined/empty safety).

`make check` and full `yarn jest` clean. 520 tests pass, 3 pre-existing skips.

## Manual verification

1. Log in as an admin, navigate to `/users`, click Assign Systems on any user, note the picker options and close the modal.
2. In a second tab, create a new FISMA system via the Add Fisma System flow.
3. Return to the Users tab (never reloaded), reopen the Assign Systems modal on the same user.
4. **Expected**: the new system appears in the picker options alongside everything from step 1. Before this fix, the new system was absent until the admin navigated away from `/users` and back.

### DIFF
--- a/src/views/UserTable/UserTable.test.tsx
+++ b/src/views/UserTable/UserTable.test.tsx
@@ -12,6 +12,74 @@ jest.mock('@/router/router', () => ({
   default: { navigate: jest.fn() },
 }))
 
+// MUI DataGrid virtualizes rows and won't render them under jsdom (no
+// layout, no measured height). Stub it with a minimal implementation
+// that renders each row's action column so the Assign Systems row action
+// is clickable in tests. Everything else in the DataGrid API is passed
+// through from the real module.
+jest.mock('@mui/x-data-grid', () => {
+  const actual = jest.requireActual('@mui/x-data-grid')
+  const react = require('react')
+  return {
+    ...actual,
+    // GridActionsCellItem uses useGridRootProps and only works inside a
+    // real DataGrid. Replace it with a plain button so it renders under
+    // our mocked DataGrid below.
+    GridActionsCellItem: (props: {
+      icon?: React.ReactNode
+      label?: string
+      onClick?: () => void
+    }) =>
+      react.createElement(
+        'button',
+        {
+          type: 'button',
+          'aria-label': props.label,
+          onClick: props.onClick,
+        },
+        props.label
+      ),
+    // MUI DataGrid virtualizes rows and won't render them under jsdom (no
+    // layout, no measured height). Stub it with a minimal implementation
+    // that renders each row's action column so row-level buttons are
+    // clickable in tests.
+    DataGrid: (props: {
+      rows?: Array<Record<string, unknown>>
+      columns?: Array<Record<string, unknown>>
+      getRowId?: (row: Record<string, unknown>) => string | number
+    }) => {
+      const { rows = [], columns = [], getRowId } = props
+      return react.createElement(
+        'div',
+        { 'data-testid': 'datagrid-mock' },
+        rows.map((row) => {
+          const id = getRowId ? getRowId(row) : (row.id as string | number)
+          return react.createElement(
+            'div',
+            { key: String(id), 'data-testid': `datagrid-row-${id}` },
+            columns.map((col) => {
+              const getActions = col.getActions as
+                | ((params: {
+                    id: string | number
+                    row: Record<string, unknown>
+                  }) => React.ReactNode[])
+                | undefined
+              if (col.type === 'actions' && getActions) {
+                return react.createElement(
+                  'div',
+                  { key: String(col.field) },
+                  getActions({ id, row })
+                )
+              }
+              return null
+            })
+          )
+        })
+      )
+    },
+  }
+})
+
 jest.mock('@/utils/config', () => ({
   __esModule: true,
   default: { IDP_ENABLED: false },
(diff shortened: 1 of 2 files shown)


jsos3-cms:
Rehomed to #582 so the Snyk/Deploy CI gates run — they can't on a fork PR. The commit was carried over unchanged with authorship preserved, and since #570 has merged this rehome is a single follow-up commit on main. Review is requested there from @danielbowne and @MackOverflow. Keeping this open as a draft for now so it can't accidentally merge; it'll be closed once #582 is merged. Thanks @tarratsco!

jsos3-cms:
#582 merged to main — closing this fork rehome as planned. Thanks @tarratsco!

PR #582, opened 2026-07-20
fix(users): refetch active systems when the Assign Systems modal opens

### LINKED ISSUE #574: Assign Systems picker goes stale within a session (no refetch on modal open)

jsos3-cms:
This is the in-repo version of #581 by @tarratsco, moved onto an in-repo branch so the org-secret CI gates (Snyk, Infrastructure/Deploy) run and it can deploy to dev on green. The commit is carried over unchanged with @tarratsco's authorship preserved. It sits on the just-merged #570, so it's a single follow-up commit on top of current main.

Closes #574.

Follow-up to #570 / #532.

## Summary

The Assign Systems picker on the Users page fetched `GET /fismasystems` once at mount and never refreshed. A FISMA system created elsewhere in the session was invisible in the picker until the admin navigated away from `/users` and back — self-heals on nav, but wrong for a picker whose whole job is to show currently assignable systems. This fires the fetch when the modal opens so the picker's option pool always reflects current server state.

## Root cause

The `/fismasystems` fetch effect added in #532 keyed on `[canRead]` only. No dep changes when the modal opens or when the global systems list mutates, so once the map was built at mount, nothing re-ran the fetch. Fast-follow from the review of #532/#570 — not a regression from those fixes; #532 correctly decoupled the picker from `context.fismaSystems` and surfaced this pre-existing gap.

## Fix

- `UserTable.tsx` — extracted a `loadActiveSystems` `useCallback` keyed on `canRead`. The mount effect calls it once; `handleOpenModal` calls it directly on each open. An `activeSystemsCtrlRef` (`useRef<AbortController | null>`) holds the latest in-flight controller so a rapid reopen aborts the earlier request rather than racing it; effect cleanup on unmount aborts via the same ref.
- Removed the intermediate `assignModalOpenTick` state (state-as-signal antipattern).

## Behavioral notes

- The picker's option pool refreshes every time the modal opens; a system added elsewhere in the session shows up on the next open, no reload or nav required.
- No spurious refetches: triggered only by mount / `canRead` flip / modal open. Closing the modal, Show-Deleted toggle, sort, and filter are unaffected.
- Rapid modal reopen no longer risks a stale earlier response clobbering a later one — the AbortController ref cancels the previous request when a new one starts.
- All #532 regression behavior preserved: fetch stays unparameterized (active-only), `canRead` still gates, and the `catch` guard on `signal.aborted` still suppresses spurious `console.error` on unmount.

## Testing

- New test: clicking the Assign Systems row action refetches `/fismasystems` (asserts the call count goes 1 → 2). Litmus-verified — removing the direct `loadActiveSystems()` call from `handleOpenModal` turns it red.
- The 8 existing `UserTable.test.tsx` tests are unchanged and still pass (unparameterized fetch under `showDecommissioned`, `canRead=false` fires zero calls, error swallow, unmount abort, malformed-response safety, `buildFismaSystemsMap` units).
- `make check` and full `yarn jest` clean (520 pass, 3 pre-existing skips).

Original fork PR #581 (by @tarratsco): https://github.com/CMS-Enterprise/ztmf-ui/pull/581

### DIFF
--- a/src/views/UserTable/UserTable.test.tsx
+++ b/src/views/UserTable/UserTable.test.tsx
@@ -12,6 +12,74 @@ jest.mock('@/router/router', () => ({
   default: { navigate: jest.fn() },
 }))
 
+// MUI DataGrid virtualizes rows and won't render them under jsdom (no
+// layout, no measured height). Stub it with a minimal implementation
+// that renders each row's action column so the Assign Systems row action
+// is clickable in tests. Everything else in the DataGrid API is passed
+// through from the real module.
+jest.mock('@mui/x-data-grid', () => {
+  const actual = jest.requireActual('@mui/x-data-grid')
+  const react = require('react')
+  return {
+    ...actual,
+    // GridActionsCellItem uses useGridRootProps and only works inside a
+    // real DataGrid. Replace it with a plain button so it renders under
+    // our mocked DataGrid below.
+    GridActionsCellItem: (props: {
+      icon?: React.ReactNode
+      label?: string
+      onClick?: () => void
+    }) =>
+      react.createElement(
+        'button',
+        {
+          type: 'button',
+          'aria-label': props.label,
+          onClick: props.onClick,
+        },
+        props.label
+      ),
+    // MUI DataGrid virtualizes rows and won't render them under jsdom (no
+    // layout, no measured height). Stub it with a minimal implementation
+    // that renders each row's action column so row-level buttons are
+    // clickable in tests.
+    DataGrid: (props: {
+      rows?: Array<Record<string, unknown>>
+      columns?: Array<Record<string, unknown>>
+      getRowId?: (row: Record<string, unknown>) => string | number
+    }) => {
+      const { rows = [], columns = [], getRowId } = props
+      return react.createElement(
+        'div',
+        { 'data-testid': 'datagrid-mock' },
+        rows.map((row) => {
+          const id = getRowId ? getRowId(row) : (row.id as string | number)
+          return react.createElement(
+            'div',
+            { key: String(id), 'data-testid': `datagrid-row-${id}` },
+            columns.map((col) => {
+              const getActions = col.getActions as
+                | ((params: {
+                    id: string | number
+                    row: Record<string, unknown>
+                  }) => React.ReactNode[])
+                | undefined
+              if (col.type === 'actions' && getActions) {
+                return react.createElement(
+                  'div',
+                  { key: String(col.field) },
+                  getActions({ id, row })
+                )
+              }
+              return null
+            })
+          )
+        })
+      )
+    },
+  }
+})
+
 jest.mock('@/utils/config', () => ({
   __esModule: true,
   default: { IDP_ENABLED: false },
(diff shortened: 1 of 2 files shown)


tarratsco:
Before:

https://github.com/user-attachments/assets/787105cb-fe3f-47c1-88da-2dc30df36b32

After:

https://github.com/user-attachments/assets/bfe5579e-bbf6-49fe-a1e3-982344f471e4

danielbowne:
I got this one.

danielbowne (approved):
Gave this a second pass and it looks solid. The refetch fires on open and only on open, the abort-and-replace handles rapid open/close and superseded responses, and it cleans up on unmount (StrictMode-safe). I also traced that the refreshed map actually reaches the picker: the modal rebuilds its option pool off fismaSystemMap, so this is a real fix and not a wasted call. The new test counts the /fismasystems requests (1 on mount, 2 after clicking Assign Systems), so it genuinely locks in the refetch-on-open behavior rather than just rendering.

Two small non-blocking things, both fine to leave as-is:

1. The modal's assigned-systems effect keys off fismaSystemMap, and this change makes the map update on every open, so /users/{id}/assignedfismasystems now fires twice per open. The first one usually gets aborted when the fresh map lands, it's an idempotent GET, and the result is identical, so there's no visible effect. If you ever want to trim it, keying the option pool off the map without also re-triggering the assigned fetch would do it.
2. The picker shows the previous options for a beat before the refetch swaps them in. Near instant, and freshness is the whole point here.

Approving, nothing blocking.

PR #588, opened 2026-07-21
fix(users): label decommissioned-system assignments in the picker

### LINKED ISSUE #575: Assign Systems picker renders unlabeled chip for assignments to later-decommissioned systems
## Recommended labels

## Summary

When a user is assigned to a FISMA system that later gets decommissioned, the Assign Systems picker still renders a deletable chip for that assignment — but the chip has no visible label. Since #532 sources the picker's option map from the active-systems endpoint only, any id in `users_fismasystems` that no longer matches an active system produces an empty `getOptionLabel` return. The user sees an empty chip with an X-to-delete icon and no way to identify which system it refers to before removing it.

## Root cause

`src/views/AssignSystemModal/AssignSystemModal.tsx`, `getOptionLabel`:

```tsx
getOptionLabel={(option: number) => {
  const system = fismaSystemMap[option]
  return system ? `${system.acronym} - ${system.name}` : ''
}}
```

After #532 the picker's `fismaSystemMap` is built from `GET /fismasystems` (active systems only, decommissioned filtered defensively in `buildFismaSystemsMap`). Decommissioned system ids that remain in `users_fismasystems` cannot be resolved through this map, so the fallback returns `''`.

This edge case pre-existed #532 in a different form: when Show Decommissioned was toggled ON, context.fismaSystems carried the decommissioned entry and the label rendered correctly, at the cost of the picker offering only decommissioned options as targets (the bug #532 fixed). #532 was the right call, but it also locks in the empty-label case, which now surfaces regardless of the toggle.

Currently documented — but not fixed — by the modal test `assigned system id absent from the map renders a chip with an empty label`. That test intentionally locks in the current behavior so a future fix flips it visibly.

## Out of scope

- Automatic cleanup or migration of assignments to decommissioned systems (a policy decision, not a UI defect).
- Any change to the picker's behavior for genuinely unknown ids not caused by decommission (e.g. seed inconsistencies). Same fallback should apply, but the ticket is scoped to the decommission path.

## Related

### Steps to Reproduce

1. In the empire seed, assign a user (e.g. Admiral Piett) to any FISMA system (Piett is seeded assigned to Executor, id 1002).
2. Decommission that system (system detail → Edit → Decommission action).
3. Reload the app as an admin (`OWNER`).
4. Navigate to Users → open the Assign Systems modal for that user.

### Expected Behavior

**Expected**: the picker either shows a readable label for the decommissioned assignment (e.g. `SSD-EX - Super Star Destroyer Executor (decommissioned)`) or clearly indicates which system the empty slot refers to.

### Actual Behavior

**Actual**: an empty MUI chip renders with only a delete icon. The user has no way to identify the assignment. MUI also emits the runtime warning `None of the options match with <id>`.

### Environment

local dev

### Acceptance Criteria

- [ ] A user assigned to a decommissioned system sees a readable identifier for that assignment in the picker (label text, badge, or dedicated row).
- [ ] The existing modal test `assigned system id absent from the map renders a chip with an empty label` is updated to assert the new label (or moved/renamed) and reads as an intentional change from the prior documented behavior.
- [ ] Active-system chips still render as `acronym - name` unchanged.
- [ ] MUI's `None of the options match with <id>` runtime warning is either suppressed (by adding a matching option) or explicitly acknowledged as expected.

### Severity

None

### Affected Component

_No response_

### Screenshots

_No response_

### Logs / Error Messages

```shell

```

### Workaround

_No response_

tarratsco:
## Summary

Closes #575.

Assignments to FISMA systems that were later decommissioned rendered as empty MUI chips in the Assign Systems picker. The user saw a deletable chip with a delete icon and no visible text - no way to identify which system the assignment referred to before removing it. MUI also logged a "None of the options match with `<id>`" warning in dev. Root cause: the picker's `fismaSystemMap` was built from active systems only (correct behavior for the dropdown), so any id that lived in `users_fismasystems` but no longer resolved to an active row produced an empty `getOptionLabel` return.

## Fix

Fetch both endpoints, label the chip, hide the entry from the dropdown, gently dim it.

- `UserTable.tsx`
  - `buildFismaSystemsMap` now records a `decommissioned: boolean` flag per entry instead of dropping decommissioned rows. Callers pass the union of active + decommissioned systems.
  - The `/fismasystems` fetch effect runs two requests in parallel - `/fismasystems` (active) and `/fismasystems?decommissioned=true` - via `Promise.allSettled`. If the decommissioned fetch fails, the loader logs a `console.warn` and proceeds with active-only rather than blanking the picker (which would recreate the original context-poisoning symptom).
  - Merge order is decommissioned first, active second - so an active entry overwrites on the unlikely case of a duplicate id.
  - `fismaSystemsMap` state type updated to reflect the new entry shape.

- `AssignSystemModal.tsx`
  - New `FismaSystemEntry` type and `labelFor(id, map)` helper.
  - `getOptionLabel` delegates to `labelFor`: active → `acronym - name`, decommissioned → `acronym - name (Decommissioned)`, unknown → `Unknown or decommissioned system (id X)` fallback.
  - New `filterOptions` wraps MUI's option filter to strip decommissioned entries from the dropdown so an admin can't reassign a retired system.
  - New `renderTags` renders chips: decommissioned chips get `sx={{ opacity: 0.65, fontStyle: 'italic' }}` - subdued but still deletable via the delete X.
  - Unassign confirmation dialog prompt updated: `Are you sure you want to unassign SSD-EX - SuperStar Destroyer Executor (Decommissioned) from Piett?`.
  - Added imports for `Chip` and `createFilterOptions`.

## Behavioral notes

- Chip in the value area: `SSD-EX - SuperStar Destroyer Executor (Decommissioned)` with subdued styling.
- Dropdown: decommissioned entries never appear - admins can never assign a user to a retired system.
- No `onChange`/POST path for a decommissioned system: it only lives in `value` as a pre-existing chip; new assignments come through the dropdown, which strips these.
- No MUI runtime warning: decommissioned systems live in options, so MUI's value-vs-options reconciliation matches. The genuinely-unknown-id fallback still emits the warning - expected and acknowledged in code, since the id isn't in either endpoint so the client just can't identify it.
- Partial-failure UX: if only the decommissioned endpoint fails, orphan chips fall back to `Unknown or decommissioned system (id X)` without a crash. Active picker options are unaffected.
- No change to the decommission flow itself: whether the backend should cascade to `users_fismasystems` on decommission is a policy question worth raising separately.

## Testing

`AssignSystemModal.test.tsx` - 5 tests:

- Active entries surface as dropdown options.
- Decommissioned entries in the map are excluded from the selectable dropdown (positive: DS-1, ISD-CHI appear; DECOM-A does not).
- Active-map assignment renders a standard chip (confirming absence of "Decommissioned" text).
- Assignment to a decommissioned system renders `SSD-EX - Super Star Destroyer Executor (Decommissioned)` with `opacity < 1` and `italic` font - this PR's headline behavior.
- Truly-unknown id still renders the id-based fallback (defense-in-depth).

`UserTable.test.tsx` - 10 tests. Highlights:

- Fetches BOTH `/fismasystems` and `/fismasystems?decommissioned=true` regardless of context.
- `buildFismaSystemsMap` tags decommissioned entries with `decommissioned: true` (paired end-to-end with the modal test above).
- Partial-failure end-to-end: active fetch succeeds, decommissioned fetch fails → drives the DataGrid to open the modal, opens the Autocomplete, and asserts `DS-1 - Death Star` + `ISD-CHI - Star Destroyer Chimaera` actually render, making it observable that `setFismaSystemsMap` was called on the graceful path - a regression that skipped the setState after the warn would fail the test. **Litmus verified.**
- Unmount-abort guard test: strengthened comment now describes the current `Promise.allSettled` + aborted-guard mechanism (not old single-fetch thing). Assertion is unchanged and still litmus-verified.
- Existing mapper unit tests updated to cover the new flag (`decommissioned: false` on active entries).
- Other coverage carried over from previous work: `canRead` gate, generic fetch error swallow, malformed `data: null` handling.

`make check` and `yarn jest` clean. 522 passed, 3 pre-existing skips.

## Manual verification

1. Log in as an admin (Tarkin, OWNER). Confirm any user is assigned to an active system (Piett -> Executor in the empire seed).
2. Navigate to Executor's system detail page → Edit → Decommission. Save.
3. Navigate to Users → click Assign Systems on Piett.
4. **Expected**: the picker shows a subdued italic chip reading `SSD-EX - Super Star Destroyer Executor (Decommissioned)`, and Executor is absent from the dropdown when you click to add.
5. Click the delete X on the chip → confirmation dialog reads the full labeled string → confirm unassigns as expected.
6. Toggle Show Decommissioned on the dashboard → return to Users → picker options are unchanged (the two-fetch design keeps it stable regardless of context state).

### DIFF
--- a/src/views/AssignSystemModal/AssignSystemModal.test.tsx
+++ b/src/views/AssignSystemModal/AssignSystemModal.test.tsx
@@ -1,7 +1,12 @@
-// Picker-side coverage for #532. The parent UserTable builds fismaSystemMap
-// from the active /fismasystems response (see UserTable.test.tsx); this
-// file verifies the picker actually surfaces the map's entries as options
-// and doesn't smuggle in the poisoned decommissioned entries any other way.
+// Picker-side coverage for the Assign Systems modal. The parent
+// UserTable builds fismaSystemMap from both /fismasystems and
+// /fismasystems?decommissioned=true (see UserTable.test.tsx). This file
+// verifies that the picker
+//   - surfaces active entries as options,
+//   - keeps decommissioned entries in the value pool so their assignments
+//     render as labeled chips (with a "(Decommissioned)" suffix and a
+//     subdued visual), and
+//   - filters those decommissioned entries out of the selectable dropdown.
 
 jest.mock('@/router/router', () => ({
   __esModule: true,
@@ -31,8 +36,12 @@ const mock = new MockAdapter(axiosInstance)
 const USER_ID = '22222222-2222-2222-2222-222222222222'
 
 const ACTIVE_MAP = {
-  1001: { acronym: 'DS-1', name: 'Death Star' },
-  1101: { acronym: 'ISD-CHI', name: 'Star Destroyer Chimaera' },
+  1001: { acronym: 'DS-1', name: 'Death Star', decommissioned: false },
+  1101: {
+    acronym: 'ISD-CHI',
+    name: 'Star Destroyer Chimaera',
+    decommissioned: false,
+  },
 }
 
 function renderModal(
@@ -75,17 +84,23 @@ test('renders the active systems from the map as picker options', async () => {
   ).toBeInTheDocument()
 })
 
-test('a poisoned decommissioned map entry never leaks into the picker', async () => {
-  // Simulates the pre-fix bug: if the map contained a decommissioned
-  // system, the picker would render it. This test locks in that the
-  // picker rendering faithfully reflects whatever map it is given, so
-  // upstream #532 fix is the single source of truth for what's shown.
+test('decommissioned entries in the map are hidden from the selectable dropdown', async () => {
+  // The map carries decommissioned entries so their existing assignments
+  // can render as labeled chips (see the "(Decommissioned)" chip test
+  // below). They must NOT appear in the dropdown as new options - an
+  // admin cannot assign a user to a decommissioned system. filterOptions
+  // strips them; this test locks it in.
   const user = userEvent.setup()
   mock.onGet(`/users/${USER_ID}/assignedfismasystems`).reply(200, { data: [] })
 
   renderModal({
     fismaSystemMap: {
-      1001: { acronym: 'DS-1', name: 'Death Star' },
+      ...ACTIVE_MAP,
+      9001: {
+        acronym: 'DECOM-A',
+        name: 'Decommissioned System A',
+        decommissioned: true,
+      },
     },
   })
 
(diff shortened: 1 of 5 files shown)


MackOverflow:
Moving this to #589 — fork PRs can't access the org secrets required for Snyk and the deploy CI gates, so those checks never run here. Your commits and authorship carry over unchanged. Thanks @tarratsco!

PR #590, opened 2026-07-21
feat(#239): display CFACTS data center environment with mismatch flag

voidspooks:
## Summary

Displays the CFACTS-reported data center environment on the System Detail page's ZTMF Insights card (CMS-Enterprise/ztmf#239 item 2; companion to backend report CMS-Enterprise/ztmf#452).

- New "Data Center Environment" field in the System Identity card, read from the provisional `data_center_environment` enrichment payload key. Renders the standard placeholder until the ztmf-insights pipeline ships the field.
- When the CFACTS value disagrees with the system's own `datacenterenvironment` (trimmed, case-insensitive, including when ZTMF has no value recorded), a warning chip shows the ZTMF value so the drift is visible in place. The comparison mirrors the backend `/datacentermismatches` report.

## Screenshot

Screenshot 2026-07-21 at 12 22 37 PM

## Testing

- `npx tsc --noEmit`, `eslint`, and the full jest suite green (531 passed; 4 new tests covering the mismatch flag, case/whitespace tolerance, missing ZTMF value, and absent payload key).
- Verified against the local full stack with seeded enrichment data.

Snyk and other secret-dependent checks will show red as usual for fork PRs (no repo secrets).

PR #591, opened 2026-07-21
feat(#239): display CFACTS data center environment with mismatch flag

jsos3-cms:
> **Note:** In-repo copy of #590 by @voidspooks, moved from the fork so Snyk and the deploy CI gates can run (those require org secrets unavailable to fork PRs). Commits and authorship are unchanged.
>
> Refs #590

---

## Summary

Displays the CFACTS-reported data center environment on the System Detail page's ZTMF Insights / System Identity card, with a "Differs from ZTMF" chip when it disagrees with the system's own recorded `datacenterenvironment` (trim + case-insensitive). Reads the provisional `data_center_environment` enrichment key and renders a placeholder until the ztmf-insights pipeline ships it. Companion to the backend report ztmf#452 (rehomed in-repo as CMS-Enterprise/ztmf#453). Part of #239.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -341,6 +341,9 @@ export type SystemEnrichmentType = {
   fisma_uuid: string
   fisma_acronym: string
   authorization_package_name: string | null
+  // CFACTS-reported data center environment (ztmf#239). Provisional payload
+  // key: absent until the ztmf-insights pipeline ships the field.
+  data_center_environment?: string | null
   primary_isso_name: string | null
   primary_isso_email: string | null
   is_active: boolean | null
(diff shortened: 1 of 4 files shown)


MackOverflow (approved):
really good, small nits, none of them blockers: 

Whitespace-only CFACTS value renders visibly blank. FieldDisplay uses value || '—', but a value like " " is truthy, so it'd show empty space instead of the — placeholder — while normalizeDCE correctly treats it as "no value" for the mismatch check. Slight inconsistency. Passing a normalized/trimmed display value (or letting FieldDisplay trim) would align the two. Very unlikely in practice.
normalizeDCE(cfactsDCE) is computed twice in the dceMismatch expression. Trivial cost; a local const normCfacts = normalizeDCE(cfactsDCE) would read a touch cleaner. Negligible.
Chip has no wrapping margin — it sits directly after a Box(mb:2). Fine visually, just noting it's unlike the mb-spaced fields around it.

jsos3-cms:
Backend companion ztmf#453 (the cross-OpDiv join fix) is in and approved, so converting this to a ready PR to land alongside it.

PR #592, opened 2026-07-21
Fix ARS control chips rendering failing controls twice in ZTMF Insights

### LINKED ISSUE #593: ARS control chips render failing controls twice in ZTMF Insights
## Summary

In the ZTMF Insights drawer, ARS control chips are built from three payload arrays: `ars_satisfied_controls`, `ars_not_satisfied_controls`, and `ars_failing_controls`. These arrays are not mutually exclusive. The pipeline defines `ars_not_satisfied_controls` as every control that is applicable and not Satisfied, which makes it a superset of `ars_failing_controls` (the Other Than Satisfied controls). A failing control therefore appears in both arrays.

The panel mapped each array to chips independently, so a flagged control rendered twice: once as a grey not-satisfied chip and once as a red failing chip, showing the same control in two conflicting states.

## Why it stayed hidden

Other Than Satisfied is rare in Archer, so `ars_failing_controls` is usually empty and the two arrays don't visibly collide. The double-render only surfaces on the systems that actually have a flagged control.

## Impact

Cosmetic but misleading: a reviewer sees the same control marked both not-satisfied and failing, which undercuts trust in the control-coverage signal and could be read as SecurityHub/ARS data being wrong or missing.

## Fix

Frontend-only. Assign each control to one bucket by precedence (failing, then satisfied, then not-satisfied remainder) so it renders once in its most-severe state. No payload change; the arrays are correct as built.

Fixed by #592.

danielbowne:
Closes #593

## Problem

The ZTMF Insights drawer builds its ARS control chips from three payload arrays: `ars_satisfied_controls`, `ars_not_satisfied_controls`, and `ars_failing_controls`. These arrays are not mutually exclusive. The pipeline defines `ars_not_satisfied_controls` as every control that is applicable and not Satisfied, which makes it a superset of `ars_failing_controls` (the Other Than Satisfied controls). A failing control therefore arrives in both arrays.

The panel mapped each array to chips on its own, so any flagged control rendered twice: once as a grey not-satisfied chip and once as a red failing chip. On systems where a control is genuinely flagged, the same control showed up in two states with conflicting colors.

This stayed hidden because Other Than Satisfied is rare in Archer, so `ars_failing_controls` is usually empty and the two arrays don't visibly collide. It only surfaces on the systems that actually have a flagged control.

## Fix

Assign each control to exactly one bucket by precedence before rendering:

- failing wins first (red, Other Than Satisfied)
- then satisfied (green)
- then not-satisfied as the remainder (grey, applicable but not yet satisfied)

Each control now renders once, in its most-severe state. This is a frontend-only change. The payload arrays are correct as built; `ars_not_satisfied_controls` is intentionally the superset, and the derivation now accounts for that.

## Tests

Added a regression test with a control present in both `ars_not_satisfied_controls` and `ars_failing_controls`. It asserts a single chip in the failing state, which fails against the old independent-render code and passes now. Full InsightsPanel suite green, typecheck clean.

### DIFF
--- a/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
+++ b/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
@@ -253,6 +253,37 @@ describe('InsightsPanel', () => {
     expect(screen.getByText('IA-02(08)').textContent).toContain('○')
   })
 
+  it('renders a control present in BOTH not-satisfied and failing once, as failing (arrays are not mutually exclusive)', () => {
+    // The pipeline builds ars_not_satisfied_controls as a SUPERSET of
+    // ars_failing_controls, so a flagged control (AC-17 here) arrives in both.
+    // It must render exactly one chip, red ✗ failing — never a second grey ○.
+    render(
+      <InsightsPanel
+        payload={{
+          ...fullPayload,
+          ars_satisfied_controls: ['IA-01'],
+          ars_not_satisfied_controls: ['IA-02(02)', 'AC-17'],
+          ars_failing_controls: ['AC-17'],
+        }}
+      />
+    )
+    fireEvent.click(screen.getByRole('button', { name: /details/i }))
+    // Exactly one AC-17 chip, and it is the failing (✗) one, not a grey ○.
+    const ac17 = screen.getAllByText('AC-17')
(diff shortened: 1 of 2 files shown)


MackOverflow (approved):
lgtm!

PR #455, opened 2026-07-07
neutralize UI language

jsos3-cms:
Rehome of #453 by @costacalvin so the required Snyk and deploy checks can run. Fork PRs can't reach the org secrets, so those checks sit pending forever on the original.

This is costacalvin's exact commit (`e4430c2` "neutralize UI language"), authorship preserved, no code changes. Full description is carried over below; discussion stays on #453.

Closes #452.

Once CI is green here, I'll close #453 in favor of this PR.

---

## Summary

Closes #452

Neutralizes HHS-branded display copy and internal identifiers in the frontend for multi-agency reuse. The ZTMF UI was built for CMS/HHS but is intended to be adoptable by other agencies — this PR makes the framework language generic while leaving all backend contract terms untouched.

**Layered approach:**
| Layer | Action |
|---|---|
| User-facing display copy ("HHS Metadata", "HHS onboarding load", "HHS-wide admins") | Neutralized via centralized constants |
| UI-internal identifiers (`hhsEditable`, `HHS_METADATA_KEYS`, `FieldSection 'hhs'`, etc.) | Renamed to neutral equivalents |
| Backend contract terms (`HHS_ADMIN`, `opdiv_id`, `/opdivs`, column keys, `@cmsgov/design-system`) | Left as-is — out of scope |

## Changes

- **`src/constants.ts`** — Added 5 `EXTENDED_METADATA_*` label constants as single source of truth for display copy
- **`src/views/SystemDetailPage/fieldConfig.ts`** — `FieldSection 'hhs'` → `'extended'`; all 12 field `section` literals updated; `HHS_METADATA_KEYS` → `EXTENDED_METADATA_KEYS`
- **`src/types.ts`** — `hhsEditable` → `extendedEditable` on `editSystemModalProps`
- **`src/views/SystemDetailPage/SystemDetailEditView.tsx`** — prop, variable, and display copy updated; inline `HHS_LOCK_TOOLTIP` replaced with imported constant
- **`src/views/SystemDetailPage/SystemDetailReadView.tsx`** — variable, guard name, and display copy updated
- **`src/views/SystemDetailPage/SystemDetailPage.tsx`** — import, loop, and prop updated
- **`src/views/EditSystemModal/EditSystemModal.tsx`** — import, prop, variables, save loops, and display copy updated
- **`src/components/AISummaryBadge/AISummaryBadge.tsx`** — tooltip updated from "HHS narrative" to "questionnaire narrative"

## Open Questions Resolved

[...]

PR #595, opened 2026-07-21
Add System Delegate user role

danielbowne:
## Summary

Adds the new `SYSTEM_DELEGATE` user role to the UI so it is selectable when adding or editing a user and gated correctly. Frontend half of the System Delegate work in CMS-Enterprise/ztmf#455; pairs with backend PR CMS-Enterprise/ztmf#458.

A delegate is scoped to specific FISMA systems the same way ISSO and ISSM are: assigned via users_fismasystems, able to read and write the data-call answers on those systems. The one difference is that a delegate is barred from the target-maturity edit control. It is an answers-only role, intended for contractors and support staff.

## What changed

- **`src/types.ts`** - add `SYSTEM_DELEGATE` to the `UserRole` union.
- **`src/utils/userRoles.ts`**
  - Add `SYSTEM_DELEGATE` to all three assignable-role tiers (OWNER/HHS/OPDIV), beside ISSM. This drives the role dropdown in the add/edit user form and makes the Assign Fisma Systems row action appear for delegate rows.
  - Decouple the two system concepts that were previously conflated in one set. A new `SYSTEM_ACCESS_ROLES` set (ISSO/ISSM/SYSTEM_DELEGATE) backs `hasSystemAccess`, so a delegate gets dashboard and system-detail access. `SYSTEM_SCOPED_ROLES` stays ISSO/ISSM only and backs `isSystemScoped`, so `canEditTarget` on the system detail page stays false for delegates. This mirrors the backend carve-out.

No admin sets were touched, so a non-admin delegate sees no user-management, add-system, create-datacall, or email-users affordances. Questionnaire read/write is writer-agnostic, so a delegate writes answers before the deadline exactly like an ISSO, and the backend re-checks system assignment.

## Tests

Extended `userRoles.test.ts`: SYSTEM_DELEGATE rows across every role predicate, an explicit assertion that `hasSystemAccess` is true while `isSystemScoped` is false (the decoupling that defines the role), and that `selectableRoles` offers it to all three admin tiers. Full UI unit suite passes; typecheck clean.

## Rollout

Draft until backend PR #458 is deployed. The role value has to exist server-side before the UI offers it. Closes #455 once both ship.

### Update: email cohort

Also adds `SYSTEM_DELEGATE` as a targetable group in the Email Users dropdown (shown as "System Delegate", beside ISSO/ISSM). Backend #458 accepts the `SYSTEM_DELEGATE` group key on `/massemails`, so an admin can email the delegate cohort directly instead of only via ALL. Covered by a new EmailModal test asserting the option exists and the raw key is posted.

### DIFF
--- a/src/components/EmailModal/EmailModal.test.tsx
+++ b/src/components/EmailModal/EmailModal.test.tsx
@@ -77,6 +77,46 @@ test('success path fires the success snackbar and surfaces sent emails', async (
   expect(mockedNavigate).not.toHaveBeenCalled()
 })
 
+test('offers System Delegate as a targetable email group and posts its key', async () => {
+  // The backend (ztmf#458) accepts a SYSTEM_DELEGATE group key on
+  // /massemails, so the delegate cohort must be selectable here. Assert both
+  // the option exists and the raw key (not the friendly label) is submitted.
+  let posted: Record<string, unknown> = {}
+  mock.onPost('/massemails').reply((config) => {
+    posted = JSON.parse(config.data as string)
+    return [200, { data: ['delegate@example.com'] }]
+  })
+
+  renderWithProviders(<EmailModal openModal={true} closeModal={jest.fn()} />)
+
+  const group = document.querySelector(
+    'select[name="email_group"]'
+  ) as HTMLSelectElement
+  // The option renders with the friendly label but carries the raw role key.
+  const option = group.querySelector(
+    'option[value="SYSTEM_DELEGATE"]'
+  ) as HTMLOptionElement
+  expect(option).not.toBeNull()
+  expect(option.textContent).toBe('System Delegate')
+
+  await userEvent.selectOptions(group, 'SYSTEM_DELEGATE')
+  await userEvent.type(
+    document.querySelector('input[name="email_subject"]') as HTMLInputElement,
+    'hello'
+  )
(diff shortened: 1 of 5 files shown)


MackOverflow (dismissed):
clean, the description is a little confusing/misleading, but not a blocker: 

Description overclaims — the "email cohort" work is not in this PR (blocking for reviewer trust).
The body's "Update: email cohort" section says it "adds SYSTEM_DELEGATE as a targetable group in the Email Users dropdown" and is "Covered by a new EmailModal test." Neither exists — the PR touches exactly 3 files (types.ts, userRoles.ts, userRoles.test.ts), with no EmailModal/email component change. A reviewer trusting the description would sign off on a feature that isn't here. Either a commit was dropped/never pushed, or the description is aspirational and should be trimmed. Reconcile before merge.

danielbowne:
@MackOverflow good catch, and you were reviewing exactly what was on the branch. Turns out it would've helped if I'd actually pushed my last fix :P The email cohort commit was sitting local and never made it up, so the branch really was just the 3 files while the description talked about 5. Pushed it now, so the EmailModal option + its test are on the PR and the description lines up. Thanks for flagging it instead of rubber-stamping.

MackOverflow (approved):
nice fix Danny

PR #596, opened 2026-07-22
Scope the Assign Systems picker to the target user's OpDivs

### LINKED ISSUE #586: Scope the add-system picker to the assignee's OpDiv
### Description

### Steps to Reproduce

1. Sign in as an `OWNER`.

### Expected Behavior

### Actual Behavior

### Acceptance Criteria

### Notes

### Severity

tarratsco:
## Summary

Closes #586. Requires backend PR CMS-Enterprise/ztmf#457.

The Assign Systems picker offered every system in the application rather than only the systems in the target user's OpDiv. An unscoped admin (OWNER / HHS_ADMIN) assigning to an FDA ISSO saw CMS systems in the dropdown and could create exactly the cross-OpDiv assignment the paired backend change now rejects. Root cause: `UserTable` fetched `/fismasystems` once for the whole page, `buildFismaSystemsMap` reduced the response to `{ name, acronym }` and dropped `opdiv_id`, and `AssignSystemModal` received only that map plus `userid` - no target-user OpDiv context at all - then put every map key into options. The admin's own visibility was the basis for the list; it should have been the assignee's.

## Fix

Back the dropdown with the server-scoped endpoint, keep the global lists as label sources only.

### `AssignSystemModal.tsx`

- Fetches the two per-user reads in parallel via `Promise.allSettled`: `/users/:id/assignablefismasystems` (server-scoped, drives the dropdown) and `/users/:id/assignedfismasystems` (chip value). Either failing degrades gracefully rather than blanking the other path.
- New `allSystems` / `decommSystems` props replace the old `fismaSystemMap` prop. The modal no longer takes a prebuilt map.
- `systemMap` merges the three metadata sources into `{ name, acronym, decommissioned }`. Merge order is decommissioned, then global active, then per-user assignable, so the freshest source wins on label text. The sources are disjoint on the flag - the backend applies `WHERE decommissioned=?` unconditionally (`fismasystems.go:106`, defaulting false), so a later loop can never silently clear an earlier decommissioned flag.
- `optionIds` is the union of assignable + currently-assigned, so MUI's value-vs-options reconciliation matches every chip. `filterOptions` narrows the dropdown back to the assignable set, so an out-of-scope orphan chips but can't be re-picked.
- `labelFor` returns the identifiable `Unknown or decommissioned system (id X)` fallback for an id missing from all three sources. This was briefly `''` mid-branch; restored, since the parent's global reads are `allSettled` and a failed `/fismasystems` would otherwise render orphan chips blank.
- `stateOwnerRef` gates the state reset on `userid` actually changing. Reopening for a different user clears chips before the new reads land; reopening for the same user keeps them visible while refreshing underneath.

### `UserTable.tsx`

- New `allSystems` / `decommSystems` state, fetched once in a `canRead`-gated effect via `Promise.allSettled` over `/fismasystems` and `/fismasystems?decommissioned=true`, and passed down as props. Held for as long as the table is mounted, so repeat modal opens reuse the labels.
- If the decommissioned fetch fails, the loader logs a `console.warn` and proceeds - decommissioned assignments chip without the suffix rather than the picker blanking.
- Removed `fismaSystemsMap` state, the `loadFismaSystems` `useCallback`, `activeSystemsCtrlRef`, the `assignModalTargetOpDivIds` memo, the `buildFismaSystemsMap` import, and now-unused imports. `handleOpenModal` no longer triggers a refetch.

### `buildFismaSystemsMap.ts`

- Deleted. Zero live consumers remain.

## On the acceptance criteria

AC4 says to prefer a server-scoped endpoint over client filtering once that endpoint exists. It exists (#457), so AC1–3 - which describe the interim client-filtering design - are superseded rather than independently implemented:

- **AC1** (filter on `opdiv_id ∈ assignedopdivids`): outcome met, mechanism moved server-side.
- **AC2** (retain `opdiv_id` so filtering is possible): moot; response from modal arrives pre-filtered.
- **AC3** (modal receives OpDiv context from `UserTable`): not met literally. The modal passes `userid`; the server resolves scope. Threading OpDiv context as well would mean filtering twice.

## Behavioral notes

- **Dropdown**: contains exactly the systems the assignable endpoint returns, formatted as `acronym - name`. Decommissioned systems never appear - the endpoint filters them server-side.
- **Out-of-scope orphan chips**: a legacy cross-OpDiv assignment is excluded from the assignable set by design, but the global active list still labels it, so the admin can see what they're removing. It stays deletable and stays out of the dropdown.
- **Decommissioned chips**: #575 behavior preserved - `SSD Executor (Decommissioned)` with `opacity: 0.65`, `fontStyle: 'italic'`, still deletable.
- **Already-assigned options**: render checked and disabled in the dropdown - removal goes through the chip rather than a toggle.
- **Assign is not optimistic**: the value commits only after the POST resolves, so a rejected write leaves no chip implying success.
- **Modal open cost**: dropped from four reads to two. Page mount reads run in parallel with the existing `/users` and `/opdivs` reads, so net wall clock is unchanged.

## Testing

`AssignSystemModal.test.tsx` - 19 tests. Globals arrive as props, so the suite drives them directly rather than mocking a third and fourth endpoint.

- **Scoping**: both per-user reads fire on open; dropdown matches the assignable response; empty response yields no options while existing chips still render; an out-of-scope orphan chips labeled but is absent from the listbox; options sorted by acronym regardless of wire order.
- **Labels**: subsystem name appended; decommissioned suffix italicized; an id absent from every source still chips identifiably.
- **Mutations** (previously untested end-to-end): selecting an option POSTs `{ fismasystemid }` and chips it; a failed POST leaves the chip off; removing a chip opens the confirm dialog; DELETEs only on confirm; cancel writes nothing; a decommissioned assignment is still removable.
- **Degradation**: assignable read fails → empty dropdown, chips survive. Assigned read fails → dropdown still usable, no chips.
- **State**: reopening for a different user clears chips before the held fetches resolve; reopening for the same user keeps them visible.

Each new test was litmus verified by breaking the behavior and confirming it flips red - reverting the `labelFor` fallback to `''` (1 fail), bypassing the confirm gate (3 fails), removing the assignable narrowing in `filterOptions` (3 fails), committing the assign optimistically before the POST (1 fail).

`UserTable.test.tsx` - deleted (585 lines). These came from #570 and #582 on main, not from this branch. All twelve targeted removed code - `buildFismaSystemsMap` unit tests, the old fetch effect inside `UserTable`, the in-flight abort guard, and the click-refetch from #582. Equivalent coverage lives in the modal suite above.

`yarn tsc --noEmit` clean, `yarn lint` clean, `yarn jest` 551 passed, 3 pre-existing skips.

## Manual verification

1. Log in as an admin with visibility across OpDivs (Tarkin, OWNER).
2. Navigate to **Users** → click **Assign Systems** on a user in a specific OpDiv.
3. **Expected**: the dropdown lists only active EMPIRE systems. Against the empire seed that's `SLD-GEN` and `SSD-EX` (`RB-1` / `RB-2` are REBELLION and absent; `DS-1` / `SD-TYR` are decommissioned and absent).
4. Assign one, close the modal, reopen for the same user. Chips render immediately with no blank flash while the reads refresh.
5. Open the modal for a user in a different OpDiv. Old chips clear before the new ones arrive, and the dropdown now shows that OpDiv's systems.
6. On a user with a pre-existing cross-OpDiv assignment: chip renders labeled, the delete X works, and the confirm dialog names the system.
7. Click the delete X on any chip → confirm the dialog opens with the system name → confirming unassigns correctly.


a-dipietro:
Rehomed to #597 so the Snyk/deploy CI gates run (they can't on a fork PR). Commits pushed unchanged, authorship preserved. Thanks @tarratsco!

PR #597, opened 2026-07-22
fix(users): scope the Assign Systems picker to the target user's OpDivs

### LINKED ISSUE #586: Scope the add-system picker to the assignee's OpDiv

a-dipietro:
In-repo version of #596 by @tarratsco, moved so the org-secret CI gates
(Snyk, Infrastructure/Deploy) can run — they can't on a fork PR. Commits
pushed unchanged; authorship preserved.

Refs #596
Closes #586

The picker offered every system in the app, letting an unscoped admin create the cross-OpDiv assignments the backend now rejects.

The dropdown now comes from GET /users/:id/assignablefismasystems, which scopes to the assignee's OpDivs server-side. UserTable fetches the global active and decommissioned lists once per page load and passes them down as label sources, so opening the modal costs only the two per-user reads.

Out-of-scope and decommissioned assignments still chip labeled so an admin can see what they are removing, while the dropdown stays narrowed so neither can be re-picked.

buildFismaSystemsMap and UserTable.test.tsx are removed; both targeted the client-filtering path this replaces, and coverage moved to AssignSystemModal.test.tsx.

Closes #586


MackOverflow (dismissed):
Just a nit, but its potentially confusing enough that I would updated it before merge. 

1. Confirm-dialog label diverges from labelFor (minor, correctness/UX). AssignSystemModal.tsx reuses labelFor() everywhere except the unassign confirmation, which hand-rolls the string:

This drops two behaviors labelFor provides:

For a decommissioned assignment, the dialog omits the (Decommissioned) suffix the chip shows.
For an id absent from all sources, the dialog says "unassign this system", while the chip reads "Unknown or decommissioned system (id 9999)". The admin sees one label on the chip and a vaguer one in the confirm prompt for the exact assignment they're revoking.
Recommend just labelFor(pendingUnassign.systemid) for consistency with the chip. The current tests (/unassign DS-1 - Death Star/, /unassign SSD-EX/) don't exercise either divergence, so this passes CI silently.

MackOverflow (commented):
my review is hitting on these three issues, have these already been resolved? 

1. fismaSystemMap is undefined — the modal crashes on render (3 sites)

The prop was removed and the logic replaced by systemMap, but three references to the old identifier survive in AssignSystemModal.tsx:

L75 — inside optionFilter's stringify: const system = fismaSystemMap[option]
L80 — the useMemo dependency array: [fismaSystemMap]
L228 — filterOptions: .filter((o) => !fismaSystemMap[o]?.decommissioned)
useMemo runs during render, so the component throws immediately. The ui / Deploy check is failing with exactly this:

L75 and L228 should reference systemMap; the dep array should be [systemMap].

2. stateOwnerRef is undefined — never declared

The new per-user effect references it at L84 / L87:

There is no const stateOwnerRef = React.useRef<GridRowId | null>(null) anywhere in the file (useRef isn't even used). This is a second ReferenceError; the crash at L80 just fires first. This is the machinery behind the PR-body claim that "reopening for a different user clears the previous chips" — so that behavior is currently non-functional.

3. The picker is not actually scoped — the core fix is missing

Even after fixing #1, the security objective isn't met. assignableIds is computed at L182 and then never used. The comment above optionIds (L~172) states the intent:

filterOptions below narrows the DROPDOWN back to the assignable set so out-of-scope orphans are not re-selectable.

…but filterOptions only strips decommissioned entries. Because optionIds is the union of assignable + currently-assigned, an out-of-scope active orphan (e.g. EXECUTOR/1002 — assigned, not decommissioned) would still appear in the dropdown as a re-selectable option. That's precisely the cross-OpDiv re-assignment this PR set out to prevent. The filter should be:

This also subsumes the decommissioned check (the server excludes decommissioned systems from the assignable set), and it's why the test "an out-of-scope orphan assignment chips… but the DROPDOWN must not offer it" asserts listbox doesn't contain SSD-EX — that test would fail on the current filter even without the crash.

a-dipietro:
Superseded by #614 — rehomed again after #596 was updated. Closing in favor of #614

PR #599, opened 2026-07-22
Record questionnaire question views for time-spent tracking

costacalvin:
## Summary
Frontend for time-spent tracking (#368): the questionnaire records when a user opens a question so time-spent can be measured. There's no user-facing UI — the metrics are delivered as SQL on the backend.

Related issue: CMS-Enterprise/ztmf#368 · Backend PR: CMS-Enterprise/ztmf#463

---

## Requirements this addresses
The issue owner asked for **logs sufficient to measure time spent** on questions, delivered as SQL/Snowflake (not an in-app report). This PR is the logging half: it captures the "start of looking at a question" signal the issue called out as missing. Editors are the primary metric; the classification itself is done server-side.

## What this PR does
- Fires a fire-and-forget `POST events/view` when a question opens (initial load, Next/Back, or sidebar), for editors and read-only viewers alike. No await; errors are swallowed so a failed ping never disrupts answering.
- The body carries only `{fismasystemid, datacallid, questionid}` — editor-vs-viewer is decided server-side, so the client sends no `readonly` flag.

---

## Scope decisions

**Pulled INTO this PR** (was floated as a possible separate ticket):
- The view-event instrumentation itself — included so accurate data accrues now rather than waiting on a follow-up.

**Left as follow-ups** (not in this PR):
- An in-app "Download time spent" button / report view — implementation preserved on branch `feature/timespent-download`.
- Surfacing viewer analytics in the UI.
- Backend-side follow-ups (Snowflake pipeline, integrity hardening, event-table separation, perf index) are listed on CMS-Enterprise/ztmf#463.

---

## Manual testing
1. Run the app + backend locally and open a questionnaire question (e.g. `/#/questionnaire/ssd-ex/FY25_ZTM/identity/imperial-identity-verification`).
2. In the Network tab, confirm a `POST /api/v1/events/view` fires with body `{fismasystemid, datacallid, questionid}` (no `readonly`).
3. Click **Next/Back**/sidebar and confirm one ping per question open.
4. (Optional) In a read-only/past-deadline session, confirm the ping still fires.
5. Verify the recorded time via the backend queries — see the "Manual testing" steps on CMS-Enterprise/ztmf#463.

PR #601, opened 2026-07-22
Record questionnaire question views for time-spent tracking

jsos3-cms:
This is the in-repo version of #599 by @costacalvin, moved onto an in-repo branch so the org-secret CI gates (Snyk, Infrastructure/Deploy) run and the change can deploy to dev on green. The commits are carried over unchanged and authorship is preserved; credit for the work is Calvin Costa's.

Refs #599 · Related issue: CMS-Enterprise/ztmf#368 · Backend companion: CMS-Enterprise/ztmf#464

## What this adds

The frontend half of time-spent tracking (#368). When a questionnaire question opens — initial load, Next/Back, or the sidebar — the page fires a fire-and-forget `POST events/view` carrying `{fismasystemid, datacallid, questionid}`. There is no user-facing UI; the metrics are delivered as SQL on the backend. The body carries the DB question id (not the function id) and no `readonly` flag — editor-vs-viewer is decided server-side. The request is not awaited and errors are swallowed, so a failed ping never disrupts answering.

## Scope

Pulled in now: the view-event instrumentation itself, so accurate data accrues immediately rather than waiting on a follow-up. Left as follow-ups: an in-app download/report view and surfacing viewer analytics in the UI; the backend-side follow-ups are listed on CMS-Enterprise/ztmf#368 / the backend companion.

## Verification

The `QuestionnairePage` suite passes (13/13, including the two new view-ping tests for editor and read-only sessions), and `tsc --noEmit` and eslint are clean. The ping was also confirmed end-to-end against the running backend branch — the real app issued `POST /api/v1/events/view` and the backend recorded the event. Full detail in the review comment below.

### DIFF
--- a/src/views/FismaTable/FismaTable.test.tsx
+++ b/src/views/FismaTable/FismaTable.test.tsx
@@ -98,3 +98,26 @@ describe('CustomFooterSaveComponent download button', () => {
     expect(calledUrl).not.toMatch(/\/export\?$/)
   })
 })
+
+describe('CustomFooterSaveComponent time-spent download button (#368)', () => {
+  it('is disabled when no rows are selected', () => {
+    renderFooter([])
+    expect(
+      screen.getByRole('button', { name: /download time spent/i })
+    ).toBeDisabled()
+  })
+
+  it('GETs the /export/timespent URL with fsids and a blob responseType', () => {
+    renderFooter([1, 2])
+    fireEvent.click(
+      screen.getByRole('button', { name: /download time spent/i })
+    )
+    expect(mockAxiosGet).toHaveBeenCalledWith(
+      expect.stringContaining('/datacalls/42/export/timespent'),
+      expect.objectContaining({ responseType: 'blob' })
+    )
+    const calledUrl: string = mockAxiosGet.mock.calls[0][0]
+    expect(calledUrl).toContain('fsids=1')
+    expect(calledUrl).toContain('fsids=2')
+  })
+})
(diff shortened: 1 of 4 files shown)


jsos3-cms:
## Review summary — go

Rehomed and verified before opening. Results on this branch:

- `QuestionnairePage` suite: 13/13 pass, including the two new tests asserting the view ping fires with the DB question id (not the function id) and no readonly flag, for both an editor and a read-only session.
- `tsc --noEmit` and eslint: clean (the one eslint warning is pre-existing in InsightsPanel and untouched here).
- End-to-end against the running backend branch: loading a questionnaire fired `POST /api/v1/events/view` with `{fismasystemid, datacallid, questionid}` and the backend recorded the event (204). Navigation re-fires the ping per question open, which is the mechanism the backend dwell query depends on.

The effect fires once per question open, uses correct deps, guards until system/datacall/question resolve, and swallows errors safely (no unhandled rejection). No regression in the existing draft-save, scores, or justification flows.

Non-blocking follow-up (coverage, not a defect — the behavior is verified above): no test asserts that Next/Back fires a second ping with the next question's id; the two existing navigation tests are one assertion away from covering it. Worth a fast-follow.

jsos3-cms:
## Non-blocking coverage follow-ups

Regression-protection only — the ping was verified end-to-end (the real app fired `POST /api/v1/events/view` and the backend recorded the event), so this doesn't block the PR. If it warrants a follow-on ticket, let us know.

- No test asserts that Next/Back/sidebar navigation fires a *second* `events/view` ping with the next question's DB id — the per-question re-fire the dwell query relies on. The two existing navigation tests are one `expect(viewPings())` assertion away from covering it.
- Minor: the guard/no-fire case and the error-swallow path are untested; `React.StrictMode` double-fires the ping in local dev only (production unaffected; backend already defers dedup as a listed follow-up).

Companion backend follow-ups and the full list are tracked on CMS-Enterprise/ztmf#368.

MackOverflow (approved):
nice clean PR, no nits!

PR #602, opened 2026-07-22
Assign system picker now allows for acronym match

### LINKED ISSUE #587: Assign-systems picker search should match on acronym, not just system name
### Description

In the "Assign FISMA Systems" picker (`AssignSystemModal`), each option is displayed as `ACRONYM - System Name` and both the acronym and the name are clearly visible in the list. But the search box matches poorly: typing an acronym often fails to surface the matching system. The displayed data is correct - the filtering logic is the problem. Search should reliably match on either the acronym or the system name.

### Steps to Reproduce

1. Open the assign-systems modal for a user (both acronym and name are visible on each row).
2. Type a system's acronym into the search box.
3. The matching system does not filter in as expected.

### Expected Behavior

Typing any substring of either the acronym or the system name (case-insensitive) filters the list to the matching systems.

### Actual Behavior

Search matching is unreliable - acronym text in particular does not consistently surface the matching row, even though the acronym is shown in the list.

### Scope

Frontend only. The system list is already fetched and the acronym is present in the option data; the filtering happens client-side in the MUI `Autocomplete`. No backend change is needed.

### Acceptance Criteria

- The picker filters on BOTH the acronym and the system name, case-insensitive substring.
- Matching uses an explicit `filterOptions` (e.g. MUI `createFilterOptions` with a `stringify` that concatenates acronym and name) so it does not depend on the display-label formatting.
- A system is findable by typing its acronym alone, its name alone, or any substring of either.

### Notes

Same component as CMS-Enterprise/ztmf-ui#586 (OpDiv scoping of the picker) but a separate concern (search/filter quality). Can be worked independently.

### Severity

Medium - usability. Users routinely identify systems by acronym; the current search forces full-name entry.

costacalvin:
## Summary

Hardens the search/filter in the "Assign FISMA Systems" picker (`AssignSystemModal`) so matching runs against the raw `acronym` + `name` fields via an explicit `filterOptions`, instead of implicitly relying on the formatted display label.

Closes #587.

> **Note:** acronym search already worked on `main`. This is a robustness/decoupling change, **not** a fix for broken behavior — see below.

## Why acronym search already worked

The previous code used `createFilterOptions<number>()` with no `stringify`. When `stringify` is omitted, MUI falls back to `getOptionLabel` to produce the text it matches against (`(stringify || getOptionLabel)(option)` in `useAutocomplete`). Here `getOptionLabel` is `labelFor`, which returns `"ACRONYM - Name"` — so the acronym was already part of the searched string, and typing an acronym alone (or a name alone) already surfaced the matching system.

## What this change does

Instead of matching on the *formatted, decorated* label, it supplies an explicit `stringify` that concatenates the raw `${acronym} ${name}` from `fismaSystemMap`:

- **Decouples search from display formatting.** The old searched text included the `" - "` glue plus decorations like `" (Decommissioned)"` and the `"Unknown or decommissioned system (id X)"` fallback. That meant search could incidentally match those decoration tokens, and would silently break if `labelFor`'s formatting ever changed. Matching now depends only on the underlying data.
- **Satisfies the acceptance criteria in #587** explicitly: an intentional `filterOptions` (`createFilterOptions` with a `stringify` concatenating acronym + name), independent of label formatting.
- MUI defaults (`ignoreCase: true`, `matchFrom: 'any'`) preserve the case-insensitive substring behavior.
- The existing decommissioned-stripping filter is preserved and composes on top of search, so decommissioned systems are still never offered as selectable options.

No change to `getOptionLabel`, chip rendering, or the decommissioned display behavior. Frontend only — no backend change.

## UI
image

### Behavioral delta

| | Before | After |
|---|---|---|
| Searched text | `"DS-1 - Death Star (Decommissioned)"` | `"DS-1 Death Star"` |
| Acronym alone | ✅ | ✅ |
| Name alone | ✅ | ✅ |
| Matches label decoration ("Decommissioned", "id N", `" - "`) | ⚠️ yes | ✅ no |
| Breaks if `labelFor` formatting changes | ⚠️ yes | ✅ no |

## Acceptance criteria

- [x] The picker filters on **both** the acronym and the system name, case-insensitive substring.
- [x] Matching uses an explicit `filterOptions` (MUI `createFilterOptions` with a `stringify` that concatenates acronym and name) so it does not depend on the display-label formatting.
- [x] A system is findable by typing its acronym alone, its name alone, or any substring of either.

## Testing

- Added regression tests in `AssignSystemModal.test.tsx` to lock in the intended behavior:
  - filter by acronym
  - filter by system name
  - case-insensitive acronym match
  - decommissioned system not surfaced even on an exact acronym match
- All 9 tests in the suite pass (`yarn jest src/views/AssignSystemModal`).
- Verified manually in the running app.

### DIFF
--- a/src/views/AssignSystemModal/AssignSystemModal.test.tsx
+++ b/src/views/AssignSystemModal/AssignSystemModal.test.tsx
@@ -198,3 +198,102 @@ test('assigned id absent from the map entirely still renders an identifiable fal
   )
   ;(console.error as jest.Mock).mockRestore?.()
 })
+
+test('typing an acronym filters the list to the matching system', async () => {
+  // Regression for #587: the picker must match on the acronym, not just the
+  // system name. Typing "ISD" surfaces ISD-CHI and drops DS-1 / Death Star.
+  const user = userEvent.setup()
+  mock.onGet(`/users/${USER_ID}/assignedfismasystems`).reply(200, { data: [] })
+
+  renderModal()
+
+  const combobox = await screen.findByRole('combobox', {
+    name: /assign fisma systems/i,
+  })
+  await user.click(combobox)
+  await waitFor(() =>
+    expect(screen.getByText(/DS-1\s*-\s*Death Star/i)).toBeInTheDocument()
+  )
+
+  await user.type(combobox, 'ISD')
+
+  await waitFor(() =>
+    expect(
+      screen.getByText(/ISD-CHI\s*-\s*Star Destroyer Chimaera/i)
+    ).toBeInTheDocument()
+  )
+  expect(screen.queryByText(/Death Star/i)).not.toBeInTheDocument()
+})
+
+test('typing a system name filters the list to the matching system', async () => {
+  const user = userEvent.setup()
+  mock.onGet(`/users/${USER_ID}/assignedfismasystems`).reply(200, { data: [] })
+
+  renderModal()
+
+  const combobox = await screen.findByRole('combobox', {
+    name: /assign fisma systems/i,
+  })
+  await user.click(combobox)
+  await waitFor(() => expect(screen.getByText(/ISD-CHI/)).toBeInTheDocument())
+
+  await user.type(combobox, 'Death')
+
+  await waitFor(() =>
+    expect(screen.getByText(/DS-1\s*-\s*Death Star/i)).toBeInTheDocument()
+  )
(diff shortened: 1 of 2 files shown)


MackOverflow:
Moving this to #603 — fork PRs can't access the org secrets required for Snyk and the deploy CI gates, so those jobs get skipped here. Your commits and authorship carry over unchanged. Thanks @costacalvin!

PR #607, opened 2026-07-23
Tarratsco/feature/system delegates

tarratsco:
## Summary

Closes #598 (part of epic CMS-Enterprise/ztmf#467; merge AFTER: https://github.com/CMS-Enterprise/ztmf/pull/473).

The `SYSTEM_DELEGATE` role exists, but the only way to create or assign any user is the admin-gated Users table, which an ISSO cannot reach, so standing up a delegate needs an administrator. This adds an ISSO-facing self-service surface on the FISMA system detail page (add/provision/renew/remove expiring delegates for a system they own) plus the per-OpDiv capability toggle that enables it. The admin Users table is deliberately left unchanged and unexposed to ISSO.

## Fix

Two surfaces, both driven entirely by the new backend endpoints, with all delegate/toggle calls behind a single client seam.

### `src/utils/delegates.ts` (new)

- The one axios seam for the feature: `fetchSystemDelegates`, `searchDelegateCandidates`, `addSystemDelegate` (email-keyed), `removeSystemDelegate`, `renewSystemDelegate`, and `setOpDivDelegateEnabled`.

### `src/views/OpDivAdmin/OpDivAdmin.tsx`, `src/views/Title/Title.tsx`

- Adds an "Add System Delegate Role" switch column, written through the dedicated `PUT /opdivs/:id/system-delegate-enabled` endpoint with a confirm dialog.
- Widens Manage OpDivs to Owner + HHS admin; HHS admin sees the grid and can flip only the toggle, while create/edit/deactivate stay Owner-only. The nav link is shown for the same set.
- Header reads "System Delegate" with the full "Add System Delegate Role" label in the column tooltip and the switch `aria-label`.

### `src/views/SystemDetailPage/SystemDelegatesSection.tsx` (new)

- Roster of current delegates with a per-row status chip (Active / Expiring soon / Expired) and expiry date; renew and remove per row (tooltipped icons).
- Attach an existing eligible delegate via a searchable Autocomplete (backend supplies eligible candidates).
- Provision a new person via a header action dialog (requires name, email, and required expiry defaulting to three months). "Provision", not "invite": the backend creates the account directly, no email is sent.
- Administrator-required (`DELEGATE_REQUIRES_ADMIN`) and capability-off (`DELEGATE_NOT_ENABLED`) responses render as inline guards; a field-map error on the attach path is surfaced rather than swallowed.

### `src/views/SystemDetailPage/SystemDetailPage.tsx`

[...]

PR #473, opened 2026-07-09
fix visual overrun on Add Fisma System modal

costacalvin:
## Summary

Closes #465 — the `OpDiv` label on the "Add FISMA System" modal was visually overrunning the input underline (appearing at or below the line instead of floating above it).

Every other `variant="standard"` TextField in `EditSystemModal` already had `InputLabelProps={{ sx: { marginTop: 0 } }}` to pin the label above the underline. The `opdiv_id` select field was the only one missing it.

## Changes

- Added `InputLabelProps={{ sx: { marginTop: 0 } }}` to the `opdiv_id` TextField in `EditSystemModal.tsx` to match the pattern used by all other fields in the form.

## UI
Before
image

After
image

## Testing

1. Navigate to the app and click **Add FISMA System**
2. Confirm the `OpDiv` label floats above the underline, consistent with all other fields
3. Confirm no regression in edit mode (OpDiv renders correctly with a pre-selected value)

PR #608, opened 2026-07-23
System Delegate self-service UI (rehome of #607)

### LINKED ISSUE #598: System Delegate self-service: delegates section on the FISMA system detail page
Part of CMS-Enterprise/ztmf#467

> Parent-epic decisions resolved 2026-07-22 (see CMS-Enterprise/ztmf#467). Depends on the backend sub-issue endpoint(s). Pending @MackOverflow acknowledgement before implementation starts.

## Feature Type

New UI on an existing page.

## Workflow Description

Add a "Delegates" section to the **FISMA system detail page**, the same page that already hosts the baseline / target-maturity selector (ISSO-editable, delegate-read-only). This keeps the delegate scope implicit: an ISSO is always acting on the one system they are viewing, so there is no need to unlock the admin Users table for ISSO.

**Section behavior:**

- Visible to users assigned to that system (ISSO / ISSM / admin); hidden from delegates themselves, mirroring how the baseline selector is shown.
- A searchable select (MUI `Autocomplete`) to attach an **existing delegate** by name or email, plus an invite path for a person who is not yet a ZTMF user.
  - **Existing delegate:** selectable only when the user is already a `SYSTEM_DELEGATE` in the ISSO's OpDiv (the backend enforces this; the picker should surface only eligible candidates and rely on the backend as the authority).
  - **Invite new:** the ISSO supplies **Name and Email**. OpDiv is not asked; it is inherited from the ISSO. No other user fields are collected.
- A **required** expiration date picker, defaulting to three months from today (see CMS-Enterprise/ztmf#467 decisions 1 and 2).
- A list of the system's current delegates, each showing its expiration and an expired/disabled badge, with a per-row remove control.
- Calls the new backend delegate endpoints; the current-user payload already exposes the ISSO's assigned systems, so no new plumbing is needed to know which systems they own.

**Error and guard states (see CMS-Enterprise/ztmf#467 decision 3):** when an entered email matches an existing account that cannot be self-served (an existing non-delegate account, an existing delegate in a different OpDiv, or an account with no OpDiv set), the backend returns an administrator-required error. The section shows a clear inline message that an administrator must handle that user, rather than silently failing or offering to invite.

**OpDiv-level enablement toggle (see CMS-Enterprise/ztmf#467 decisions 6 and 7):**

- Add an "Add System Delegate Role" toggle to the **Manage OpDivs** panel, one per OpDiv.
- The toggle is visible and editable **only** to the **HHS admin** and **Owner** roles; it is hidden or read-only for every other role.
- Enabling the toggle for an OpDiv turns on the System Delegate self-service surface (the Delegates section above) for systems in that OpDiv; the toggle state is read from and written to the backend endpoint.

**Reuse:** the multi-select `Autocomplete` pattern from `AssignSystemModal`, and the current-user context that already carries the actor's assigned systems and role.

## Acceptance Criteria

- [ ] A "Delegates" section renders on the FISMA system detail page for users assigned to that system; it is hidden from delegate users.
- [ ] A searchable select attaches an existing eligible delegate by name/email; an invite path collects Name + Email for a new person, with OpDiv never shown (inherited from the ISSO).
- [ ] A required expiration date picker is present and defaults to three months from today.
- [ ] Current delegates for the system are listed with expiration and an expired/disabled badge.
- [ ] A remove control revokes a delegate from the system.
- [ ] Administrator-required responses (existing non-delegate, delegate in a different OpDiv, or no OpDiv set) surface a clear inline message; the flow does not offer to invite over an existing email.
- [ ] An "Add System Delegate Role" toggle renders per OpDiv on the Manage OpDivs panel.
- [ ] The toggle is editable only by HHS admin and Owner roles; other roles cannot change it (hidden or read-only).
- [ ] Toggling reflects the OpDiv's System Delegate enablement via the backend endpoint.
- [ ] The admin Users table remains unchanged and is **not** exposed to ISSO.
- [ ] Add/remove/expiry actions call the new backend endpoints and reflect results without a full reload.

## Additional Context

- The `SYSTEM_DELEGATE` role already exists in the frontend role taxonomy and in every admin's assignable-role list; this issue does not add the role, only the ISSO-facing surface for it.
- Gating today is a single admin-read predicate that hides the entire Users surface from ISSO; this feature deliberately lives on the system detail page instead of widening that predicate.

jsos3-cms:
## Summary

In-repo rehome of #607 by @tarratsco (Onix Tarrats Calderon) so the repository CI gates (which fork PRs can't run) execute against these commits. Original authorship is preserved across all 5 commits; the original PR #607 stays open as a draft until this merges.

**Closes #598** · part of epic CMS-Enterprise/ztmf#467 · paired backend is the rehome **CMS-Enterprise/ztmf#474** (rehome of #473).

Frontend for **System Delegate self-service**: the ISSO Delegates section on the FISMA system detail page, and the per-OpDiv "Add System Delegate Role" toggle on the Manage OpDivs panel.

## ⚠️ Merge order

This must merge **after** the backend rehome **ztmf#474** — the UI consumes the delegate endpoints that PR adds. Do not merge this first.

## What changed

- New `SystemDelegatesSection.tsx` (delegate roster + picker) on the system detail read view; Contacts moved to the left column.
- Per-OpDiv System Delegate toggle in `OpDivAdmin`.
- Role gating in `userRoles.ts`: managing delegates is **ISSO + admin only**; ISSM sees the roster **read-only** by design (per epic #467 decision 5; the backend overrode #598's original ISSM-inclusive text).
- `delegates.ts` / `decommission.ts` utils, `types.ts` additions, roster-refresh fix + status chips.

## Status

**Opened as a draft pending review.** It will be converted to a ready PR once the review verdict is Go; if issues are found it stays draft with the findings noted.

## Provenance / verification

5 commits cherry-picked verbatim from #607 (`tarratsco/ztmf-ui:tarratsco/feature/system-delegates`), authorship preserved, atop current `main` (which has advanced past the fork's base — `tsc`/`lint` on the rehomed tree is being verified separately before this is marked ready).

---

## Reviewer changes applied on top of the original (cc @tarratsco)

Resolved in a follow-up commit; the original 5 commits are preserved beneath it:
- The add call opts out of the global auth interceptor (`skipAuthHandling`) and `classifyAddError` now keys on the `DELEGATE_NOT_ENABLED` code (paired backend change in the BE rehome) — so the capability-off case renders the intended in-dialog guard instead of the previously-unreachable branch behind the interceptor's generic toast. The test was updated to assert the coded rejection (it now reflects the real interceptor-bypass path).

## Deliberately out of scope for this ticket (follow-on)

Flagged so the boundary is explicit — not changed here:
- **Delegate add/renew on a decommissioned system** — not specially gated by this PR or the backend, and not a stated requirement. Open product question for the epic owner: should an ISSO be able to provision delegates on a decommissioned system?
- **Read-view Contacts reading-order** — moving Contacts into the left column changes DOM/reading order; worth an accessibility sign-off during review (not a regression; the existing read-view test still passes).
- **`addMonthsISO` month-overflow** — acceptable for an editable default expiry (per review); left as-is.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -50,6 +50,30 @@ export type OpDiv = {
   name: string
   is_parent: boolean
   active: boolean
+  // Per-OpDiv "Add System Delegate Role" capability. NOT NULL on the backend
+  // (defaults false), so every OpDiv resolves to a real boolean. Gates the
+  // ISSO delegate self-service surface for systems in this OpDiv.
+  system_delegate_enabled: boolean
+}
+
+// A system delegate as returned by GET /fismasystems/:id/delegates (the
+// roster). access_expires_at is RFC3339; null means never expires (only
+// non-delegate users are null, so a delegate row always carries a date).
+export type DelegateRow = {
+  userid: string
+  fullname: string
+  email: string
+  access_expires_at: string | null
+}
+
+// An attachable delegate candidate from
+// GET /fismasystems/:id/delegate-candidates. The backend returns only
+// eligible users (already a delegate, in the system's OpDiv, not deleted,
+// not already attached), so the picker trusts the list as-is.
+export type DelegateCandidate = {
+  userid: string
+  fullname: string
+  email: string
 }
 
 // One known datacenter environment from GET /api/v1/datacenterenvironments.
--- a/src/types.ts
+++ b/src/types.ts
@@ -74,6 +74,11 @@ export type DelegateCandidate = {
   userid: string
   fullname: string
   email: string
+  // The candidate query returns the full user row, so this always arrives
+  // (null only for a non-expiring account). Optional here so fixtures can
+  // omit it. An expired candidate can still be attached, but stays denied at
+  // auth until renewed, so the picker flags it.
+  access_expires_at?: string | null
 }
 
 // One known datacenter environment from GET /api/v1/datacenterenvironments.
--- a/src/utils/decommission.ts
+++ b/src/utils/decommission.ts
@@ -13,6 +13,24 @@ export function getTodayISO(): string {
   return `${yyyy}-${mm}-${dd}`
 }
 
+/**
+ * Returns the date `months` months from today as an ISO date string
+ * (YYYY-MM-DD) in local time. Used to seed the delegate expiration picker
+ * with its +3-month default. Day-of-month overflow (e.g. Aug 31 + 6mo)
+ * rolls forward via the Date constructor, which is fine for a default.
+ *
+ * @param months - Number of months to add to today.
+ * @returns The shifted date as a YYYY-MM-DD string.
+ */
+export function addMonthsISO(months: number): string {
+  const d = new Date()
+  d.setMonth(d.getMonth() + months)
+  const yyyy = d.getFullYear()
+  const mm = String(d.getMonth() + 1).padStart(2, '0')
+  const dd = String(d.getDate()).padStart(2, '0')
+  return `${yyyy}-${mm}-${dd}`
+}
+
 /** Truncates notes for display in confirmation dialogs */
 export function truncateNotes(notes: string): string {
   const trimmed = notes.trim()
--- a/src/utils/delegates.ts
+++ b/src/utils/delegates.ts
@@ -76,7 +76,12 @@ export async function addSystemDelegate(
   systemId: number,
   body: AddDelegateBody
 ): Promise<void> {
-  await axiosInstance.post(`/fismasystems/${systemId}/delegates`, body)
+  // skipAuthHandling so the capability-off 403 reaches this caller instead of the
+  // global interceptor swallowing it into a generic toast; the component keys on
+  // the DELEGATE_NOT_ENABLED code to render an in-dialog guard.
+  await axiosInstance.post(`/fismasystems/${systemId}/delegates`, body, {
+    skipAuthHandling: true,
+  })
 }
 
 /**
--- a/src/utils/delegates.ts
+++ b/src/utils/delegates.ts
@@ -1,0 +1,135 @@
+import axiosInstance from '@/axiosConfig'
+import type { DelegateRow, DelegateCandidate, OpDiv } from '@/types'
+
+/**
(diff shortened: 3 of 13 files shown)


jsos3-cms:
**Reviewed — Go.** `tsc --noEmit` clean, lint clean (one pre-existing unrelated `InsightsPanel` warning), full jest suite 586 passing — verified on this rehome branch, which sits on newer `main` than the fork base and integrates cleanly. Access control is correctly UX-gated on the frontend (`canViewDelegates`/`canManageDelegates`, ISSM read-only, delegate hidden, OpDiv-scoped) and independently enforced at every write path by the backend (#474). The roster-refresh fix (candidate effect keyed on `rosterVersion`, `AbortController`-guarded) is sound and regression-tested; `types.ts` matches the backend JSON contract.

**One fix to apply (before or immediately after landing — non-blocking, but should not be skipped):** the capability-off inline Alert in `SystemDelegatesSection.tsx` (`classifyAddError`) is effectively **dead code**. The global axios interceptor tags every 403 `__authHandled` and fires a generic toast before the component's inline branch runs, so the polished in-dialog "OpDiv disabled" guard never shows in production. Narrow trigger (the toggle is flipped off between page load and the add click) and it fails to a generic toast that still carries the backend's real message — not a security/data issue. Fix is the existing codebase idiom: pass `skipAuthHandling: true` on the `addSystemDelegate` axios call in `src/utils/delegates.ts` (same as `SystemEnrichmentCard.tsx`). Note the mock-boundary test that "covers" this path can't catch it, since it mocks `@/utils/delegates` above the interceptor.

**Non-blocking follow-ups:** request a typed `DELEGATE_NOT_ENABLED` code from the backend so the guard keys on a code, not a bare 403 (also fixes the above cleanly); the PR description references that code but the merged backend only defines `DELEGATE_REQUIRES_ADMIN`; and the read-view Contacts reorder is worth an a11y reading-order sanity check.

⚠️ **Merge order:** this must merge **after** the backend rehome **CMS-Enterprise/ztmf#474**, which adds the endpoints this UI consumes.

MackOverflow (approved):
Clean, there's some good internal flags here, but lgtm

PR #612, opened 2026-07-24
ZTMF Insights: finding-name chips, cross-source ARS control union, per-source sections

### LINKED ISSUE #613: ZTMF Insights: finding-name chips, cross-source ARS control union, per-source sections
Rework the ZTMF Insights drawer so every evidence source reads consistently and every scored source chip has a section that explains it.

### Findings as chips
- Kion and SecurityHub findings render as compact pass/fail chips labeled by the finding name, with description, mapped control, severity, and met/failed state in the hover. SecurityHub collapses from several `SecurityHub:` rows into one block.
- Hardenize shows a section whenever it has data: failing findings as rows with affected domains, a "scanned clean" line when there are none, or a pass/fail chip block once its passing array ships (affected domains carried in the chip hover).

### ARS Controls as a cross-source union
- The ARS Controls section becomes the union of every control any source touches (CFACTS coverage plus each Kion/SecurityHub/Hardenize check's mapped control), rolled up weakest-link. A control one source passes and another fails is flagged as a conflict and shown as an amber chip whose hover names each source.

### Source model
- ARS is the control catalog (the ARS Controls section), not an evidence source, so the standalone ARS chip is dropped and its coverage rolls up under CFACTS.
- Chip order and drawer section order aligned (CFACTS, Hardenize, SecurityHub, Kion); "Based on" derived from the active source chips.

## Backend data enablement (SDL enrichment view, v18)

The passing-check chips and the control union depend on data added to the Security Data Lake enrichment view (`VW_ZTMF_INSIGHTS`) this cycle:

- Added structured `hardenize_passing` and `sechub_passing` arrays (shape `{id, nist_controls, description, level}`, matching the existing `kion_passing`), so all three finding sources honor the same passing-checks contract. Previously the `hardenize_l*_passing` fields were hardcoded NULL, so Hardenize could only ever show a "scanned clean" line.
- View v18 was promoted to the CORE snapshot via CREATE OR REPLACE for the two new columns (service role re-granted, `SNAPSHOT_TS` preserved), then reloaded to local dev and impl (9,880 rows each). Prod carries over through the existing sync lambdas via `SELECT *`.
- Result: EPPEC q36, for example, now ships a 7-element `hardenize_passing` (the DNS checks) that renders as green chips instead of the text line.

Known data gaps, tracked separately and not blocking the frontend:
- `sechub_passing` is always empty in this pipeline (SecurityHub is failing-only, 0 of 11,856 rows).
- `hardenize_passing` currently carries only `id` + `level` (the Hardenize dictionary has 42 entries, 0 with NIST controls or descriptions). Chips render today; the hover text and the ARS-control-union fold light up automatically once the Hardenize dictionary is seeded with NIST + descriptions.

Addressed in #612.

danielbowne:
## What

Reworks the ZTMF Insights drawer so every evidence source reads consistently and every scored source chip has a section that explains it.

### Findings as chips
- Kion and SecurityHub findings render as compact pass/fail chips labeled by the finding name. The hover carries the description, mapped control, severity, and met/failed state. SecurityHub collapses from several `SecurityHub:` rows into one block.
- Hardenize shows a section whenever it has data: failing findings as rows with their affected domains, a "scanned clean" line when there are none, or a pass/fail chip block once its passing array ships. Affected domains also show in the chip hover so nothing is lost when Hardenize renders as chips.

image

### ARS Controls as a cross-source union
- The ARS Controls section is now the union of every control any source touches: CFACTS coverage plus the control each Kion, SecurityHub, and Hardenize check maps to. Each control rolls up weakest-link, so a fail from any source outranks a pass. A control that one source passes and another fails is flagged as a conflict and rendered as an amber chip whose hover names each source and states how it resolved.

### Source model
- ARS is treated as the control catalog (the ARS Controls section), not an evidence source, so the standalone ARS chip is dropped. Its coverage rolls up under CFACTS, which is the source that assesses it.
- Chip order and drawer section order are aligned (CFACTS, Hardenize, SecurityHub, Kion), and "Based on" is derived from the active source chips.

## Backend data enablement (SDL enrichment view, v18)

The passing-check chips and the control union depend on data added to the Security Data Lake enrichment view (`VW_ZTMF_INSIGHTS`) this cycle:

- Added structured `hardenize_passing` and `sechub_passing` arrays (shape `{id, nist_controls, description, level}`, matching the existing `kion_passing`), so all three finding sources honor the same passing-checks contract. Previously the `hardenize_l*_passing` fields were hardcoded NULL, so Hardenize could only ever show a "scanned clean" line.
- View v18 was promoted to the CORE snapshot via CREATE OR REPLACE for the two new columns (service role re-granted, `SNAPSHOT_TS` preserved), then reloaded to local dev and impl (9,880 rows each). Prod carries over through the existing sync lambdas via `SELECT *`.
- Result: EPPEC q36, for example, now ships a 7-element `hardenize_passing` (the DNS checks) that renders as green chips instead of the text line.

Known data gaps, tracked separately and not blocking the frontend:
- `sechub_passing` is always empty in this pipeline (SecurityHub is failing-only, 0 of 11,856 rows).
- `hardenize_passing` currently carries only `id` + `level` (the Hardenize dictionary has 42 entries, 0 with NIST controls or descriptions). Chips render today; the hover text and the ARS-control-union fold light up automatically once the Hardenize dictionary is seeded with NIST + descriptions.

## Testing
- Full suite green (563 passing). Adds `rollupControls` unit tests and expands the panel tests to cover the chip flip, the control union, conflict handling, and the per-source sections.
- Verified against dev data on the network questions: EPPEC q31/q36 and MIDAS q31 (all four sources active, Hardenize pass/fail chips, SecurityHub block, control union with a conflict chip).

## Notes
- Hardenize passing chips render from the structured `hardenize_passing` array. Where the Hardenize dictionary still lacks NIST mappings and descriptions, chips render with id + level; the hover text and the control-union fold happen automatically once the dictionary is seeded.

Closes #613

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -471,7 +471,7 @@ export type InsightPayload = {
   // or []. `ars_satisfied_controls` length == `ars_controls_satisfied`.
   ars_satisfied_controls?: string[] | null
   // Applicable-but-not-satisfied controls (applicable − satisfied). Informational,
-  // rendered greyed — distinct from `ars_failing_controls` (Archer-explicit fails).
+  // rendered greyed — distinct from `ars_failing_controls` (explicit fails).
   // satisfied + not_satisfied == ars_controls_total when the pipeline emits them.
   ars_not_satisfied_controls?: string[] | null
   ars_failing_controls?: string[] | null
--- a/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
+++ b/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
@@ -2,6 +2,7 @@ import { render, screen, fireEvent } from '@testing-library/react'
 import InsightsPanel, {
   OptionInsightBadges,
   severityStyle,
+  rollupControls,
 } from './InsightsPanel'
 import type { InsightPayload } from '@/types'
 import CONFIG from '@/utils/config'
@@ -52,19 +53,16 @@ const fullPayload: InsightPayload = {
 }
 
 describe('InsightsPanel', () => {
-  it('renders the ZTMF Insights label and all five source chips', () => {
+  it('renders the ZTMF Insights label and the four source chips', () => {
     render(<InsightsPanel payload={fullPayload} />)
     expect(screen.getByText('ZTMF Insights')).toBeInTheDocument()
-    // Chips render the source name with a trailing colon (e.g. "Kion:").
-    for (const source of [
-      'Kion',
-      'SecurityHub',
-      'Hardenize',
-      'CFACTS',
-      'ARS',
-    ]) {
+    // Chips render the source name with a trailing colon (e.g. "Kion:"). ARS is
+    // the control catalog (the "ARS Controls" section), not an evidence source —
+    // its coverage rolls up under CFACTS, so there is no standalone ARS chip.
+    for (const source of ['Kion', 'SecurityHub', 'Hardenize', 'CFACTS']) {
       expect(screen.getByText(`${source}:`)).toBeInTheDocument()
     }
+    expect(screen.queryByText('ARS:')).not.toBeInTheDocument()
   })
 
   it('renders the suggested maturity pill when the suggestion differs from the prior score', () => {
@@ -152,17 +150,19 @@ describe('InsightsPanel', () => {
   })
 
   it('renders a finding uniformly: code, title, description, severity, How to fix', () => {
+    // Hardenize is the remaining FindingRow source (Kion/SecurityHub now render as
+    // the compact chip block), so the uniform card shape is asserted through it.
     render(
       <InsightsPanel
         payload={{
           suggested_score: 1,
           findings: {
-            sechub: [
+            hardenize: [
               {
-                id: 'IAM.10',
-                title: 'MFA should be enabled',
-                description: 'Enable MFA for all IAM users.',
-                remediation: 'Turn on MFA in the IAM console.',
+                id: 'WWW_HSTS_MISSING',
+                title: 'HSTS not enabled',
+                description: 'Enable HSTS on this host.',
+                remediation: 'Send a Strict-Transport-Security header.',
                 severity: 'MEDIUM',
               },
             ],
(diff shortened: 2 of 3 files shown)


MackOverflow (approved):
I think this is a big improvement, but I am going to flag two things, one of which you cover in the body: 

⚠️ Semantic shift worth confirming with stakeholders (not a bug): the header count "N of M satisfied" no longer reflects the backend's official ARS totals (ars_controls_total / ars_controls_satisfied are now unused). M is now only the count of controls some source touched. A CMS reviewer used to seeing e.g. "40 of 120 satisfied" will now see a much smaller denominator. The PR body acknowledges this ("the total IS the union of source controls") — flagging so it's a conscious product call, not a silent regression. <<< do we feel like this is confusing? Not sure of a better way to present? 
⚠️ A passing technical check can upgrade a CFACTS "not satisfied" control to "satisfied." Since satisfied (2) > unsatisfied (1), if CFACTS lists a control in ars_not_satisfied_controls but a single Kion check maps a pass to it, the control flips green. That may over-state coverage — a narrow AWS Config-style check passing isn't the same as CFACTS confirming the control. Worth confirming this is the intended interpretation of "weakest-link" (the principle usually argues the other direction — a fail overriding a pass, which it also does correctly). <<< a real blocker, I think we have to bake in that CFACTS is the SOT (per our convo yesterday) even if we're showing cleared checks that might suggest otherwise. This is just a reflection of our 'thin' checks giving false positives, not something we can really fix until we get better coverage, but something we'll want to consider carefully how we display.

danielbowne:
Thanks Mack, and good catches on both.

Big picture first: the win I care about here is that this is finally an accurate baseline. The old panel showed zero trust the way the mockups imagined it, and a lot of that was concept with no real data behind it. Going forward the process flips. The zero trust side brings the actual data and real system examples to the devs, and we build the UI to match the evidence instead of guessing at it. The first HTML passes were ideas and dreams. This is the first version grounded in what the pipeline actually says.

On your two flags:

1. Denominator: agreed, this is a conscious call, not a silent regression. M is now the controls some source actually touched, so it reads smaller than the old ars_controls_total. I am open to showing the official ARS total alongside it if that reads clearer to a CMS reviewer. Happy to work through the presentation with you.

2. This is the real one, and it lines up with the point above: CFACTS is the source of truth. A thin AWS Config style check clearing should not upgrade a control CFACTS marked not satisfied to green. The fail direction is right (a fail overrides a pass), but the pass direction over-states until our checks get real coverage. I will adjust the rollup so a technical pass does not flip a CFACTS not-satisfied to satisfied, and we can decide how to surface the "cleared check but not CFACTS-confirmed" case (likely the same conflict treatment). That keeps CFACTS authoritative and treats the checks as supporting signal, not a second opinion that outranks it.

PR #614, opened 2026-07-24
fix(users): scope the Assign Systems picker to the target user's OpDivs

### LINKED ISSUE #586: Scope the add-system picker to the assignee's OpDiv

a-dipietro:
In-repo version of #596 by @tarratsco, moved so the org-secret CI gates
(Snyk, Infrastructure/Deploy) can run — they can't on a fork PR. Commits
pushed unchanged; authorship preserved.

Redo of #597 - new to pick up changes from review.

Refs #596
Closes #586

The picker offered every system in the app, letting an unscoped admin create the cross-OpDiv assignments the backend now rejects.

The dropdown now comes from GET /users/:id/assignablefismasystems, which scopes to the assignee's OpDivs server-side. UserTable fetches the global active and decommissioned lists once per page load and passes them down as label sources, so opening the modal costs only the two per-user reads.

Out-of-scope and decommissioned assignments still chip labeled so an admin can see what they are removing, while the dropdown stays narrowed so neither can be re-picked.

buildFismaSystemsMap and UserTable.test.tsx are removed; both targeted the client-filtering path this replaces, and coverage moved to AssignSystemModal.test.tsx.

Closes #586


MackOverflow (dismissed):
this looks great, good fixes all around!

jsos3-cms:
Rebased onto `main` to clear the conflict — mergeable again.

The conflict was with #603 (acronym match), which merged after this branch was cut. Rather than let the merge silently drop #603, I combined the two: kept this PR's server-scoped picker (`assignablefismasystems` + the `assignableIds` narrowing) and re-added #603's acronym/name search on top of it. The dropdown filter now stringifies each option as `acronym + name` (matching the raw acronym, not the decorated label) and then narrows to the assignable set — they compose. I also ported #603's acronym/name search tests onto this PR's `assignablefismasystems`/`renderModal` harness.

So there's a small amount of genuinely new code to eyeball — the `optionFilter` (acronym/name stringify) wired into `filterOptions`, plus three restored search tests. Everything else is the approved diff, rebased.

Verified locally on the rebased branch: full frontend suite green (595 passing), `tsc` + lint clean; AssignSystemModal is 22/22 including the restored search tests.

The rebase force-push dismissed the prior approval — @MackOverflow, could you re-review/approve when you get a chance? cc @a-dipietro

jsos3-cms (approved):

MackOverflow (approved):
lgtm!

PR #615, opened 2026-07-24
ZTMF Insights: reorder drawer, alignment disclaimer, failing-first chips, collapse large blocks

danielbowne:
Follow-up to #612, focused on how the drawer reads and scales, and directly addressing the two flags on that PR.

## Reorder
- "Based on" now leads the drawer, then the CFACTS reasoning, then the per-source sections (Hardenize, SecurityHub, Kion). The control-union section moves to the bottom.

## Aligns with ARS Controls
- The control-union section is relabeled "Aligns with ARS Controls:" and the "N of M satisfied" count is removed. The union denominator is not the official ARS total, so that count read as a smaller, misleading figure to a CMS reviewer (the concern raised on #612). The chips still carry each control's state.
- New disclaimer on that section (info button, keyboard reachable, AA): alignment maps automated evidence to ARS controls and is not an assessed control satisfaction determination. CFACTS remains the system of record for control status. This makes the "a cleared technical check is not the same as CFACTS confirming the control" point explicit in the UI.

## Failing-first + collapse
- Failing chips render first in every block, so problems sit at the top.
- The passing bulk collapses behind a "Show all N passing checks" toggle once it exceeds a couple of rows (mainly Hardenize, which can carry 30+ passing checks). Failing findings are always shown and never collapsed.

## Testing
- Full suite green. Panel tests expanded to cover the reorder, the dropped count, the disclaimer (focus + tooltip), failing-first ordering, and the collapse toggle (failing stays visible, passing hidden until expanded).

image
image


jsos3-cms (changes requested):
Nice iteration — the reorder, dropping the misleading union count, failing-first, and the collapse all read well, and the disclaimer is the right call. I ran the branch (58/58 green, `tsc` + lint clean) and dug into the a11y and state behavior. Two items undercut the PR's own "keyboard reachable, AA" goal and should be fixed before merge, plus a couple of test/nice-to-have items and one pre-existing issue worth filing separately.

## Must fix in this PR

1. **The info button is invisible to screen readers.** `InsightsPanel.tsx:647-666` — the icon renders with `aria-hidden="true"` (MUI `SvgIcon`'s default when no `titleAccess` is passed; nothing here overrides it). It's keyboard-focusable and shows the tooltip for sighted users, but it's removed from the accessibility tree, so a screen reader never announces it (`getByRole('img', { name: /About ARS alignment/i })` can't find it). Simplest fix is to wrap it in an `IconButton` with an `aria-label`, matching the existing pattern in `QuestionnareModal.tsx:503-510`; passing `aria-hidden={false}` also works.

2. **Icon contrast is below AA.** `InsightsPanel.tsx:657` — `#8a94a6` on the `#f8f9fe` panel is ~2.91:1, under the 3:1 WCAG 2.1 SC 1.4.11 non-text minimum. The greys already used in this file (`#555`, `#5c636a`) clear 5.8–7:1; one-line swap.

3. **Collapse state leaks across questions.** `InsightsPanel.tsx:945` (`showAllPassing`) — because `<InsightsPanel>` (`QuestionnairePage.tsx:1338`) and `FeedCheckBlock` (`InsightsPanel.tsx:590`) keep the same key across a question change, React reuses the instance and `showAllPassing` persists: expand "Show all N passing checks" on one question and the next opens with passing still expanded, without a click. Key the panel/block to the payload (question) identity, or reset `showAllPassing` in an effect keyed on the payload. Worth pairing with the rerender test below, since nothing currently exercises a payload change.

## Nice to have in this PR

- The reorder test only asserts the two bookends (`Based on` vs `Aligns with ARS Controls`); adding assertions for where CFACTS and the per-source feed sections land would catch a partial-reorder regression.
- No `rerender`-with-changed-payload test exists — adding one would lock in the fix for item 3 and is the coverage gap that let it through.
- A failing-first assertion for the ARS-controls-union chip block (only the feed blocks are covered today; the union code is correct by inspection but untested).
- Nit: the ARS chip `key` (`state-id-index`) is redundant — `id` is already unique per the rollup. Harmless; drop if you're in there.

## File for later (out of scope here)

- The drawer's pre-existing `open`/`details` state (`InsightsPanel.tsx:308`) has the same root cause as item 3 — it also persists across question changes — but it predates this PR. Worth a separate ticket to reset per-question (or key the panel once) rather than widening this PR's scope.

Everything else checked out: failing-first ordering is correct and stable (plain concat, and `conflict` controls correctly surface as failing), the count removal is clean (the similar sentence in `justificationContext.ts` uses a distinct CFACTS backend field, not the removed union total), and the reorder didn't disturb any render guards.

danielbowne:
Good review, all three must-fixes are in `56d00c5`, plus a regression you did not see because it predates this branch.

**1. Info icon a11y.** You were right, and the root cause was worse than the aria-hidden. `SvgIcon` stamps `aria-hidden` when no `titleAccess` is passed, so the role and name sat in a hidden node. But even un-hidden it was still broken: MUI Tooltip merges `children.props` last, so the icon's own `aria-label` won and the tooltip was never wired as `aria-describedby`. A screen reader would have gotten "About ARS alignment" and never the disclaimer itself. So the full disclaimer is now the accessible name, `aria-hidden={false}` explicitly, one announcement, no duplication. Went with that over `IconButton` to keep the 14px glyph's layout as is.

The test was the real miss: it used `getByLabelText`, which reads the attribute directly and passes on a hidden node. Switched to `getByRole('img', ...)` so this fails if it regresses.

**2. Contrast.** Confirmed your number, `#8a94a6` on `#f8f9fe` is 2.91:1. Now `#5c636a` at 5.79:1.

**3. Collapse leak.** Reproduced and fixed. Reset off the rendered check set rather than a question id: the payload carries no question identity (`fismasystemid`/`questionid` are on the `Insight` wrapper, not inside the payload), and keying the panel at its callsite would also reset the drawer's open state, which is the pre-existing issue you flagged for a separate ticket. Two questions with an identical check set are indistinguishable there, and the carried state is invisible in that case. Rerender test added.

**Nice-to-haves,** all four done. The reorder test now asserts through the middle (Based on to CFACTS to SecurityHub to Kion to ARS align) instead of just the bookends, there is a rerender-with-changed-payload test, failing-first is asserted on the ARS union block, and the redundant chip key index is gone.

**One more, not from your list.** Testing your item 3 surfaced that the chip swap in the previous PR dropped the "How to fix" remediation for Kion and SecurityHub. `FindingRow` carried it, and after the swap `FindingRow` only renders for Hardenize. The chip hover had description plus control plus state, no remediation. That matters more than it sounds: some dictionary descriptions are bare labels, so `account-without-compliant-password-policy` hovers as just "Password Policy" while the actual explanation lives in remediation. It is back in the hover and the accessible name, behind the same `INSIGHTS_SUGGEST_FIX_ENABLED` flag `FindingRow` uses, with tests both ways.

62 tests in the panel suite, 570 across the repo, `tsc` and lint clean.

danielbowne:
Two more commits on here. Heads up that one of them is unrelated to the feature and went in out of urgency.

## `383980f` — react-router bump, urgent, needed to get CI through

Three high-severity advisories were published against `react-router@7.16.0` in the last two days, so the snyk dependency scan started failing on every open PR in this repo. This one passed clean on 07-24 and fails today with no code change.

- Cross-site Scripting, SNYK-JS-REACTROUTER-18313128
- Open Redirect, SNYK-JS-REACTROUTER-18313144
- Inefficient Algorithmic Complexity, SNYK-JS-REACTROUTER-18313148

That is not a cosmetic red X. `Analysis` gates the `ui / Deploy` job, and `main` requires a dev deployment to merge, so a failing snyk test skips the deploy and nothing can go in. It would also have blocked the deploy after merge. I fixed it here rather than in a separate PR to unblock immediately, so this branch carries a dependency change it otherwise would not. Happy to split it out if you would rather review it on its own.

Both packages were pinned exact and had drifted apart, `react-router` at 7.16.0 direct plus a second copy at 7.15.0 under `react-router-dom@7.15.0`, which is why each advisory showed two vulnerable paths. Both are now 7.18.0, the upgrade snyk names as the fix, and the lockfile is down to a single resolution with no references to 7.15.0 or 7.16.0 left.

I read the 7.16 to 7.18 changes against this app before bumping. The reverse-proxy CSRF and origin-validation fix applies to the server adapters, which a Vite SPA does not use. There is no `useNavigation` call anywhere in `src/`, so the discriminated-union type change does not bite. No future flags are set, so the `unstable_trailingSlashAwareDataRequests` to `v8_` rename, which is a hard config error rather than a silent fallback, is not in play. Typecheck, lint, the full suite and a production build all pass, and snyk test plus `ui / Deploy` are green on this branch now.

## `4e9edb6` — a real bug in my own previous commit, plus tightening the a11y fix

Putting remediation on the chips introduced a bug I had missed: it was not gated on pass/fail. A passing entry carrying remediation would have told a reviewer how to fix a check that **passed**, which is wrong guidance in a compliance tool, and on Hardenize it would have hung a full remediation paragraph off each of 30-plus focusable passing chips. Nothing in the pipeline emits remediation on passing entries today, I checked and none of the 4,685 current ones carry it, so it was latent rather than live. But the type permits it, so the guard is on the render path. It matches the call the passing arrays already make for severity: a pass is not a severity event, and it is not something to fix either.

On your info-button item, my first fix worked but was the wrong shape. Stuffing all 240 characters of the disclaimer into `aria-label` produced a ~260 character accessible name and left the tooltip's own text orphaned in the accessibility tree, reachable a second time by a virtual cursor. Now `describeChild` wires the disclaimer as `aria-describedby`, so the name stays "About ARS alignment" and the prose is the description, which is the name/description split that was wanted in the first place. The title is passed as a node rather than a string deliberately: with a string, Tooltip also sets a native `title` attribute and the browser draws its own tooltip on top of ours.

The collapse reset is also better than what I first shipped. Deriving the reset key from the rendered check ids aliased in the case that matters most, since the same check maps to several questions, so identical ids meant no reset and the block came up pre-expanded, which is exactly the leak you flagged. It also could not distinguish a pass/fail split and collapsed id-less findings to indistinguishable empty strings. `InsightPayload` genuinely has no question identity, but the page rendering the panel does, so the real question id is threaded through as a prop. It is applied during render rather than in an effect, so the next question does not paint expanded for a frame and `aria-expanded` does not flip after paint.

65 tests in the panel suite, 573 across the repo, typecheck and lint clean.

jsos3-cms (approved):
**Approving — all three must-fixes from my earlier review are resolved, verified on the current head.**

1. **Info-button a11y** — fixed properly. `aria-hidden={false}` un-hides the icon, and `describeChild` wires the disclaimer as `aria-describedby` so the accessible *name* stays "About ARS alignment" and the prose is the description (no 260-char name, no orphaned tooltip text). Confirmed the mechanism against the installed MUI `Tooltip`/`SvgIcon` source, and the test now uses `getByRole('img', ...)` so it fails if this regresses.
2. **Icon contrast** — `#5c636a` ≈ 5.79:1, clears the 3:1 non-text minimum.
3. **Collapse-state leak** — fixed with the adjust-during-render idiom keyed on the real `questionid` (not a check-set signature, which aliased across questions sharing a check set — good catch on your own follow-up), with regression tests for both the cross-question reset and same-question no-collapse.

Nice-to-haves all landed (mid-drawer reorder assertions, rerender-with-changed-payload test, ARS-union failing-first assertion, chip-key nit), plus the remediation-on-passing-checks gating bug you caught while testing — right call to gate it on the render path.

The react-router 7.16→7.18 bump riding along is fine to keep here: it's the fix for the three Snyk advisories that were failing the deploy gate on every open PR, it's scoped to its own commit, and it's SPA-safe (no `useNavigation`, no future flags, server-adapter CVE surface unused) — verified with a clean prod build.

One non-blocking follow-up (not gating): the disclaimer test asserts the accessible *name* but not `toHaveAccessibleDescription(...)`, so a future drop of `describeChild` could slip past CI while today's assertions still pass. Worth a one-line hardening when convenient.

Thanks for the thorough iteration.

Issue #460, opened 2026-07-08
Render extended system fields as validated selects in the Add/Edit modal (frontend for ztmf#395)


danielbowne:
### Feature Type
UI Component

### Workflow Description
The extended system metadata fields (`fips`, `system_type`, `hva`, `cloud_system`, `cloud_service_model`, `goco_coco_gogo`, `system_operator`, `legacy`) currently render as free-text inputs in the Add/Edit System modal and the System Detail edit view (shipped as text in PR #459 / #449). Free text lets dirty values back in (`Saas` vs `SaaS`, `GOVO`, `Minor Stand-alone`), which is exactly the data-quality problem the canonicalization effort is fixing.

This is the frontend slice of CMS-Enterprise/ztmf#395. Once the backend serves the canonical allowed values per field (the reference-data pattern #392 established for `GET /api/v1/datacenterenvironments`), the modal and detail edit view should render these fields as validated select boxes instead of text inputs, so a user picks from a controlled list and the field reads consistently across every OpDiv.

Field shapes (from the FY25 inventory analysis; final set owned by #395):
- Single-select: `fips` (Low/Moderate/High), `system_type` (~7 canonical), `goco_coco_gogo` (GOCO/COCO/GOGO), `system_operator`, `cloud_system` (Yes/No), `legacy` (Yes/No), `hva`.
- Multi-select: `cloud_service_model` — the data has real combinations (IaaS/PaaS/SaaS), so this needs multi-select rather than one option per combo.

### Acceptance Criteria
- [ ] Extended fields render as backend-driven selects (values fetched from the #395 reference endpoint, mirroring the `datacenterenvironments` dropdown), not free text, in both the Add/Edit modal and the detail edit view.
- [ ] `cloud_service_model` is a multi-select; the others are single-select.
- [ ] Yes/No fields (`cloud_system`, `legacy`) render as a two-option select or toggle.
- [ ] A value not in the served set (legacy/unmapped) still displays and stays valid on an untouched save (same pattern as the datacenterenvironment legacy-value option).
- [ ] `null`/unset is handled (e.g. CMS systems before their seed) — no blank required-field lockout.
- [ ] Table columns continue to show the values.

### Additional Context
[...]

PR #626, opened 2026-07-27
Render extended metadata fields as validated selects

tarratsco:
## Summary

Part of the metadata-standardization effort (frontend of `ztmf#395`; real backend dependency is `ztmf#433`). The extended system-metadata fields (`fips`, `system_type`, `hva`, `cloud_system`, `cloud_service_model`, `goco_coco_gogo`, `system_operator`, `legacy`) render as free-text inputs, which lets dirty values back in (`Saas` vs `SaaS`, `GOVO`, `Minor Stand-alone`) — the exact data-quality problem canonicalization is fixing. This PR renders them as validated selects driven by the canonical vocabulary, in both the Add/Edit modal and the system detail edit view.

The vocabulary endpoint (`[ztmf#433](https://github.com/CMS-Enterprise/ztmf/issues/433)`) is not built yet, so the canonical set (agreed on `ztmf#395`) is served locally as a stopgap, isolated behind one module so the switch to the endpoint is a one-file change. `cloud_vendor` stays free text (out of scope on #395).

NOTE: First draft until ztmf#433 code comes in.

### `src/utils/systemMetadataVocab.ts` (new) — the swap point

* `SYSTEM_METADATA_VOCAB`: the canonical allowed values per field, hardcoded from the #395 agreed set.
* `fetchSystemMetadataVocab` / `useSystemMetadataVocab`: mirror `fetchDataCenterEnvironments`. When #433 ships, only `fetchSystemMetadataVocab` changes to a GET; every consumer is unchanged.
* `toSelectOptionsWithCurrent`: options for a single-select, appending the system's current value as a disabled option when it is not in the served set (legacy value stays visible and valid on an untouched save — same pattern as the `datacenterenvironment` dropdown).
* `parseCombo` / `serializeCombo`: convert `cloud_service_model` between its stored sorted slash-joined combo (e.g. `IaaS/PaaS`) and the multi-select parts; empty selection serializes to null.
* `multiSelectOptionsWithCurrent`: canonical parts plus any legacy part not in the canon, so an odd stored part is not dropped.

### `src/views/SystemDetailPage/fieldConfig.ts`

* Adds a `multiselect` field type. The seven canonical fields become `select`; `cloud_service_model` becomes `multiselect`. `cloud_vendor` and the owner/name fields stay free text.

### `src/views/SystemDetailPage/SystemDetailEditView.tsx` and `src/views/EditSystemModal/EditSystemModal.tsx`

[...]

PR #628, opened 2026-07-27
Render extended metadata fields as validated selects (rehome of #626)

jsos3-cms:
## Summary

In-repo rehome of #626 by @tarratsco (Onix Tarrats Calderon) so the repository CI gates (which fork PRs can't run) execute against this commit. Original authorship is preserved; the original PR #626 stays open as a draft until this merges.

Part of the metadata-standardization effort — the frontend of `ztmf#395` (backend dependency `ztmf#433`, epic `ztmf#431`). It renders the extended system-metadata fields (`fips`, `system_type`, `hva`, `cloud_system`, `cloud_service_model`, `goco_coco_gogo`, `system_operator`, `legacy`) as **validated selects** driven by a canonical vocabulary, in both the Add/Edit modal and the system-detail edit view, replacing the free-text inputs that let dirty values back in.

## Standalone-safe before the backend endpoint

The vocabulary endpoint (`ztmf#433`) isn't built yet, so the canonical set is served locally as a stopgap (`SYSTEM_METADATA_VOCAB`), isolated behind one module (`fetchSystemMetadataVocab`/`useSystemMetadataVocab`) so the switch to the real endpoint is a one-file change. A system's current (possibly legacy) value is preserved as a disabled option so existing data still displays and isn't silently changed on edit. `cloud_vendor` stays free-text (out of scope). The `cloud_system = No ⇒ cloud_service_model = null` write rule is intentionally left to `ztmf#433`.

## Provenance / verification

Commit cherry-picked verbatim from #626 (`tarratsco/ztmf-ui:tarratsco/enhancement/hhs-extended-metadata-as-selects`), authorship preserved, atop current `main` (which has advanced past the fork's base — the fork's net diff showed phantom Insights/QuestionnairePage changes that are not part of this commit; real scope is the 6 metadata-select files). `tsc --noEmit` is clean on this branch; full lint/tests verified before this is marked ready.

## Review — NO-GO as-is (opened as draft)

The vocab module, `parseCombo`/`serializeCombo`, the `#433` stopgap isolation, and `cloud_vendor` scoping are all solid. Two confirmed, reproducible integration bugs (independent of the `#433` backend work) need fixing before this is ready, both small same-file fixes:

1. **Optional selects render permanently in error state** — `SystemDetailEditView.tsx` select branch (`error`/`helperText`) is missing the `field.required &&` guard that the sibling text-field branch has, so the 7 new optional fields always show red even with a valid value. Guaranteed visual + a11y defect.
2. **Clearing a field in the Add/Edit modal doesn't persist** — `EditSystemModal.tsx` sends JS `null` on clear; the backend `blankToNil` contract treats `null` as "leave unchanged" (only `""` clears). Should send `e.target.value` (select) and `serializeCombo(parts) ?? ''` (multiselect), mirroring `SystemDetailEditView`.
3. Strip `ztmf#`/`ztmf-ui#` refs from code comments in `systemMetadataVocab.ts` and `SystemDetailEditView.test.tsx` (repo comment convention).

`tsc`/lint clean and 18/18 touched tests pass; the two bugs are behavioral and not covered by the current tests (the edit-view test used an all-valid `formValid` stand-in, which is what let Bug 1 slip).

## Depends on `ztmf#433` (tracked)

This ships the canonical vocabulary as a **local stopgap** (`SYSTEM_METADATA_VOCAB`) because the backend vocabulary endpoint is not built yet. The dependency is deliberately isolated so the switch is a one-file change:

- [ ] **When `ztmf#433` (backend vocabulary + validation endpoint) ships, swap `fetchSystemMetadataVocab` from the local `SYSTEM_METADATA_VOCAB` constant to a `GET` against that endpoint** (mirroring `fetchDataCenterEnvironments`). Every consumer already awaits the fetched result, so nothing else changes.
- [ ] Once `ztmf#433` owns write-validation, wire the `cloud_system = No ⇒ cloud_service_model = null` cross-field rule (intentionally not enforced on the frontend here).

Until `ztmf#433` lands, this PR is standalone-safe: the stopgap resolves the canonical set locally with no call to the not-yet-built endpoint.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -94,6 +94,19 @@ export type DataCenterEnvironment = {
   ordr: number
 }
 
+// One allowed value for an extended-metadata field, from
+// GET /api/v1/systemattributes. `field` is the column ("fips", "system_type",
+// "cloud_service_model", ...); `value` is a canonical option or the label for
+// a boolean field. `selectable` false hides a row from the dropdown (e.g. the
+// help row carrying only a `description`). Rows arrive ordered by `ordr`.
+export type SystemAttribute = {
+  field: string
+  value: string
+  description: string | null
+  selectable: boolean
+  ordr: number
+}
+
 export type userData = {
   userid: string
   email: string
@@ -132,19 +145,21 @@ export type FismaSystemType = {
   reactivated_date: string | null
   reactivation_notes: string | null
   opdiv_id?: number | null
-  // Extended metadata fields (migration 0044+)
+  // Extended metadata fields, typed and canonicalized by the backend.
+  // Enum strings carry a canonical value or null; the tri-state booleans are
+  // true / false / null = Unknown; cloud_service_model is a decomposed list.
   isso_name?: string | null
-  hva?: string | null
+  hva?: boolean | null
   fips?: string | null
   system_type?: string | null
-  cloud_system?: string | null
-  cloud_service_model?: string | null
+  cloud_system?: boolean | null
+  cloud_service_model?: string[] | null
   cloud_vendor?: string | null
   system_operator?: string | null
   goco_coco_gogo?: string | null
   system_owner?: string | null
   system_owner_email?: string | null
-  legacy?: string | null
+  legacy?: boolean | null
   // Risk-based target maturity (ztmf#398). null = no target asserted yet;
   // the UI presents the Advanced default.
   target_maturity_tier?: string | null
--- a/src/types.ts
+++ b/src/types.ts
@@ -96,13 +96,12 @@ export type DataCenterEnvironment = {
 
 // One allowed value for an extended-metadata field, from
 // GET /api/v1/systemattributes. `field` is the column ("fips", "system_type",
-// "cloud_service_model", ...); `value` is a canonical option or the label for
-// a boolean field. `selectable` false hides a row from the dropdown (e.g. the
-// help row carrying only a `description`). Rows arrive ordered by `ordr`.
+// "cloud_service_model", ...); `value` is a canonical option. `selectable` false
+// hides a row from the dropdown. Rows arrive ordered by `ordr`. The endpoint is
+// a plain allowed-values list; field help copy lives in the UI, not here.
 export type SystemAttribute = {
   field: string
   value: string
-  description: string | null
   selectable: boolean
   ordr: number
 }
--- a/src/utils/systemMetadataVocab.test.ts
+++ b/src/utils/systemMetadataVocab.test.ts
@@ -1,0 +1,102 @@
+import {
+  fetchSystemMetadataVocab,
+  SYSTEM_METADATA_VOCAB,
+  toSelectOptionsWithCurrent,
+  parseCombo,
+  serializeCombo,
+  multiSelectOptionsWithCurrent,
+} from './systemMetadataVocab'
+
+describe('fetchSystemMetadataVocab', () => {
+  test('resolves the canonical vocabulary (stopgap until the endpoint)', async () => {
+    const vocab = await fetchSystemMetadataVocab()
+    expect(vocab).toBe(SYSTEM_METADATA_VOCAB)
+    expect(vocab.fips).toEqual(['High', 'Moderate', 'Low'])
+    expect(vocab.goco_coco_gogo).toEqual(['GOCO', 'COCO', 'GOGO'])
+    expect(vocab.cloud_service_model).toEqual(['IaaS', 'PaaS', 'SaaS', 'Other'])
+  })
+})
+
+describe('toSelectOptionsWithCurrent', () => {
+  const allowed = ['High', 'Moderate', 'Low']
+
+  test('maps allowed values to value/label options', () => {
+    expect(toSelectOptionsWithCurrent(allowed, null)).toEqual([
+      { value: 'High', label: 'High' },
+      { value: 'Moderate', label: 'Moderate' },
+      { value: 'Low', label: 'Low' },
+    ])
+  })
+
+  test('appends a legacy value not in the set as a disabled option', () => {
+    const opts = toSelectOptionsWithCurrent(allowed, 'Undefined')
+    expect(opts).toHaveLength(4)
+    expect(opts[3]).toEqual({
+      value: 'Undefined',
+      label: 'Undefined',
+      disabled: true,
+    })
+  })
+
+  test('does not duplicate a current value already in the set', () => {
+    expect(toSelectOptionsWithCurrent(allowed, 'Low')).toHaveLength(3)
+  })
+
+  test('adds nothing for a null/empty current value', () => {
+    expect(toSelectOptionsWithCurrent(allowed, null)).toHaveLength(3)
+    expect(toSelectOptionsWithCurrent(allowed, '')).toHaveLength(3)
+    expect(toSelectOptionsWithCurrent(allowed, undefined)).toHaveLength(3)
+  })
+})
+
+describe('parseCombo', () => {
+  test('splits a slash combo into trimmed parts', () => {
+    expect(parseCombo('IaaS/PaaS')).toEqual(['IaaS', 'PaaS'])
+  })
+
+  test('tolerates legacy comma/semicolon delimiters and whitespace', () => {
+    expect(parseCombo('IaaS, PaaS')).toEqual(['IaaS', 'PaaS'])
+    expect(parseCombo('IaaS;PaaS ; SaaS')).toEqual(['IaaS', 'PaaS', 'SaaS'])
+  })
+
+  test('returns an empty array for null/empty', () => {
+    expect(parseCombo(null)).toEqual([])
+    expect(parseCombo('')).toEqual([])
+    expect(parseCombo(undefined)).toEqual([])
+  })
+})
+
+describe('serializeCombo', () => {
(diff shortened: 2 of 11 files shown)


jsos3-cms:
**Review verdict: GO** ✅

This rehome rebuilds on @tarratsco's current five commits from #626 (authorship preserved, committer-signed so it runs our repository CI gates, which fork PRs can't). I reviewed it against the retyped backend contract (ztmf#490, combining #488 + #489) three ways — an independent reference implementation, the MUI v5 select docs, and an empirical codex pass tracing the save/load paths against the actual Go wire types — plus a cross-check against the earlier #626 review and the #490 end-to-end findings. All of them converge on GO with no blocking issues.

**Contract correctness (the load-bearing part).** The four retyped fields are sent and received at their proper JSON types — `hva`/`cloud_system`/`legacy` as `boolean | null`, `cloud_service_model` as `string[] | null` — an exact match to `backend/internal/model/fismasystems.go`. The subtle part is that the clear semantics differ per field, and this version gets both right:

- Tri-state booleans use raw-JSON presence tracking on the backend, so an explicit `null` clears them → the FE sends `null` for "Unknown" (`selectValueToBool('') → null`).
- `cloud_service_model` has no presence tracking (a Go slice can't distinguish an omitted key from JSON `null` — both decode to `nil`), so the backend rule is "`nil` = leave unchanged, empty `[]` = clear." The FE sends `[]` on deselect-all, never `null` (which would silently no-op). `buildExtendedDiff` uses nullish-coalescing (`?? null`), so `false` and `[]` pass through untouched.

Net effect: no metadata PUT should 400 on a type mismatch, and no clear should silently no-op.

**Earlier review items — all resolved.** Bug A (optional selects rendering a permanent error state) is fixed via `error={field.required && !formValid[field.key]}`, with a regression test. Bug B (multiselect clear) is handled by the `[]` clear signal above, also tested. Note my original Bug B guidance ("send `''` on clear") was written under the pre-retype string contract and is now superseded — sending `''` for a boolean field would type-mismatch under the new strict decode; the per-field `null` / `[]` / `''` handling here is the correct post-retype behavior. The #490 end-to-end findings (string save → 400, `.split()`-on-array crash on load) are both gone — the vocab layer was rewritten around arrays and booleans rather than patched.

**Beyond the reference.** This version also fetches the real `GET /api/v1/systemattributes` vocab endpoint (shape-verified exact match to the BE struct) instead of a hardcoded stopgap; sends only changed fields via `buildExtendedDiff` (which avoids a stale-overwrite footgun on concurrent edits); implements the cloud_system=No → hide-and-clear cloud fields cross-field rule (mirrors a real backend validation and closes the gap flagged on #490); and correctly treats a bare `[]` as empty for the read-view card.

**Checks:** `tsc` clean, `lint:js` 0 errors (2 pre-existing InsightsPanel warnings, unrelated), jest metadata suites green (43/43).

⚠️ **Coordinated-merge gate (for whoever merges).** This is a breaking cross-repo API contract change, and `main` auto-deploys prod on merge. This PR and ztmf#490 must land together — neither prod deploy is safe standalone (the old prod frontend 400s against the new backend; a new prod frontend 400s against the old backend). Please coordinate the two merges and the rollout window rather than merging either alone.

@MackOverflow @danielbowne — tagging you both for review. @danielbowne, one open question carried over from #490: the `legacy` field's help-text wording — can you confirm the intended copy so we finalize it here?

danielbowne:
@jsos3-cms confirming the `legacy` copy.

None of the three readings are right, and that's exactly why the field needs fixing. `legacy` is a **funding disposition**: HHS marking that it does not intend to spend money remediating that system. It is not a statement about the system's age or technology, it is not a decommissioning signal (decommissioning has its own columns: `decommissioned`, `decommissioned_date`, `decommissioned_by`), and it is not CFACTS-defined - it is self-reported by HHS, which is correct here, since HHS is the only authority on its own funding intent. A FISMA system is an authorization boundary, so technical-age questions aren't answerable at that grain anyway.

**The label is the actual bug, not the missing tooltip.** "Legacy" reads as "old technology" to every user who sees it, and no amount of hover text undoes a wrong label. So the primary fix is the rename:

- Label: **Not Funded for Remediation**

**Help text**, if we're adding the prop anyway:

> HHS has not committed funding to remediate known gaps on this system. This is a funding disposition, not a judgement about the system's age or technology.

On the `FieldConfig` change itself: adding an optional `helpText?: string` to the interface is worth doing, but scope it as a generic affordance rather than a `legacy` special case - `hva`, `goco_coco_gogo`, and `system_operator` are all terms of art that will want the same treatment. Optional prop, render only when present, no backend dependency.

That last part is a correction to my own earlier direction, flagging it so it doesn't get built the old way: on ui#460 (7/15) I said tooltip text should come from the backend vocabulary endpoint as a per-field description. I withdrew that on 7/27. The vocab endpoint stays a plain allowed-values list with no description column, which keeps it a straight reuse of the `datacenterenvironments` pattern instead of a new payload shape. Help text lives in the UI.

**Tri-state still applies and matters most on this field.** Yes / No / Unknown, with `null` = "no disposition recorded". 47% of rows are null today (518/1114 in the HHS staged data), so anything that collapses null into No is asserting "HHS will fund fixing this" on half the inventory - the opposite of unknown. Don't coerce.

One knock-on for the filter work: the dashboard filter/sort on this field is still worth building, but the framing changes. It isn't "show me the oldest systems," it's "show me systems with known gaps that nobody is currently paying to close." That's the stronger prioritization signal, and it's what the column should read as in the table header too, not "Legacy."

tarratsco:
@danielbowne Im changing the label to: Not Funded for Remediation
But Im also changing the labels of the options to avoid confusing double negative so it would look like this:

| value (wire)  | current label | proposed label |
| :------- | :------- | :------- |
| 'true' (legacy = true)  | Yes    |   Not funded   |   
| 'false' (legacy = false)  | No    |   Funded   |   
| '' (null)  | Unknown    |   Unknown  |   

Let me know if you'd prefer something else for any of the labels.

MackOverflow (dismissed):
Looks good Jono, two nits: 

Potential bug: clearing a free-text field in the Add/Edit modal doesn't persist
This is the same class of bug the (stale) description claims was fixed — but it survives in the free-text branch of the modal. In EditSystemModal.tsx, the default (text) control:

onChange={(e) => setField(field.key, e.target.value || null)}
sends null when the field is emptied. Per the backend blankToNil contract documented throughout this PR, null = "leave unchanged" and only '' clears. So clearing cloud_vendor, system_owner, or system_owner_email (type email falls through to this same branch — there's no email case) and hitting Save silently drops the clear: the value persists unchanged, no error.

The sibling edit view gets this right — SystemDetailEditView.tsx default text branch sends e.target.value (i.e. '' on clear):

} else {
  props.onFieldChange(field.key, e.target.value)   // '' → blankToNil clears
}
Fix: in the modal's text branch, send e.target.value rather than e.target.value || null, matching the edit view. buildExtendedDiff already normalizes an untouched unset field, so this won't spuriously send '' for fields the user never touched.

This is also untested — the modal tests cover enum clear (fips → '') and dirty-diff omission, but not free-text clear, which is exactly why it slipped. Add a "clearing Cloud Vendor persists an empty string" case.

🟡 Risk: removal of legacy-value preservation
optionsForField and its comment assert "off-canon values can no longer be stored (the backend CHECK-constrains these columns), so there is no legacy-value preservation to do." The previous approach kept a system's current value as a disabled option; this version drops it and relies on that backend guarantee.

If any pre-existing/migrated row holds a value outside the canonical set (e.g. HHS-imported data landing before the CHECK constraint + backfill), the edit select's value={current || ''} becomes an out-of-range MUI value → console warning and a blank display, with the real value invisible to the editor. Please confirm with the backend team that the systemattributes migration has cleaned/backfilled all existing rows before this merges. If that guarantee isn't firm yet, consider retaining the disabled-current-option fallback for the transition.

costacalvin:
Closes #460. Added for tracking purposes.

jsos3-cms:
Pulled @tarratsco's two follow-ups from #626 into the rehome (rebased on latest `main`, authorship preserved, committer-signed):

- `9da78b5` — @MackOverflow's free-text-clear fix. The modal's default text branch now sends `e.target.value` (so `''` clears via `blankToNil`) instead of `e.target.value || null`, so clearing `cloud_vendor` / `system_owner` / `system_owner_email` persists instead of silently no-op'ing. Includes the "clearing Cloud Vendor persists an empty string" test you asked for.
- `ce9487a` — the `legacy` reframe per @danielbowne: field label → **Not Funded for Remediation**, option labels Not funded / Funded / Unknown (wire values unchanged — `true` / `false` / `null`), plus a generic `helpText?` on `FieldConfig` (not a legacy special-case, so hva/goco/system_operator can reuse it).

Re-verified locally: `tsc` clean, `lint:js` 0 errors (2 pre-existing InsightsPanel warnings), jest 7 suites / 88 passing (the two new tests included).

@MackOverflow — these commits supersede the version you approved, so the approval is stale; when you get a moment, could you re-confirm? Your free-text-clear nit is addressed with the exact fix + a test. The legacy-value-preservation risk you raised also holds: @danielbowne's #490 dev pre-flight found zero off-canon rows and the `#433` CHECK enforces the canonical set going forward, so there's no off-canon value for the select to drop. The coordinated-merge with ztmf#490 still stands.

MackOverflow (approved):
lgtm

PR #630, opened 2026-07-28
Add ISSO maturity-determination disclaimer to the Insights panel

danielbowne:
Closes CMS-Enterprise/ztmf-insights#40

The Insights panel presents automated findings without framing what they are for, which leaves room for an ISSO to read an automated result as an authoritative maturity determination. This adds a persistent note stating that the findings are decision-support and that the final determination belongs to the ISSO.

## What changed

- A `role="note"` callout at the bottom of the Insights panel, rendered outside the details `Collapse` so it stays visible whether or not the drawer is open, and shows even on minimal payloads that have no details toggle at all.
- The copy is the wording suggested in the issue, verbatim: "These automated findings are intended to give ISSOs better tools for assessing their system's maturity. They are decision-support only — the final maturity determination is the ISSO's responsibility." It is exported as `ISSO_DETERMINATION_DISCLAIMER` so the tests assert against the same string the panel renders.
- The recommendation badge next to each answer option carries the same framing in its hover and accessible name. That badge sits outside the panel, where the note is not visible, so it needs to say it on its own. It names the ISSO rather than using second person, since the badge renders for every role that can view the questionnaire.
image

## Styling and accessibility

Separated by a divider, small text, subdued grey, with the info icon marked decorative so screen readers do not read the same sentence twice. Foreground and background are an explicit pair rather than inherited text, so the note holds 6.6:1 contrast (AA for small text is 4.5:1) regardless of the surface it sits on, and it is set off by a border rather than color alone so it still reads as a callout under forced-colors and high-contrast modes.

One note on the acceptance criteria: the app has a single light theme (`theme.ts` hardcodes `mode: 'light'`), so there is no dark theme to check against. The explicit color pair is what covers that criterion in practice.

## Tests

Seven new cases: the note renders on a minimal payload with no details toggle, both required statements are present, it stays visible while the drawer is open, the icon is not exposed as a named element, it renders once per panel, and the badge carries the framing in both its hover and its accessible name. Full suite passes at 611.

## Review note

The disclaimer copy is the issue's suggested wording, unchanged. The issue asks for the copy to be reviewed for accuracy before merge, so a confirmation on the wording is the main thing needed here.

### DIFF
--- a/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
+++ b/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
@@ -658,10 +658,15 @@ describe('OptionInsightBadges', () => {
     // carries its own determination framing rather than relying on the panel's.
     render(<OptionInsightBadges score={1} insight={insight} />)
     const badge = screen.getByLabelText(/^ZTMF Insights —/)
-    expect(badge).toHaveAccessibleName(/final maturity determination is yours/i)
+    // Names the ISSO, not "you" — the badge renders for every role that can view
+    // the questionnaire, so second person would misattribute the determination.
+    expect(badge).toHaveAccessibleName(
+      /final maturity determination is the ISSO's/i
+    )
+    expect(badge).not.toHaveAccessibleName(/determination is yours/i)
     fireEvent.focus(badge)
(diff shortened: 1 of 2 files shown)


MackOverflow (approved):
lgtm

PR #631, opened 2026-07-29
Name the check in plain language on Insights control chip hovers

danielbowne:
An ARS control chip is labeled with the control id, so its hover is the only place that names the check behind it. It named it by slug alone:

```
IA-5
Kion · account-without-compliant-password-policy · failed
```

The payload carries a human sentence for that check (`"description": "Password Policy"`), and the check chips in the per-source blocks have shown it since they were introduced. The control chips never did, so the reader was left to decode the slug.

## What changed

- `ControlEvidence` gains an optional `description`, populated in `rollupControls` from the check's `description ?? title`. That is the same fallback `CheckTooltip` uses, so both hovers read alike: Kion puts its sentence in `description`, SecurityHub in `title`.
- The hover renders the sentence on its own indented line under the slug line. A control with several evidence entries stays scannable that way, and the slug is what a reviewer scans for first.
- A new `controlEvidenceSpoken` folds the sentence into the control chip's accessible name. Screen readers never receive tooltip node content, so without it the sentence would be sighted-only.
- Covers all three finding sources, passing and failing. CFACTS evidence is unchanged: it comes from the `ars_*` control arrays and has no per-check text to show.

The same chip now reads:

```
IA-5
Kion · account-without-compliant-password-policy · failed
    Password Policy
```

## On the regression

`ControlEvidence` was introduced in #612 with `source`, `check`, and `state`, so nothing was deleted from it. The behavior change is real all the same. Before that PR the check chips were labeled by NIST tag, so hovering a chip reading "IA-5" gave you the full sentence. After it, a chip labeled with a control id is the control chip, whose hover only had the slug. Same label, less information, as a side effect of the chip relabeling rather than a deliberate call.

## Tests

Four new cases: the sentence is carried for every finding source (Kion via `description`, SecurityHub via the `title` fallback, Hardenize), CFACTS evidence stays bare, a title-only finding does not repeat itself as both the check and the sentence, and a render test asserting the sentence appears in both the hover and the accessible name. Full suite passes at 615.

Worth flagging separately: `account-without-compliant-password-policy` maps to the sentence "Password Policy", which is a thin dictionary entry. This surfaces whatever the dictionary has. Improving the entries themselves is CMS-Enterprise/ztmf-insights#36, not this PR.

### DIFF
--- a/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
+++ b/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
@@ -927,6 +927,23 @@ describe('rollupControls (cross-source ARS control union)', () => {
     expect(evidenceFor('SC-8')?.description).toBe('HSTS not enabled')
   })
 
+  it('does not repeat a title-only finding as both the check and the sentence', () => {
+    // No id and no description, so both the check name and the sentence resolve
+    // to the same title — printing it twice in the hover (and announcing it twice
+    // to AT) reads as a bug.
+    const rolled = roll({
+      findings: {
+        sechub: [{ title: 'MFA should be enabled', nist_controls: 'IA-2' }],
+      },
+    })
+    expect(rolled[0].evidence[0]).toEqual({
+      source: 'SecurityHub',
+      check: 'MFA should be enabled',
+      state: 'failing',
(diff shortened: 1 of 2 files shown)


MackOverflow (approved):
lgtm!

PR #632, opened 2026-07-29
Distinguish inferred passes from observed ones in the Insights panel

danielbowne:
SecurityHub's vendor feed carries only FAILED records and never emits a PASSED one, so `sechub_passing` is derived upstream as "mapped control with no active failure". That cannot tell "evaluated and passed" apart from "never scanned".

This matters more than the wording, because a passing entry is not display-only. `rollupControls` counts every `{source}_passing` entry as `satisfied` evidence in the cross-source ARS control union, so a control whose only evidence is an inferred pass flips from grey not-assessed to green satisfied. On a partially scanned system the panel would be asserting a control is met on the strength of a check that never ran. The ISSO disclaimer added in #630 says the findings are decision-support; it does not say green may mean unscanned.

Relevant now rather than hypothetical: the upstream view is about to start shipping `sechub_passing` on roughly 1,500 rows, and the panel switches rendering on the array's presence alone, with no code change.

## What changed

- `INFERRED_PASS_SOURCES` names the sources whose passing entries are inferred from the absence of a finding. SecurityHub is the only member today.
- Those entries are **withheld from the control rollup** rather than counted as affirmative `satisfied` evidence.
- Everywhere a pass would be claimed now says findings were not reported instead:
  - block summary: `N of M checks reported no findings`
  - chip hover and accessible name: `No finding reported` in place of `Passed`
  - collapse toggle: `Show all N checks with no findings`
- The chips still render green, since the check is genuinely not failing. Only the claim about why changes.

Removing a source from the set is all that is required once its feed emits real passes. Nothing else needs to change.

## Tests

Five cases covering the summary wording, the chip accessible name, the withheld rollup evidence, that a real-pass source such as Kion is unaffected, and the collapse toggle wording once a block spills past the collapse threshold. Each pairs a positive assertion with a negative one so the old "passed" phrasing cannot creep back. Full suite passes at 662.

## Note on layering

The cleanest home for this is the payload: a flag distinguishing an observed pass from "no finding reported" would let the UI stop hardcoding a per-source data-quality fact. That was raised upstream. This lands the correct behavior now, and collapses to a one-line deletion if the flag arrives later.

The same question applies to the archived Kion checks currently dropped from scoring. If those start counting as passes, they want this treatment rather than a silent flip.

### DIFF
--- a/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
+++ b/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
@@ -1296,10 +1296,50 @@ describe('FeedCheckBlock (feed pass/fail checks)', () => {
       />
     )
     expand()
-    expect(screen.getByText(/1 of 2 checks passed/)).toBeInTheDocument()
+    // SecurityHub's feed is failure-only, so a "pass" is the absence of a
+    // finding, not an observed pass — the summary must not claim otherwise.
+    expect(
+      screen.getByText(/1 of 2 checks reported no findings/)
+    ).toBeInTheDocument()
+    expect(screen.queryByText(/1 of 2 checks passed/)).not.toBeInTheDocument()
     // Feed chips are labeled by the SecurityHub finding code.
     expect(screen.getByText('IAM.1').textContent).toContain('✓')
     expect(screen.getByText('IAM.10').textContent).toContain('✗')
+    // The chip still reads ✓, but neither its accessible name nor its hover may
+    // say "Passed".
+    expect(
+      screen.getByRole('img', { name: /^IAM\.1 — .* — No finding reported$/ })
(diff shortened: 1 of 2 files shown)


MackOverflow (approved):
great fix! Just noting that we have this assymetry: 

Within one panel, a SecurityHub chip stays green ✓ while the ARS control it maps to may not. The chip renders variant='satisfied' and the rollup withholds it, so IAM.1 shows green in the SecurityHub block but AC-3 may sit grey in the ARS union. The author addresses this deliberately ("the check is genuinely not failing"), and the tooltip/summer wording carries the "no finding reported" caveat — so it's defensible, just an asymmetry a careful ISSO could notice.

Not something that blocks this fix, but something to probably discuss when we have the groundover and we want to think about how to make insights even better.

PR #636, opened 2026-07-29
Correct the legacy field label to "Legacy System" (#634)

### LINKED ISSUE #634: Correct the legacy field label: it is a legacy-system flag, not a funding disposition
## Summary

The `legacy` field shipped in #628 under the wrong label. It currently renders as **Not Funded for Remediation** with options **Not funded / Funded / Unknown**. That is my miscommunication, not a build error, and it needs to be reverted.

What I got wrong: I conflated the **label** with the **hover description**. The ask on #460 was for hover text explaining these fields, and I answered it by renaming the field instead. The label should carry the field's name as HHS reports it, and any explanation of what the flag means belongs in the hover copy underneath it. Renaming the column was the wrong instrument, and the team built exactly what I asked for.

So `legacy` is what it says it is: a legacy-system flag, HHS's self-reported yes/no on whether the system is legacy technology (the framing @MackOverflow and @stu-aq raised on ztmf#395). Nothing in the HHS source data or the backend column supports reading it as a funding disposition.

## Current state

- `src/views/SystemDetailPage/fieldConfig.ts:192-203` — label `Not Funded for Remediation`, `booleanLabels: { true: 'Not funded', false: 'Funded' }`, plus `helpText` describing a funding disposition.
- `src/utils/systemMetadataVocab.ts:81-84` — the `BooleanLabels` doc comment uses this field as its worked example.
- Tests pinning the wrong copy: `src/views/SystemDetailPage/SystemDetailReadView.test.tsx:86-91`, `src/utils/systemMetadataVocab.test.ts:148-166`.

The UI also diverges from the backend vocabulary here: `systemattributes` seeds `legacy` as `Yes` / `No` (migration `0053systemattributes.go:76-77`), and the UI overrides those with Not funded / Funded.

## What it should be

- Field label: **Legacy System**
- Option labels: **Yes / No / Unknown** — drop the `booleanLabels` override and fall back to the shared default, which then matches the backend vocabulary.
- Remove the funding-disposition `helpText`. Keep the `helpText?` prop on `FieldConfig` — it stays the right home for the hover copy on this field and on `hva`, `goco_coco_gogo`, and `system_operator`.

Tri-state behavior is unchanged and still matters: `true` / `false` / `null`, with `null` meaning not recorded (about 47% of rows). Do not coerce null to No.

## Backend

No backend change needed. The column is a plain `boolean` (`0052fismasystemsretype.go`), the vocabulary rows are already `Yes` / `No`, and the API carries no display copy. Frontend copy fix only.

## Acceptance criteria

- [ ] Add/Edit modal and system-detail edit view show the field as **Legacy System** with Yes / No / Unknown.
- [ ] Read view renders `true` as Yes and `false` as No.
- [ ] Funding-disposition help text removed; `helpText?` prop retained.
- [ ] Wire values unchanged (`true` / `false` / `null`), tri-state preserved, no coercion of null.
- [ ] Tests updated to pin the corrected copy.

## Open question

What the hover text should say now. The criteria for "legacy" are still HHS self-reported with nothing attached, which is the definition ask @MackOverflow originally raised on ztmf#395. Suggest shipping the relabel on its own and settling the hover copy separately rather than blocking the fix on it.

## Knock-on

The `legacy` filter/sort work for the dashboard was rationalized on #460 as "systems nobody is paying to remediate." That rationale goes away with this correction — the work still stands, framed as legacy technology.

costacalvin:
## Summary
PR #628 shipped the `legacy` field labeled **"Not Funded for Remediation"** with options **"Not funded / Funded / Unknown"**, framing a legacy-system flag as a funding disposition. This restores its correct name and a plain **Yes/No/Unknown** interface, and drops the funding-disposition help text.

Closes #634.

## Changes
- **`fieldConfig.ts`** (only production change): the `legacy` entry
  - `label`: `"Not Funded for Remediation"` → `"Legacy System"`
  - Removed the `booleanLabels: { true: 'Not funded', false: 'Funded' }` override → falls back to default **Yes / No / Unknown**
  - Removed the funding-disposition `helpText` value (the `helpText?` / `booleanLabels?` properties remain on the `FieldConfig` interface)
- Reworded two doc comments that used the old label as their example (`fieldConfig.ts`, `systemMetadataVocab.ts`)
- **Tests:** rewrote the read-view assertion for the new copy; de-referenced the retired copy from the vocab-util fixtures (`Not funded/Funded` → neutral `Active/Inactive`)

All rendering (edit modal, detail edit view, detail read view) reads `label` / `booleanLabels` / `helpText` generically from the field config, so the config entry is the only production change; everything else adapts.

## Acceptance criteria
- [x] Modal and detail views display "Legacy System" with Yes/No/Unknown options
- [x] Read views render appropriately
- [x] Help text removed while retaining the `helpText` property on the interface
- [x] Wire values unchanged; tri-state (true/false/null) preserved
- [x] Affected tests updated
- [x] No backend modifications

### BEFORE
image

image

### AFTER
image

image

## Verification
- `jest` on all 4 affected suites: 48/48 pass
- ESLint clean on edited files; no new `tsc` errors introduced
- No remaining references to the old copy anywhere in `src/`
- On ZTMF, go to a system's details page. Click edit. Scroll down to the extended metadata section and view the Legacy System field. Verify the three options are Yes, No, Unknown.

### DIFF
--- a/src/utils/systemMetadataVocab.test.ts
+++ b/src/utils/systemMetadataVocab.test.ts
@@ -145,9 +145,9 @@ describe('booleanOptions', () => {
   })
 
   it('overrides the true/false labels but keeps values and Unknown', () => {
-    expect(booleanOptions({ true: 'Not funded', false: 'Funded' })).toEqual([
-      { value: 'true', label: 'Not funded' },
-      { value: 'false', label: 'Funded' },
+    expect(booleanOptions({ true: 'Active', false: 'Inactive' })).toEqual([
+      { value: 'true', label: 'Active' },
+      { value: 'false', label: 'Inactive' },
       { value: '', label: 'Unknown' },
     ])
   })
@@ -162,9 +162,9 @@ describe('display formatting', () => {
   })
 
   it('applies custom true/false labels, leaving Unknown', () => {
-    const labels = { true: 'Not funded', false: 'Funded' }
-    expect(formatBool(true, labels)).toBe('Not funded')
-    expect(formatBool(false, labels)).toBe('Funded')
+    const labels = { true: 'Active', false: 'Inactive' }
+    expect(formatBool(true, labels)).toBe('Active')
+    expect(formatBool(false, labels)).toBe('Inactive')
     expect(formatBool(null, labels)).toBe('Unknown')
   })
 
--- a/src/utils/systemMetadataVocab.ts
+++ b/src/utils/systemMetadataVocab.ts
@@ -80,8 +80,7 @@ export function optionsForField(
 
 // Custom labels for the true/false ends of a tri-state boolean. For fields
 // whose Yes/No reads ambiguously against the label (e.g. a negatively-phrased
-// "Not Funded for Remediation", where "No" would mean "it is funded"). Unknown
-// is always Unknown.
+// label where "No" would be a double negative). Unknown is always Unknown.
 export type BooleanLabels = { true: string; false: string }
 
 /**
--- a/src/views/SystemDetailPage/SystemDetailReadView.test.tsx
+++ b/src/views/SystemDetailPage/SystemDetailReadView.test.tsx
@@ -83,14 +83,12 @@ describe('extended metadata formatting', () => {
     expect(screen.getByText('IaaS, PaaS')).toBeInTheDocument()
   })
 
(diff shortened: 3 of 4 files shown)


MackOverflow (approved):
clean! Looks great Calvin!

PR #637, opened 2026-07-29
fix(dashboard): scope average system score to the selected data call (#633)

### LINKED ISSUE #633: Dashboard average system score is below the scale minimum when a single data call is selected
### Description

With a single data call selected, the dashboard's **Average System Score** is far too low. It reads **0.56** while the same view shows a highest of 4.82 and a lowest of 1.00. An average below the minimum of the scale is impossible, so the denominator is wrong.

The scale is 1.0 to 5.0 (the HHS-aligned shift applied at aggregation). A system can never score below 1.0.

Root cause looks like the average is computed over **all active systems** rather than the systems that actually have answers in the selected data call. Systems with no answers for that call contribute zero and drag the mean down.

Measured against the loaded dataset:

| data call | systems with answers | true average |
| --- | --- | --- |
| FY2023 Q4 | 226 | 3.26 |
| FY2025 Q3 | 238 | 3.44 |
| FY23 (HHS) | 1,201 | 3.27 |
| FY24 (HHS) | 1,216 | 3.29 |
| FY25 (HHS) | 957 | 3.46 |

Active systems in total: 1,342.

For FY2025 Q3 the dashboard shows 0.56. Dividing the summed per-system averages by 1,342 instead of 238 produces roughly that, which is consistent with the denominator being the full system count.

Related: the **Total Systems** tile also shows 1,342 with a single data call selected, i.e. it is not scoped to the selection either. Those two may share a cause.

This became visible after loading several years of historical data, because previously most systems had answers in the only call that mattered. It is a pre-existing calculation issue that the wider dataset exposed, not a regression from the load.

### Steps to Reproduce

1. Open the dashboard.
2. Filter to a single data call that only some systems participated in.
3. Compare the Average System Score tile against the highest and lowest tiles.

### Expected Behavior

The average is computed only over systems that have answers in the selected data call. With that call selected the tile should read approximately 3.44, between the displayed lowest and highest. The Total Systems tile should reflect the systems in scope for the selection.

### Actual Behavior

Average reads 0.56, below the minimum possible score of 1.0 and well below the displayed lowest system score of 1.00. Total Systems shows the full active count regardless of the selected call.

### Environment

dev

### Acceptance Criteria

- [ ] Average System Score is calculated over systems with answers in the selected data call
- [ ] The average always falls between the displayed lowest and highest system scores
- [ ] Total Systems reflects the current selection, or is labelled so it is clear it does not
- [ ] Coverage for a data call in which only a subset of systems participated

### Additional Context

Worth checking whether the fix belongs in the frontend or in the aggregate scores endpoint that feeds these tiles, since the denominator may be chosen server side.

costacalvin:
## Summary
Closes #633

The dashboard's **Average System Score** tile read far below the scale minimum (e.g. `0.56`) when a single data call was selected — impossible given the 1.0–5.0 scale and the displayed lowest score of `1.00`.

Root cause: the average summed the per-system scores of only the systems that have answers in the selected data call, but divided by the count of **all active systems** (~1,342). Systems with no answers in the call contributed nothing to the numerator yet still inflated the denominator, dragging the mean far below the true value (~3.44 for FY2025 Q3).

This is a frontend-only fix. The backend `/scores/aggregate?datacallid=` endpoint already returns only systems with answers in the selected call, and `buildDashboardMaps` already builds a selection-scoped score map — the dilution was purely in the tile calculation.

## Changes

- **Average denominator** now counts only systems that actually contribute a score, computed in the same loop as the numerator so both always cover the same set. This guarantees the average falls between the displayed lowest and highest.
- **Total Systems tile** is now a **`Scored / Total Systems`** ratio (e.g. `238 / 1,342`), making the selection-scoped count explicit alongside the active-system total.
- Adds a `StatisticsBlocks` test suite (previously untested) covering subset participation, the average-between-bounds invariant, the empty-selection edge case, and thousands-separator formatting.

## Acceptance criteria

- [x] Average System Score is calculated over systems with answers in the selected data call
- [x] The average always falls between the displayed lowest and highest system scores
- [x] Total Systems reflects the current selection (shown as `Scored / Total`, clearly labelled)
- [x] Coverage for a data call in which only a subset of systems participated

### BEFORE
image

### AFTER
image

## Testing

- New `StatisticsBlocks.test.tsx` — 5 tests, all passing
- Full suite green (667 passed), lint + type-check clean
- Verify in UI, you see the scored / total systems tile.

### DIFF
--- a/src/views/StatisticBlocks/StatisticsBlocks.test.tsx
+++ b/src/views/StatisticBlocks/StatisticsBlocks.test.tsx
@@ -1,0 +1,103 @@
+import { render, screen } from '@testing-library/react'
+import StatisticsBlocks from './StatisticsBlocks'
+import type { FismaSystemType, SystemScoreEntry } from '@/types'
+
+// StatisticsBlocks reads the full active-system list from the outlet context and
+// the selection-scoped score map from props. The bug in ztmf-ui#633 was the
+// average dividing by the whole context list; these tests pin the denominator to
+// the systems that actually have a score.
+let mockFismaSystems: FismaSystemType[] = []
+jest.mock('../Title/Context', () => ({
+  __esModule: true,
+  useContextProp: () => ({ fismaSystems: mockFismaSystems }),
+}))
+
+const sys = (id: number, acronym: string): FismaSystemType =>
+  ({ fismasystemid: id, fismaacronym: acronym }) as FismaSystemType
+
+const score = (
+  n: number,
+  tier?: SystemScoreEntry['tier']
+): SystemScoreEntry => ({
+  score: n,
+  tier,
+})
+
+// Read the big numeral out of a tile located by its label. The high/low tiles
+// fold an acronym into the label element, so callers pass a regex for those.
+const tileValue = (label: string | RegExp): string => {
+  const tile = screen.getByText(label).closest('.MuiPaper-root')
+  const numeral = tile?.querySelector('h2')
+  return (numeral?.textContent ?? '').replace(/\s+/g, ' ').trim()
+}
+
+describe('StatisticsBlocks — ztmf-ui#633 selection-scoped scoring', () => {
+  afterEach(() => {
+    mockFismaSystems = []
+  })
+
+  it('averages only over systems with a score, not the full active list', () => {
+    // Three active systems, only two answered the selected call. The average must
+    // be (2 + 5) / 2 = 3.5, NOT (2 + 5) / 3 = 2.33 (the old diluted denominator).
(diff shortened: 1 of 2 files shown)


MackOverflow (approved):
LGTM! 

Note for the future: Formatting inconsistency across tiles. The Average tile renders {avgSystemScore} raw (e.g. 3.5), while Highest/Lowest use .toFixed(2) (e.g. 3.50). So a dashboard row can show 3.5 next to 4.82 and 1.50. This is pre-existing (the average was always rendered raw), so it's out of scope — but if we want to address in the future, normalizing the average to .toFixed(2) would make the four tiles visually consistent. If you do, update the test assertions ('3.5' → '3.50', '0' → '0.00').

PR #638, opened 2026-07-29
feat(#609, #610): cross-navigation between System Info and the Questionnaire

### LINKED ISSUE #609: Button from System Info to Questionnaire
### Feature Type

UI Component

### Workflow Description

There should be a button or link on the System Info page to take you to the system's Questionnaire.

### Acceptance Criteria

- [ ] A Questionnaire button exists
- [ ] It links to the Questionnaire page/first element

### Additional Context

Sometimes when I am on the System Info page, I would like to go to the Questionnaire.  Currently I have to go back to the dashboard and find the system again.

### LINKED ISSUE #610: Button from Questionnaire to System Info
### Feature Type

UI Component

### Workflow Description

There should be a button or link on the  Questionnaire System Info page to take you to the System Info page.

### Acceptance Criteria

- [ ] A System Info button exists
- [ ] It links to the System Info page
- [ ] Maybe add in a button for the Pillar Scores Modal, too?

### Additional Context

Sometimes when I am on the Questionnaire page, I would like to go to the System Info page.  Currently I have to go back to the dashboard and find the system again.

voidspooks:
Closes #609
Closes #610

Both tickets are the same complaint from opposite directions: the System Info page and the Questionnaire were only reachable from the dashboard, so moving between a system's detail page and its questionnaire meant going back and finding the system again.

The routes are keyed on different identifiers, questionnaire on `:fismaacronym` and detail on `:fismasystemid`, and each page already holds both, so neither direction needs a new lookup to build its link.

## #609, System Info to Questionnaire

01-system-info-questionnaire-button

`SystemDetailHeader` takes the acronym as a prop and renders a Questionnaire button to the left of Edit, targeting `/questionnaire/<acronym-lowercased>`, the same URL shape the dashboard's own questionnaire action builds in `FismaTable`.

It deliberately carries no route state. An absent `location.state.datacallid` is what lets `QuestionnairePage` resolve the cycle itself, which is the right default arriving from System Info with no per-row call context, and it keeps the target a plain shareable URL.

A decommissioned system links too rather than hiding the button, leaving the questionnaire's existing "no questionnaire is available for this system" alert to explain the outcome. The button is suppressed while the page is in edit mode, so a dirty form keeps Save and Cancel as its only actions, matching how that button group already behaves.

## #610, Questionnaire to System Info, plus Pillar Scores

Screenshot 2026-07-29 at 10 36 22 AM

The questionnaire header row already held Compare Datacalls next to the breadcrumbs, so System Info and Pillar Scores join it there.

System Info navigates to `/systems/<fismasystemid>`, resolved from the acronym the route carries. It sits behind the same `hasSystemAccess` check the dashboard puts on its own System Details action. Worth being precise about what that gate does today: every role in the union passes it, delegates included, so in practice it only suppresses the button while `userInfo` is unloaded or carries an unrecognized role. I kept it so the two move together if system-detail access ever narrows.

Pillar Scores reuses `PillarScoresModal`. Unlike `ScoreDiffModal` it takes its rows as a prop rather than fetching, so the handler makes the same `GET /scores/aggregate?fismasystemid=…&include_pillars=true` call the dashboard action makes and opens on success. A failure notifies rather than opening an empty modal, which would read as "this system has no scores". I did not port the dashboard's module-level `pillarScoresCache`: that memoizes across many rows, this is one system and one click.

The modal title takes its acronym from the resolved system rather than the lowercased URL param, so the casing matches the dashboard.

Small cleanup while I was in there: the selected-data-call fallback chain is now a single `viewedDataCallId` const, since both header modals need the same value.

## No unsaved-work guard, on purpose

Leaving the questionnaire for System Info does not prompt. In-progress answers are persisted through `saveDraft` / `loadDraft` keyed on user, system, question and data call, so navigating away and back restores the draft, and the existing Compare Datacalls button sets no guard either.

## Testing

Ten new tests, six in `SystemDetailHeader.test.tsx` (a new file, that component had no coverage) and four in `QuestionnairePage.crossnav.test.tsx`. Both use the `createMemoryRouter` data-router harness the deep-link suite documents, so the components get the same stable `useNavigate` the app does and the resulting pathname is assertable.

Coverage is the lowercased acronym target, the deliberately null route state, the decommissioned case still linking, edit-mode suppression, the acronym resolving to the right `fismasystemid`, the access gate, the aggregate call firing with the right params before the modal opens, and the failure path leaving the modal closed.

Full frontend gate green locally: `npx tsc --noEmit` clean, `eslint src` zero errors, and the whole jest suite at 52 suites and 677 tests passing. No backend change, so nothing for the Go or emberfall suites.

Snyk and the other secret-dependent checks will show red on this fork PR as usual.

### DIFF
--- a/src/views/QuestionnairePage/QuestionnairePage.crossnav.test.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.crossnav.test.tsx
@@ -1,0 +1,216 @@
+import { render, screen, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { createMemoryRouter, RouterProvider } from 'react-router-dom'
+import { Routes as AppRoutes } from '@/router/constants'
+import QuestionnairePage from './QuestionnairePage'
+import type { userData } from '@/types'
+
+// Cross-navigation coverage for ui#610: the questionnaire header needs a link
+// back to System Info and a Pillar Scores button, alongside the existing
+// Compare Datacalls button.
+
+jest.mock('@/utils/config', () => ({
+  __esModule: true,
+  default: { INSIGHTS_SUGGEST_FIX_ENABLED: false },
+}))
+
+jest.mock('@/axiosConfig', () => ({
+  __esModule: true,
+  default: { get: jest.fn(), post: jest.fn(), put: jest.fn() },
+}))
+const mockGet = require('@/axiosConfig').default.get as jest.Mock
+
+// draftStore uses crypto.subtle (unavailable in jsdom); stub to no-ops.
+jest.mock('./draftStore', () => ({
+  saveDraft: jest.fn().mockResolvedValue(true),
+  loadDraft: jest.fn().mockResolvedValue(null),
+  clearDraft: jest.fn().mockResolvedValue(undefined),
+}))
+
+// notify() reaches notistack's standalone enqueueSnackbar, which needs a
+// mounted provider; keep isAuthHandled real so the error path is exercised as
+// written (same pattern as QuestionnairePage.test).
+const notifyMock = jest.fn()
+jest.mock('@/utils/notify', () => {
+  const actual = jest.requireActual('@/utils/notify')
+  return {
+    ...actual,
+    notify: (...args: unknown[]) => notifyMock(...args),
+  }
+})
+
+let mockCtx: Record<string, unknown>
+jest.mock('../Title/Context', () => ({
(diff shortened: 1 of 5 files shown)


MackOverflow:
Moving this to #640 — fork PRs can't access the org secrets required for Snyk and the deploy CI gates, so those jobs can't run here. Your commits and authorship carry over unchanged. Thanks @voidspooks!

PR #640, opened 2026-07-29
feat(#609, #610): cross-navigation between System Info and the Questionnaire

### LINKED ISSUE #609: Button from System Info to Questionnaire

### LINKED ISSUE #610: Button from Questionnaire to System Info

MackOverflow:
> **Note:** In-repo copy of #638 by @voidspooks. Moved from the fork so Snyk and the deploy CI gates can run — those jobs require org secrets unavailable to fork PRs. Commits and authorship are unchanged.
>
> Refs #638

---

Closes #609
Closes #610

Both tickets are the same complaint from opposite directions: the System Info page and the Questionnaire were only reachable from the dashboard, so moving between a system's detail page and its questionnaire meant going back and finding the system again.

The routes are keyed on different identifiers, questionnaire on `:fismaacronym` and detail on `:fismasystemid`, and each page already holds both, so neither direction needs a new lookup to build its link.

## #609, System Info to Questionnaire

01-system-info-questionnaire-button

`SystemDetailHeader` takes the acronym as a prop and renders a Questionnaire button to the left of Edit, targeting `/questionnaire/<acronym-lowercased>`, the same URL shape the dashboard's own questionnaire action builds in `FismaTable`.

It deliberately carries no route state. An absent `location.state.datacallid` is what lets `QuestionnairePage` resolve the cycle itself, which is the right default arriving from System Info with no per-row call context, and it keeps the target a plain shareable URL.

A decommissioned system links too rather than hiding the button, leaving the questionnaire's existing "no questionnaire is available for this system" alert to explain the outcome. The button is suppressed while the page is in edit mode, so a dirty form keeps Save and Cancel as its only actions, matching how that button group already behaves.

## #610, Questionnaire to System Info, plus Pillar Scores

Screenshot 2026-07-29 at 10 36 22 AM

The questionnaire header row already held Compare Datacalls next to the breadcrumbs, so System Info and Pillar Scores join it there.

System Info navigates to `/systems/<fismasystemid>`, resolved from the acronym the route carries. It sits behind the same `hasSystemAccess` check the dashboard puts on its own System Details action. Worth being precise about what that gate does today: every role in the union passes it, delegates included, so in practice it only suppresses the button while `userInfo` is unloaded or carries an unrecognized role. I kept it so the two move together if system-detail access ever narrows.

Pillar Scores reuses `PillarScoresModal`. Unlike `ScoreDiffModal` it takes its rows as a prop rather than fetching, so the handler makes the same `GET /scores/aggregate?fismasystemid=…&include_pillars=true` call the dashboard action makes and opens on success. A failure notifies rather than opening an empty modal, which would read as "this system has no scores". I did not port the dashboard's module-level `pillarScoresCache`: that memoizes across many rows, this is one system and one click.

The modal title takes its acronym from the resolved system rather than the lowercased URL param, so the casing matches the dashboard.

Small cleanup while I was in there: the selected-data-call fallback chain is now a single `viewedDataCallId` const, since both header modals need the same value.

## No unsaved-work guard, on purpose

Leaving the questionnaire for System Info does not prompt. In-progress answers are persisted through `saveDraft` / `loadDraft` keyed on user, system, question and data call, so navigating away and back restores the draft, and the existing Compare Datacalls button sets no guard either.

## Testing

Ten new tests, six in `SystemDetailHeader.test.tsx` (a new file, that component had no coverage) and four in `QuestionnairePage.crossnav.test.tsx`. Both use the `createMemoryRouter` data-router harness the deep-link suite documents, so the components get the same stable `useNavigate` the app does and the resulting pathname is assertable.

Coverage is the lowercased acronym target, the deliberately null route state, the decommissioned case still linking, edit-mode suppression, the acronym resolving to the right `fismasystemid`, the access gate, the aggregate call firing with the right params before the modal opens, and the failure path leaving the modal closed.

Full frontend gate green locally: `npx tsc --noEmit` clean, `eslint src` zero errors, and the whole jest suite at 52 suites and 677 tests passing. No backend change, so nothing for the Go or emberfall suites.

Snyk and the other secret-dependent checks will show red on this fork PR as usual.

### DIFF
--- a/src/views/QuestionnairePage/QuestionnairePage.crossnav.test.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.crossnav.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen, waitFor } from '@testing-library/react'
+import { act, render, screen, waitFor } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { createMemoryRouter, RouterProvider } from 'react-router-dom'
 import { Routes as AppRoutes } from '@/router/constants'
@@ -276,4 +276,47 @@ it('flushes a pending draft when the page unmounts mid-debounce', async () => {
   })
   expect(questionId).toBe(7006)
   expect(draft.notes).toBe('Imperial posture note')
+})
+
+it('does not lose a newer edit made while an earlier save is in flight', async () => {
+  // Regression for the identity-clear race (#640 review): edit A's debounced
+  // save starts; edit B replaces the pending payload under the SAME generation
+  // (saveGenRef only moves on explicit clears); save A's completion must not
+  // clear B's payload, or an unmount before B's own debounce flushes nothing.
+  const { saveDraft } = require('./draftStore') as { saveDraft: jest.Mock }
+  let resolveSaveA!: (saved: boolean) => void
+  saveDraft.mockImplementationOnce(
+    () => new Promise<boolean>((resolve) => (resolveSaveA = resolve))
+  )
+  mockGet.mockImplementation((url: string) => {
+    if (url.includes('/questions'))
+      return Promise.resolve({ data: { data: QUESTIONS } })
+    if (url.includes('/options'))
+      return Promise.resolve({ data: { data: OPTIONS_7006 } })
+    return Promise.resolve({ data: { data: [] } })
+  })
+
+  const { unmount } = renderPage()
+  const notesField = await screen.findByLabelText('Justification notes')
+
+  // Edit A; let its 1s debounce fire so saveDraft(A) is genuinely in flight.
+  await userEvent.type(notesField, 'A')
+  await waitFor(() => expect(saveDraft).toHaveBeenCalledTimes(1), {
+    timeout: 3000,
+  })
+  expect(saveDraft.mock.calls[0][4].notes).toBe('A')
+
+  // Edit B while A remains unresolved.
+  await userEvent.type(notesField, 'B')
+
+  // A resolves successfully. Identity check must leave B's payload pending.
+  await act(async () => {
+    resolveSaveA(true)
+  })
+
+  // Leave before B's own debounce fires.
+  unmount()
+
+  expect(saveDraft).toHaveBeenCalledTimes(2)
+  expect(saveDraft.mock.calls[1][4].notes).toBe('AB')
 })
--- a/src/views/QuestionnairePage/QuestionnairePage.crossnav.test.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.crossnav.test.tsx
@@ -1,0 +1,216 @@
+import { render, screen, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { createMemoryRouter, RouterProvider } from 'react-router-dom'
+import { Routes as AppRoutes } from '@/router/constants'
+import QuestionnairePage from './QuestionnairePage'
+import type { userData } from '@/types'
+
+// Cross-navigation coverage for ui#610: the questionnaire header needs a link
+// back to System Info and a Pillar Scores button, alongside the existing
+// Compare Datacalls button.
+
+jest.mock('@/utils/config', () => ({
+  __esModule: true,
+  default: { INSIGHTS_SUGGEST_FIX_ENABLED: false },
+}))
+
+jest.mock('@/axiosConfig', () => ({
+  __esModule: true,
+  default: { get: jest.fn(), post: jest.fn(), put: jest.fn() },
+}))
+const mockGet = require('@/axiosConfig').default.get as jest.Mock
+
+// draftStore uses crypto.subtle (unavailable in jsdom); stub to no-ops.
+jest.mock('./draftStore', () => ({
+  saveDraft: jest.fn().mockResolvedValue(true),
+  loadDraft: jest.fn().mockResolvedValue(null),
+  clearDraft: jest.fn().mockResolvedValue(undefined),
+}))
+
+// notify() reaches notistack's standalone enqueueSnackbar, which needs a
+// mounted provider; keep isAuthHandled real so the error path is exercised as
+// written (same pattern as QuestionnairePage.test).
+const notifyMock = jest.fn()
+jest.mock('@/utils/notify', () => {
+  const actual = jest.requireActual('@/utils/notify')
+  return {
+    ...actual,
+    notify: (...args: unknown[]) => notifyMock(...args),
+  }
+})
+
+let mockCtx: Record<string, unknown>
+jest.mock('../Title/Context', () => ({
+  useContextProp: () => mockCtx,
+}))
+
+const QUESTIONS = [
+  {
+    questionid: 900,
+    question: 'Question for Imperial Identity Verification',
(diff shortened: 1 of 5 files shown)


costacalvin:
## Review summary — request changes

Solid, well-scoped PR: all ACs for #609/#610 are met, the 10 new tests pass, lint is clean, the `viewedDataCallId` refactor is behavior-preserving, and the Pillar Scores call/casing/gating all match the dashboard. Verified working as expected on the dev/local environment. No security concerns. One control-flow issue to address first:

### 🔴 Blocker — draft-save race on System Info navigation

The draft save is debounced 1s (`QuestionnairePage.tsx:1031`) and the effect cleanup cancels the pending timer on unmount (`QuestionnairePage.tsx:1049`). The new System Info button is the first in-app navigation that unmounts the questionnaire (existing Next/Back/sidebar `navigate()` calls stay on the route), and `beforeunload` doesn't fire on SPA navigation. So editing an answer and clicking System Info within ~1s loses that edit — which contradicts the PR's "navigating away and back restores the draft" rationale. Narrow/recoverable, but newly introduced. Suggest flushing the draft synchronously in the handler before `navigate()`.

### 🟡 Non-blocking / Follow up suggestions

- **No way back from the `noQuestions` state.** The alert returns before the action header (`QuestionnairePage.tsx:1198` vs `QuestionnairePage.tsx:1229`), so the intended decommissioned/out-of-scope flow (System Info → Questionnaire → "no questionnaire available") is one-way. `system` is resolved there, so rendering the link is cheap. Prefer adding it here with a test; otherwise a follow-up is fine.
- **`res.data?.data ?? []`** in `handleOpenPillarScores` — tiny, safe guard so the modal never receives `undefined` on an incomplete envelope. Worth doing while touching the handler.
- **Router `Link` vs `onClick + navigate`** — semantically better for page-to-page nav (open-in-new-tab, copy link). Recommend it for the System Info / Questionnaire buttons, but fine as a separate cleanup if the CMS button composition is awkward.
- **Redundant decommissioned test** — the "still links a decommissioned system" case asserts the same thing as the lowercased-acronym test. Prefer converting it into real coverage of the `noQuestions` return link (see above) rather than keeping it as-is.

@voidspooks @MackOverflow

voidspooks:
@MackOverflow could you pull one more commit into this branch? Calvins review fixes are on my fork at `voidspooks/ztmf-ui:feat/609-610-cross-nav`, commit `89729e5`. This PR head is in the upstream repo where I only have read, so I cannot push it here myself.

Full frontend gate is green on it locally: tsc clean, eslint no errors, and the whole jest suite at 52 suites and 677 tests passing. I will write up what changed once it is in.

MackOverflow:
working on this @voidspooks

MackOverflow:
@voidspooks updated added, ready for review @costacalvin

costacalvin:
### 🔴 Required before merge: newer draft edit can be lost during an in-flight save @voidspooks 
The PR fixes revealed a pre-existing data loss bug that was previously undiscovered. Recommend fixing in this PR before merging, even though it wasn't necessarily introduced by this PR.

### What happens
If a debounced save starts and the user makes another edit while that save is still in flight, navigating away can silently discard the newer edit. Only the first value is persisted.
### Why
Edits in the same session share one currentGen. saveGenRef changes only when the draft is explicitly cleared or reverted, not for every new edit.

Example sequence:

1. Edit A triggers the debounce and saveDraft(A) starts.
2. Edit B replaces pendingDraftRef.current, using the same generation.
3. saveDraft(A) resolves and clears the ref containing B.
4. The user navigates away before B’s debounce fires.
5. Unmount cancels B’s timer, and the flush finds nothing to save.
6. Edit B is lost.

voidspooks:
@costacalvin confirmed, and good catch. Your identity-based clear is what landed, gated on `saved` so a failed write stays in the ref for the unmount flush to retry. The regression test follows your sequence exactly: save A held unresolved, edit B lands, A resolves, unmount before B's debounce, flush must persist B. I verified it fails against the previous head the same way you reproduced it, only saveDraft(A) recorded, and passes with the fix.

Beyond the unit suite (678 green) I also exercised it against the local stack: typed an edit, let its save fire, typed a second edit and left for System Info inside the debounce window, and the second edit survived the round trip. While validating I also drove the no-questionnaire return link live by pointing a seeded system at an environment with no functions, so that path is confirmed working end to end, not just under jest.

@MackOverflow one more commit to pull in when you have a sec: `f5fa8ec` on `voidspooks/ztmf-ui:feat/609-610-cross-nav`.

MackOverflow:
f5fa8ec is in — this addresses the required-before-merge in-flight-save race.

The bug: edits in one session share a generation (saveGenRef only moves on an explicit clear/revert). So if edit A's debounced save was in flight and edit B replaced the pending payload under that same generation, save A's completion cleared the ref that now held B. An unmount before B's own debounce then flushed nothing, and B was silently lost. The old clear also ran ahead of the saved check, so a failed write discarded the payload the unmount flush should have retried.

The fix: the completion handler now captures the exact payload object it wrote and clears the ref only when saved && pendingDraftRef.current === pending — an identity check rather than a generation check. A newer edit installs a distinct object, so an older save's completion can no longer null it out, and a failed write stays pending for the unmount flush to retry.

Coverage: new regression test follows the review sequence exactly — save A held unresolved → edit B lands → A resolves → page unmounts before B's debounce → flush must persist B. Confirmed to fail against the previous head (only saveDraft(A) recorded) and pass with the fix.

Head is now f5fa8ec; commits stay authored by @voidspooks. Ready for re-review @costacalvin.

costacalvin (approved):
This PR cleanly delivers the cross-navigation for #609/#610: a Questionnaire link on System Info and System Info / Pillar Scores buttons on the Questionnaire, keyed correctly across the acronym/id route split, with the Pillar Scores fetch, casing, and access gating all matching the dashboard's existing behavior. Every issue raised in review was addressed and verified — the semantic router links, the noQuestions return link, the res.data?.data ?? [] fallback, and most importantly the draft-persistence work: the mid-debounce unmount flush plus the follow-up fix that clears the pending draft by object identity (gated on a successful save) close the in-flight-save race that could have silently dropped a newer edit. Both fixes ship with real regression tests that I confirmed fail on the prior head and pass now. The local gate is clean (tsc at baseline with zero new errors, eslint clean, all cross-nav/header tests green) and in-repo CI — including Snyk and Deploy — is fully passing. No correctness, security, or authorization concerns remain, so this is good to merge.

LGTM! 🚀

PR #641, opened 2026-07-29
Label non-assignable OpDiv grants and stop revoking them

### LINKED ISSUE #600: Assign OpDivs modal renders unlabeled chips for non-assignable grants and silently revokes them on save
## Recommended labels

`bug`, `area/frontend`, `lang/typescript`, `data-loss`

## Summary

`OpDivGrantModal` uses a single list, `opdivOptions`, for two different jobs: deciding what an admin may **add**, and resolving what an existing grant is **called**. `UserTable` narrows that list to assignable tenants (`!od.is_parent && od.active`), so any grant the user already holds to a parent or inactive OpDiv has no label source and no place in the desired set.

Two defects follow, and the first hides the second:

1. **Unlabeled chip.** The grant renders as an empty MUI chip with a live delete icon. The admin cannot tell which OpDiv it is.
2. **Silent revocation on save.** `handleSave` filters the desired set down to the assignable list before sending it. For an unscoped admin the backend removes anything omitted, so clicking Save revokes a grant the admin never saw and never intended to touch.

### Part 1 - one list doing two jobs

`src/views/UserTable/UserTable.tsx:569-582` fetches every OpDiv, builds a full code map for the table column, then derives a narrower assignable subset for the modal:

```tsx
const all = await fetchOpDivs(true)
const codeMap: Record<number, string> = {}
all.forEach((od) => { codeMap[od.opdiv_id] = od.code })
setOpDivCodeMap(codeMap)

let assignable = all.filter((od) => !od.is_parent && od.active)
if (isOpDivTier(userInfo)) {
  const own = new Set(userInfo.assignedopdivids ?? [])
  assignable = assignable.filter((od) => own.has(od.opdiv_id))
}
setOpDivOptions(assignable)
```

The comment directly above already states the principle:

> Pull the full list (incl. inactive/parent) so any granted id resolves to a code in the OpDivs column; derive the assignable subset from the same response for the grant modal.

That reasoning was applied to the table column and not carried into the modal. Only `opdivOptions` is passed down, so `src/views/OpDivGrantModal/OpDivGrantModal.tsx:50-56` builds its label map from the narrow list:

```tsx
const opdivMap = React.useMemo(() => {
  const map: Record<number, { code: string; name: string }> = {}
  for (const od of opdivOptions) {
    map[od.opdiv_id] = { code: od.code, name: od.name }
  }
  return map
}, [opdivOptions])
```

and `optionLabel` (lines 103-106) falls through to an empty string:

```tsx
const optionLabel = (opdivId: number) => {
  const od = opdivMap[opdivId]
  return od ? `${od.code} - ${od.name}` : ''
}
```

`optionLabel` was written by #378 for dropdown rows, where every id is by construction in the options list and the `''` branch is unreachable. #438 later reused it to render chips built from **fetched grants**, which can contain ids the options list does not. Same shape as the `AssignSystemModal` label bug, a label helper written against one list, then reused against a different one.

### Part 2 - the save filter conflates two reasons for absence

`OpDivGrantModal.tsx:108-124`:

```tsx
const scopedIds = localOpDivs.filter((id) => sortedOptionIds.includes(id))
setUserOpDivs(String(userid), scopedIds)
```

`setUserOpDivs` is a full replace, not a merge (`src/utils/userOpdivs.ts`):

> replaces the full grant set in one batch request. The backend reconciles the desired set against current grants (adds missing, removes extra)

The filter's intent is legitimate and documented: an `OPDIV_ADMIN` must never submit an id outside their scope, because the backend rejects the whole request if they do (`usersopdivs.go:189-197` returns `ErrForbidden`). The defect is that `sortedOptionIds` excludes ids for **two unrelated reasons**, and the filter treats them the same:

- *outside the acting admin's own scope* - correct to strip, avoids a 403
- *not assignable to anyone* (parent or inactive) - should be preserved

### Why only unscoped admins lose data

The backend's removal loop (`usersopdivs.go:242-247`) guards each removal:

```go
for _, id := range current {
    if !desiredSet[id] && (authdUser.HasUnscopedRead() || authdUser.IsAssignedOpDiv(id)) {
        toRemove = append(toRemove, id)
    }
}
```

For an `OPDIV_ADMIN`, a grant they do not hold fails `IsAssignedOpDiv` and is **not** removed, so the backend already protects them. For an unscoped admin `HasUnscopedRead()` is true, the guard passes for every id, and anything the FE omitted is removed.

So the data-loss path is specifically **unscoped admin (`OWNER` / `HHS_ADMIN`) + a target holding a parent or inactive grant**. This is also why the FE fix is narrow: for unscoped admins the filter is not protecting anything, because the scope gate they are being protected from does not apply to them.

## Suggested fix

Two independent changes. Part 2 is the one that loses data.

**Part 1 - give the modal a label source separate from its option source.**

Build a full `Record<number, {code, name}>` in `UserTable` from the same `fetchOpDivs(true)` response already in hand, and pass it as a new `opdivLabelMap` prop. `OpDivGrantModal` merges it under `opdivOptions` for labels only; `sortedOptionIds` stays derived from `opdivOptions` so the dropdown is unchanged. This mirrors the `allSystems` / `assignable` split used in `AssignSystemModal` after #586. Add an id-based fallback (`OpDiv #{id}`) for the case where even the full list cannot resolve an id, so a chip is never blank.

**Part 2 - only filter when the filter is actually protecting something.**

```tsx
const scopedIds = hasUnscopedRead(userInfo)
  ? localOpDivs
  : localOpDivs.filter((id) => sortedOptionIds.includes(id))
```

An unscoped admin sends the full set, so nothing is silently dropped. A scoped admin keeps the existing filter, which still prevents the 403, and the backend's `toRemove` guard already preserves any grant they cannot see. Requires `userInfo` in the modal (via `useContextProp`, as `UserTable` does) or an `isUnscoped` prop.

Worth considering as a follow-up rather than here: the modal is a full-replace UI built on a list the admin cannot fully see. A delta-based endpoint, or surfacing non-assignable grants as read-only chips that are visible but not deletable, would remove the whole class of problem. Out of scope for a bug fix.

## Related
- Fix in #586: a label helper written for an options list, then reused against a fetched-value list that is a superset.
- Found while verifying #586 against the empire seed.

### Steps to Reproduce

1. Reload the empire seed so users hold the EMPIRE grant (`_test_data_empire.sql:130` grants EMPIRE to every empire-themed user; EMPIRE is seeded `is_parent = TRUE` at line 14 and re-asserted at line 20).
2. Sign in as an unscoped admin (`OWNER`, e.g. Grand Moff Tarkin).
3. Navigate to Users. Confirm the target row's OpDivs column reads `EMPIRE`.
4. Click Assign OpDivs on that user.

### Expected Behavior

**Expected**: the existing EMPIRE grant renders with a readable label. Saving without touching it leaves the grant intact.

### Actual Behavior

**Actual**: an empty chip renders with only a delete icon. EMPIRE is correctly absent from the dropdown, but it is also absent from the desired set, so clicking Save issues `PUT /users/:id/opdivs {opdiv_ids: []}` and the grant is revoked. The users table then shows no OpDivs for that user.

### Environment

Local Dev

### Acceptance Criteria

- [ ] A grant to a parent or inactive OpDiv renders with a readable label in the modal, matching what the users-table OpDivs column shows for the same grant.
- [ ] No chip in the modal is ever blank; an unresolvable id falls back to an id-based label.
- [ ] An unscoped admin opening the modal for a user with a non-assignable grant and clicking Save leaves that grant intact.
- [ ] Adding or removing an assignable OpDiv still works unchanged for both unscoped and scoped admins.
- [ ] A scoped admin still never submits an id outside their own scope (no 403 regression).
- [ ] The dropdown still excludes parents and inactive OpDivs. This ticket does not change what is assignable.
- [ ] Tests cover: a non-assignable grant renders labeled; save preserves it for an unscoped admin; the scoped-admin filter still strips out-of-scope ids.

### Severity

None

### Affected Component

_No response_

### Screenshots

_No response_

### Logs / Error Messages

```shell

```

### Workaround

_No response_

tarratsco:
## Summary

Closes #600

The Assign OpDivs modal rendered a user's existing grants to non-assignable OpDivs (a parent or inactive division) as blank chips with only a delete icon, and silently revoked them when an unscoped admin saved without making changes.

Root cause: the modal used a single scoped list (`opdivOptions`) for two jobs: populating the dropdown and resolving labels for existing grants. When a granted OpDiv fell outside that list, it had no label (blank chip), and the save filtered it out of the payload. The backend's removal logic only protects scoped admins, so for an unscoped admin (OWNER/HHS_ADMIN) an omitted grant read as a revocation and the grant was dropped.

This is a frontend-only fix that mirrors the existing `AssignSystemModal` pattern in the same table, which already separates a full label source from the scoped options.

## Changes

* **Full label source.** `UserTable` already fetches all OpDivs (incl. parent/inactive); it now builds a full `opdivLabelMap` from that same response and passes it to the modal, separate from the scoped options. The modal labels grants from it, with an `OpDiv #{id}` fallback so a chip is never blank.
* **Options vs. dropdown.** The Autocomplete options are now the union of assignable + current grants, so an out-of-scope chip resolves (no MUI "value not in options" warning); `filterOptions` narrows the dropdown back to the assignable set so those grants stay non-selectable.
* **Scope-gated save.** The save-time filter is gated on a new `enforceCallerScope` prop, computed as `isOpDivTier(userInfo)` (the same predicate that narrows the options). Scoped callers drop out-of-scope IDs; unscoped callers send every grant as-is, preserving non-assignable grants.

## Acceptance criteria

* [x] Non-assignable grants display readable labels.
* [x] No blank chips; fallback to `OpDiv #{id}`.
* [x] Unscoped admin saves preserve non-assignable grants.
* [x] Assignable OpDiv add/remove unchanged.
* [x] Scoped admins cannot submit out-of-scope grants.
* [x] Dropdown excludes parent/inactive OpDivs.
* [x] Coverage for labeled rendering and preservation.

## UI

BEFORE

https://github.com/user-attachments/assets/96d62e4e-2d11-4bfe-93ae-06f32f7e761e

AFTER

https://github.com/user-attachments/assets/7defa265-ad0e-4d72-8efe-f9beda60ab42

## Testing

* Reframed the existing scope test as the scoped-admin path; added scoped-preservation, labeled-rendering, and `OpDiv#{id}`-fallback tests. Each new test is litmus-verified, failing against the pre-fix code.
* `yarn jest`: full suite green (670 passed); `tsc --noEmit` clean (pre-existing unrelated warnings only).

### DIFF
--- a/src/views/OpDivGrantModal/OpDivGrantModal.test.tsx
+++ b/src/views/OpDivGrantModal/OpDivGrantModal.test.tsx
@@ -54,6 +54,16 @@ const opdivOptions: OpDiv[] = [
   },
 ]
 
+// Full label source (incl. the non-assignable OpDiv 99, e.g. a parent/inactive
+// division), so the modal can label a grant to it even though it is absent from
+// opdivOptions. Id 77 is intentionally absent to exercise the "OpDiv #{id}"
+// fallback.
+const opdivLabelMap: Record<number, { code: string; name: string }> = {
+  1: { code: 'AAA', name: 'Division A' },
+  2: { code: 'BBB', name: 'Division B' },
+  99: { code: 'ZZZ', name: 'Parent Division' },
+}
+
 function renderModal(
   overrides: Partial<React.ComponentProps<typeof OpDivGrantModal>> = {}
 ) {
@@ -64,6 +74,8 @@ function renderModal(
       userid={USER_ID}
       userName="Test User"
       opdivOptions={opdivOptions}
+      opdivLabelMap={opdivLabelMap}
+      enforceCallerScope={true}
       onChanged={jest.fn()}
       {...overrides}
     />
@@ -75,9 +87,9 @@ beforeEach(() => {
   mockedNavigate.mockReset()
 })
 
-// The most important test: verifies the scopedIds filter strips out-of-scope
-// grants before the PUT so an OPDIV_ADMIN never triggers a backend 403.
-test('PUT body excludes grants the target user holds outside the caller scope', async () => {
+// Scoped caller (OPDIV_ADMIN, enforceCallerScope=true): the save strips
+// out-of-scope grants before the PUT so the backend scope gate never 403s.
+test('scoped caller: PUT body excludes grants the target holds outside caller scope', async () => {
   // Target user holds [1, 2, 99]. OpDiv 99 is absent from opdivOptions
   // (out of caller scope), so only [1, 2] must reach the batch endpoint.
   mock
@@ -85,7 +97,7 @@ test('PUT body excludes grants the target user holds outside the caller scope', 
     .reply(200, { data: [1, 2, 99] })
   mock.onPut(`/users/${USER_ID}/opdivs`).reply(204)
 
-  renderModal()
+  renderModal({ enforceCallerScope: true })
   await waitFor(() => expect(mock.history.get).toHaveLength(1))
 
   await userEvent.click(screen.getByRole('button', { name: /^save$/i }))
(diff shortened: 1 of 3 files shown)


MackOverflow:
Moving this to #642 — fork PRs can't access the org secrets required for Snyk and the deploy CI gates, so those checks never run here. Your commits and authorship carry over unchanged. Thanks @tarratsco!

PR #642, opened 2026-07-29
Label non-assignable OpDiv grants and stop revoking them

### LINKED ISSUE #600: Assign OpDivs modal renders unlabeled chips for non-assignable grants and silently revokes them on save

MackOverflow:
> **Note:** In-repo copy of #641 by @tarratsco. Moved from the fork so Snyk and the deploy CI gates can run — those jobs require org secrets unavailable to fork PRs. Commits and authorship are unchanged.
>
> Refs #641

---

## Summary

Closes #600

The Assign OpDivs modal rendered a user's existing grants to non-assignable OpDivs (a parent or inactive division) as blank chips with only a delete icon, and silently revoked them when an unscoped admin saved without making changes.

Root cause: the modal used a single scoped list (`opdivOptions`) for two jobs: populating the dropdown and resolving labels for existing grants. When a granted OpDiv fell outside that list, it had no label (blank chip), and the save filtered it out of the payload. The backend's removal logic only protects scoped admins, so for an unscoped admin (OWNER/HHS_ADMIN) an omitted grant read as a revocation and the grant was dropped.

This is a frontend-only fix that mirrors the existing `AssignSystemModal` pattern in the same table, which already separates a full label source from the scoped options.

## Changes

* **Full label source.** `UserTable` already fetches all OpDivs (incl. parent/inactive); it now builds a full `opdivLabelMap` from that same response and passes it to the modal, separate from the scoped options. The modal labels grants from it, with an `OpDiv #{id}` fallback so a chip is never blank.
* **Options vs. dropdown.** The Autocomplete options are now the union of assignable + current grants, so an out-of-scope chip resolves (no MUI "value not in options" warning); `filterOptions` narrows the dropdown back to the assignable set so those grants stay non-selectable.
* **Scope-gated save.** The save-time filter is gated on a new `enforceCallerScope` prop, computed as `isOpDivTier(userInfo)` (the same predicate that narrows the options). Scoped callers drop out-of-scope IDs; unscoped callers send every grant as-is, preserving non-assignable grants.

## Acceptance criteria

* [x] Non-assignable grants display readable labels.
* [x] No blank chips; fallback to `OpDiv #{id}`.
* [x] Unscoped admin saves preserve non-assignable grants.
* [x] Assignable OpDiv add/remove unchanged.
* [x] Scoped admins cannot submit out-of-scope grants.
* [x] Dropdown excludes parent/inactive OpDivs.
* [x] Coverage for labeled rendering and preservation.

## Testing

* Reframed the existing scope test as the scoped-admin path; added scoped-preservation, labeled-rendering, and `OpDiv#{id}`-fallback tests. Each new test is litmus-verified, failing against the pre-fix code.
* `yarn jest`: full suite green (670 passed); `tsc --noEmit` clean (pre-existing unrelated warnings only).

### DIFF
--- a/src/views/OpDivGrantModal/OpDivGrantModal.test.tsx
+++ b/src/views/OpDivGrantModal/OpDivGrantModal.test.tsx
@@ -76,6 +76,7 @@ function renderModal(
       opdivOptions={opdivOptions}
       opdivLabelMap={opdivLabelMap}
       enforceCallerScope={true}
+      callerGrantIds={[1, 2]}
       onChanged={jest.fn()}
       {...overrides}
     />
@@ -129,6 +130,28 @@ test('unscoped caller: PUT body preserves grants outside the assignable set', as
   expect(body.opdiv_ids).toHaveLength(3)
 })
 
+// Scoped caller whose OWN grant (99) is to an OpDiv since re-parented or
+// deactivated: 99 is in callerGrantIds (the backend still sees IsAssignedOpDiv
+// = true) but absent from opdivOptions (parent/inactive is filtered out). The
+// save must PRESERVE 99 - filtering on the narrower assignable set would strip
+// it from the PUT and the backend's toRemove gate (pure grant membership) would
+// then silently revoke the target's grant. The save boundary is the caller's
+// raw scope, not the dropdown's assignable set.
+test('scoped caller: preserves a caller-held grant that is no longer assignable', async () => {
+  mock.onGet(`/users/${USER_ID}/assignedopdivs`).reply(200, { data: [1, 99] })
+  mock.onPut(`/users/${USER_ID}/opdivs`).reply(204)
+
+  renderModal({ enforceCallerScope: true, callerGrantIds: [1, 99] })
+  await waitFor(() => expect(mock.history.get).toHaveLength(1))
+
+  await userEvent.click(screen.getByRole('button', { name: /^save$/i }))
+
+  await waitFor(() => expect(mock.history.put).toHaveLength(1))
+  const body = JSON.parse(mock.history.put[0].data)
+  expect(body.opdiv_ids).toEqual(expect.arrayContaining([1, 99]))
+  expect(body.opdiv_ids).toHaveLength(2)
+})
+
 // A grant to a non-assignable OpDiv (99, absent from opdivOptions) still chips
 // with a readable label from opdivLabelMap - never a blank chip.
 test('labels a grant to a non-assignable OpDiv from the full label map', async () => {
@@ -169,13 +192,14 @@ test('dropdown excludes a non-assignable grant even though it chips', async () =
   expect(screen.queryByRole('option', { name: /ZZZ/ })).not.toBeInTheDocument()
 })
 
-// A scoped caller's save strips out-of-scope ids regardless, so a delete would
-// be a silent no-op. The out-of-scope chip renders WITHOUT a delete affordance,
-// while an in-scope chip keeps it.
-test('scoped caller: an out-of-scope grant chip is not deletable', async () => {
+// A grant outside the caller's own backend scope (99 here is held by the target
+// via another admin, not by this caller) is stripped on save regardless, so a
+// delete would be a silent no-op: it renders WITHOUT a delete affordance, while
+// an in-scope chip keeps it.
+test('scoped caller: a chip outside the caller scope is not deletable', async () => {
   mock.onGet(`/users/${USER_ID}/assignedopdivs`).reply(200, { data: [1, 99] })
 
-  renderModal({ enforceCallerScope: true })
+  renderModal({ enforceCallerScope: true, callerGrantIds: [1, 2] })
 
   const inScope = (await screen.findByText('AAA - Division A')).closest(
     '.MuiChip-root'
@@ -186,6 +210,21 @@ test('scoped caller: an out-of-scope grant chip is not deletable', async () => {
 
   expect(inScope.querySelector('.MuiChip-deleteIcon')).not.toBeNull()
   expect(outOfScope.querySelector('.MuiChip-deleteIcon')).toBeNull()
+})
+
+// A caller-held grant that is merely non-assignable now (99 in callerGrantIds
+// but absent from opdivOptions, e.g. re-parented/deactivated) stays deletable:
+// removing it is a real, permitted revocation, unlike an out-of-scope grant.
+test('scoped caller: a caller-held but non-assignable chip stays deletable', async () => {
+  mock.onGet(`/users/${USER_ID}/assignedopdivs`).reply(200, { data: [1, 99] })
+
+  renderModal({ enforceCallerScope: true, callerGrantIds: [1, 99] })
+
+  const heldNonAssignable = (
+    await screen.findByText('ZZZ - Parent Division')
+  ).closest('.MuiChip-root') as HTMLElement
+
+  expect(heldNonAssignable.querySelector('.MuiChip-deleteIcon')).not.toBeNull()
 })
 
 // An unscoped caller's removal really revokes, so their out-of-scope chip must
@@ -343,6 +382,7 @@ test('closing after a fetch failure resets error state so Save re-enables on reo
       opdivOptions={opdivOptions}
       opdivLabelMap={opdivLabelMap}
       enforceCallerScope={true}
+      callerGrantIds={[1, 2]}
       onChanged={jest.fn()}
     />
   )
@@ -355,6 +395,7 @@ test('closing after a fetch failure resets error state so Save re-enables on reo
       opdivOptions={opdivOptions}
       opdivLabelMap={opdivLabelMap}
       enforceCallerScope={true}
+      callerGrantIds={[1, 2]}
       onChanged={jest.fn()}
     />
   )
@@ -406,6 +447,7 @@ test('stale fetch from a prior user is discarded when userid changes', async () 
       opdivOptions={opdivOptions}
       opdivLabelMap={opdivLabelMap}
       enforceCallerScope={true}
+      callerGrantIds={[1, 2]}
       onChanged={onChanged}
     />
   )
--- a/src/views/OpDivGrantModal/OpDivGrantModal.test.tsx
+++ b/src/views/OpDivGrantModal/OpDivGrantModal.test.tsx
@@ -54,6 +54,16 @@ const opdivOptions: OpDiv[] = [
   },
 ]
 
+// Full label source (incl. the non-assignable OpDiv 99, e.g. a parent/inactive
+// division), so the modal can label a grant to it even though it is absent from
+// opdivOptions. Id 77 is intentionally absent to exercise the "OpDiv #{id}"
+// fallback.
+const opdivLabelMap: Record<number, { code: string; name: string }> = {
+  1: { code: 'AAA', name: 'Division A' },
+  2: { code: 'BBB', name: 'Division B' },
+  99: { code: 'ZZZ', name: 'Parent Division' },
+}
+
 function renderModal(
   overrides: Partial<React.ComponentProps<typeof OpDivGrantModal>> = {}
 ) {
@@ -64,6 +74,8 @@ function renderModal(
       userid={USER_ID}
       userName="Test User"
       opdivOptions={opdivOptions}
+      opdivLabelMap={opdivLabelMap}
+      enforceCallerScope={true}
       onChanged={jest.fn()}
       {...overrides}
     />
@@ -75,9 +87,9 @@ beforeEach(() => {
   mockedNavigate.mockReset()
 })
 
-// The most important test: verifies the scopedIds filter strips out-of-scope
-// grants before the PUT so an OPDIV_ADMIN never triggers a backend 403.
-test('PUT body excludes grants the target user holds outside the caller scope', async () => {
+// Scoped caller (OPDIV_ADMIN, enforceCallerScope=true): the save strips
+// out-of-scope grants before the PUT so the backend scope gate never 403s.
+test('scoped caller: PUT body excludes grants the target holds outside caller scope', async () => {
   // Target user holds [1, 2, 99]. OpDiv 99 is absent from opdivOptions
   // (out of caller scope), so only [1, 2] must reach the batch endpoint.
   mock
@@ -85,7 +97,7 @@ test('PUT body excludes grants the target user holds outside the caller scope', 
     .reply(200, { data: [1, 2, 99] })
   mock.onPut(`/users/${USER_ID}/opdivs`).reply(204)
 
-  renderModal()
+  renderModal({ enforceCallerScope: true })
   await waitFor(() => expect(mock.history.get).toHaveLength(1))
 
   await userEvent.click(screen.getByRole('button', { name: /^save$/i }))
@@ -95,6 +107,47 @@ test('PUT body excludes grants the target user holds outside the caller scope', 
   expect(body.opdiv_ids).toHaveLength(2)
   expect(body.opdiv_ids).toEqual(expect.arrayContaining([1, 2]))
   expect(body.opdiv_ids).not.toContain(99)
+})
+
+// Unscoped caller (OWNER/HHS_ADMIN, enforceCallerScope=false): the save must
+// PRESERVE the target's non-assignable grants. Omitting 99 would read as a
+// revocation to the backend and silently drop the grant.
+test('unscoped caller: PUT body preserves grants outside the assignable set', async () => {
+  mock
+    .onGet(`/users/${USER_ID}/assignedopdivs`)
+    .reply(200, { data: [1, 2, 99] })
+  mock.onPut(`/users/${USER_ID}/opdivs`).reply(204)
+
+  renderModal({ enforceCallerScope: false })
+  await waitFor(() => expect(mock.history.get).toHaveLength(1))
+
+  await userEvent.click(screen.getByRole('button', { name: /^save$/i }))
+
+  await waitFor(() => expect(mock.history.put).toHaveLength(1))
+  const body = JSON.parse(mock.history.put[0].data)
+  expect(body.opdiv_ids).toEqual(expect.arrayContaining([1, 2, 99]))
+  expect(body.opdiv_ids).toHaveLength(3)
+})
+
+// A grant to a non-assignable OpDiv (99, absent from opdivOptions) still chips
+// with a readable label from opdivLabelMap - never a blank chip.
+test('labels a grant to a non-assignable OpDiv from the full label map', async () => {
+  mock.onGet(`/users/${USER_ID}/assignedopdivs`).reply(200, { data: [99] })
+
+  renderModal()
+
+  // The chip renders the label map entry, not a blank chip or a raw id.
+  expect(await screen.findByText('ZZZ - Parent Division')).toBeInTheDocument()
+})
+
+// A grant to an OpDiv missing from the label map falls back to "OpDiv #{id}"
+// rather than an empty chip.
+test('falls back to "OpDiv #{id}" for a grant missing from the label map', async () => {
+  mock.onGet(`/users/${USER_ID}/assignedopdivs`).reply(200, { data: [77] })
+
+  renderModal()
+
+  expect(await screen.findByText('OpDiv #77')).toBeInTheDocument()
 })
 
 test('success: modal closes and onChanged fires after save', async () => {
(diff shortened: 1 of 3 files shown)


jsos3-cms:
Thanks @tarratsco — reviewed the rehome (#641 → here). I checked the core logic against the backend contract directly, and it's correct. Approving with two minor, optional notes.

## The fix matches the backend contract exactly

The linchpin is whether the batch endpoint (`PUT /users/{userid}/opdivs` → `SetUserOpDivs`) preserves a target's out-of-scope grants. Reading the handler (`backend/cmd/api/internal/controller/useropdivs.go`), its reconciliation only adds/removes ids the caller holds for a scoped admin, but for an unscoped admin (`HasUnscopedRead`) `toRemove` includes any current id absent from the desired set. So:

- **Scoped (OPDIV_ADMIN):** must send only in-scope ids (the scope gate 403s otherwise), and the backend leaves out-of-scope grants untouched → the `enforceCallerScope` filter is right.
- **Unscoped (OWNER/HHS_ADMIN):** the backend revokes any omitted grant → sending the full set (incl. non-assignable) is required → the preserve path is right.

The root-cause writeup ("the backend's removal logic only protects scoped admins") is precisely what the handler's `toRemove` loop does. Good diagnosis.

The chip UX follows correctly: out-of-scope chips are non-deletable for a scoped caller (the save strips them regardless, so a delete would be a silent no-op) and deletable for an unscoped caller (their removal really revokes). Keeping the dropdown scoped via `filterOptions` while `options` carries current grants is the right way to resolve the chip without offering it for re-selection.

## Two optional notes (non-blocking)

1. `enforceCallerScope={isOpDivTier(userInfo)}` is correct for every role that can reach this modal today. If you want it to mirror the backend's exact predicate, `!isUnscopedWriteAdmin(userInfo)` is the direct inverse of the backend's `HasUnscopedRead` gate — it stays correct if a scoped *non-OpDiv* write role is ever added, whereas `isOpDivTier` would treat such a role as unscoped. Equivalent today; just future-proofing.
2. Benign edge: if `userInfo.role` were unset when the modal mounts, `isOpDivTier` → false → the unscoped path. Not reachable given UserTable gates access on role before the modal opens, so no change needed — just noting it.

## Tests / quality

Coverage is thorough (scoped-strip, unscoped-preserve, label fallback, `OpDiv #{id}` fallback, dropdown exclusion, chip deletability both ways), and the comments explain the *why* well. Since the backend enforces scope regardless, this is a data-integrity fix (stops an unscoped admin's accidental revocation), not a security gap.

LGTM once the CI gates (the reason for the rehome) are green.

jsos3-cms (dismissed):

jsos3-cms:
Following up on my review above — I ran an adversarial second pass to double-check, and it surfaced a real edge I missed. My "matches the backend contract exactly" was right for the unscoped path but overstated: the **scoped (OPDIV_ADMIN) path has a same-class silent-revocation hole**.

`assignableIds` is derived from `opdivOptions`, which `UserTable` narrows to `!is_parent && active` *before* intersecting with the caller's own grants. But the backend's scope is pure grant-membership — `IsAssignedOpDiv`, sourced from `users_opdivs` with no parent/active filter (the `SetUserOpDivs` scope gate and its `toRemove` loop). So `assignableIds` is a **narrower** boundary than the backend's actual scope, and it's doing double duty: narrowing the dropdown (correct) *and* defining "what may I preserve on save" (too narrow).

**Repro:** an OPDIV_ADMIN holds a grant to OpDiv X; X is later reorganized to `is_parent=true` or deactivated (no cascade to `users_opdivs`); a target user also holds X. The admin opens Assign OpDivs for that user (`sharesOpDiv` passes), makes no changes, clicks Save. X isn't in the admin's `assignableIds` (parent/inactive-filtered out), so `enforceCallerScope` strips it from the PUT — but the backend still sees `IsAssignedOpDiv(X)=true` for the admin, so X lands in `toRemove` and the target's grant to X is **silently revoked**. That's the same defect this PR closes, reached via the caller's own scope instead of the unscoped-admin path.

**Suggested fix:** thread the caller's *raw* own-grant ids (not the parent/active-filtered `opdivOptions`) into the modal as the save-time preserve boundary, keeping `opdivOptions` for the dropdown only — plus a test for the "OPDIV_ADMIN owns a now-non-assignable OpDiv shared with the target" case (no current fixture exercises it).

Not necessarily merge-blocking (narrow — needs a reorg *after* the grant), but it's the same class as the bug being fixed, so worth closing on the caller side too rather than leaving a second door open.

tarratsco:
@jsos3-cms pushed change to address the feedback!

MackOverflow:
@jsos3-cms — the follow-up fix for the scoped-path edge you flagged is now on this branch as `2196d05`.

@tarratsco's "pushed change to address the feedback" landed on the fork branch, so it wasn't visible here (this is the in-repo copy on a separate branch). I pulled that commit up onto this branch, preserving his authorship, so #642 now carries the fix.

**What `2196d05` does:** it adds a `callerGrantIds` prop carrying the caller's raw backend scope (`userInfo.assignedopdivids`, unfiltered by parent/active) and keys the save-preserve filter and the chip-lock on that instead of the narrower `assignableIds`. So a grant the caller holds to an OpDiv that was later re-parented or deactivated is preserved on save (and its chip stays deletable) rather than stripped and silently revoked by the backend's `toRemove` gate — closing the same-class hole on the caller side, exactly as you described.

Verified locally: `yarn jest src/views/OpDivGrantModal` is green (21/21, including two new tests — a caller-held-but-non-assignable preserve case and its deletability), and `tsc --noEmit` is clean. CI has re-triggered on the new head.

jsos3-cms (dismissed):
Reviewed `2196d05` (the follow-up to the scoped-path edge I flagged). **Approving** — the fix is correct, and I verified the premise directly against the backend rather than taking the comments at their word.

## The fix is right, and for the right reason

The load-bearing claim is that `callerGrantIds` carries the caller's *true* backend scope. It does, and the two sides are not merely consistent — they are literally the same data. `/users/current` returns `model.UserFromContext`, populated by `findUser`, whose select list includes `(SELECT ARRAY_AGG(opdiv_id) FROM users_opdivs WHERE userid = users.userid) AS assignedopdivids` (`backend/internal/model/users.go:487`). The backend's removal gate gates on `authdUser.IsAssignedOpDiv(id)`, which reads that same `AssignedOpDivIDs` field (`usersopdivs.go:243`). So keying the save filter on `assignedopdivids` makes the frontend's preserve boundary exactly the backend's `toRemove` authority — no wider, no narrower. That is the correct boundary, and `assignableIds` (parent/active-filtered) was strictly too narrow for the job it was doing.

The superset invariant the prop docs assert is structurally guaranteed at the call site, not just asserted: `UserTable.tsx:589` derives `opdivOptions` by filtering `!is_parent && active` and *then* intersecting with `new Set(userInfo.assignedopdivids ?? [])` — the same expression now passed as `callerGrantIds`. So `callerScope ⊇ assignableIds` holds by construction, and it survives the `catch` branch at line 595 that empties `opdivOptions` on a fetch failure (dropdown goes empty, save still preserves correctly).

Splitting the chip lock is also right: a caller-held grant to a now parent/inactive OpDiv is genuinely revocable by that caller, so it should keep its delete affordance, while a grant from *another* admin stays locked because the save strips it regardless. Those are now two distinct cases instead of one conflated one.

## Verification I ran

`tsc --noEmit` clean; `jest src/views/OpDivGrantModal` green at 21/21; only one production call site (`UserTable.tsx:931`), and the prop is required so the compiler enforces it.

The part worth calling out: I mutation-tested the two new tests by reverting `OpDivGrantModal.tsx` to `2196d05~1` while keeping the new test file. Both fail, with exactly the failure modes the fix predicts — the preserve test gets `[1]` instead of `[1, 99]` (the silent revocation, reproduced), and the deletability test gets `null` for the delete icon (the over-broad lock). That is real coverage of the `assignableIds` / `callerScope` divergence, not fixtures that would pass either way.

## Adversarial pass: one finding raised, refuted

I ran a second pass prompted to refute this change. It flagged as blocking that `assignedopdivids` is typed `number[] | null` and passed through unvalidated by `authLoader`, so a caller who "actually holds `[1, 99]`" could arrive with `callerGrantIds = []`, save `[]`, and have the backend revoke the target's grants.

That is not reachable. The field is null precisely when the caller has zero `users_opdivs` rows — `ARRAY_AGG` over an empty set returns NULL — which is the same condition that makes `IsAssignedOpDiv(id)` false for every id on the backend. In that state `toRemove` is empty and a `PUT []` from a scoped caller removes nothing. The frontend's preserve set and the backend's removal authority are computed from one table, so they cannot disagree; they fail closed together. Worth recording since the reasoning is not obvious from the frontend alone.

Also noted and checked: the unscoped OWNER/HHS_ADMIN path is untouched (`enforceCallerScope === false` still sends `localOpDivs` verbatim), and the `callerScope` memo has no referential-stability consumers — it is read inside `handleSave` and inline in `renderTags`, neither of which is a dependency array.

## One nit, non-blocking

`UserTable.tsx:938-940` still reads "Same predicate that narrowed `opdivOptions` above, so options and save agree." This commit deliberately makes options and save use *different* sets — that separation is the entire fix — so the comment now describes the pre-fix design in a security-sensitive spot. Suggest something closer to: `enforceCallerScope` selects the scoped save path, while dropdown assignability (`opdivOptions`) and the save-time preserve boundary (`callerGrantIds`) are intentionally different sets.

Fine as a follow-up or a quick amend; not worth holding the merge.

## Pre-existing, for the backlog rather than this PR

The caller's scope comes from route-loader `userInfo`, which is not refreshed when the caller's own grants change mid-session. If the caller *lost* a grant, the stale-wide set produces a loud 403 at the pre-gate (`usersopdivs.go:191-198`), which is a safe failure. If the caller *gained* one, the stale-narrow set could strip it. Both directions predate this commit — `opdivOptions` reads the same stale `userInfo` — so this changes nothing about the merge, but it is the last remaining way the frontend's idea of scope can drift from the backend's, and it would be worth an issue.

jsos3-cms:
@tarratsco one loose end from my earlier review that the follow-up commit didn't pick up. Non-blocking, and I've already approved, so this is your call whether to fold in now or leave for later.

`enforceCallerScope={isOpDivTier(userInfo)}` in `UserTable.tsx` is correct for every role that can reach this modal today. `isOpDivTier` is `OPDIV_ADMIN | OPDIV_READONLY_ADMIN` and `isUnscopedWriteAdmin` is `OWNER | HHS_ADMIN | ADMIN`, so across the roles that can actually open Assign OpDivs, `isOpDivTier(u)` and `!isUnscopedWriteAdmin(u)` return the same thing. Nothing is wrong as merged.

The reason I'd still prefer `!isUnscopedWriteAdmin(userInfo)` is that it is the direct inverse of the predicate the backend actually branches on. `SetUserOpDivs` splits on `HasUnscopedRead()`, so the frontend question is really "is this caller unscoped?" and `isOpDivTier` answers a different question that happens to correlate. If a scoped write role is ever added that is not `OPDIV_*`, `isOpDivTier` returns false, the modal treats that caller as unscoped, the save sends the target's full grant set, and the backend's pre-gate rejects every id the caller doesn't hold. That fails loud with a 403 rather than corrupting anything, so it is a future-proofing point and not a latent bug. `!isUnscopedWriteAdmin` would keep working for such a role with no change here.

There's a second, smaller reason. The comment above these helpers in `userRoles.ts` says both `isHHSTier` and `isOpDivTier` deliberately cover the read-only variant of their tier, and that callers "must not use these as a write gate." `enforceCallerScope` isn't a permission gate in the strict sense, since it shapes the request rather than deciding whether the write happens, so I don't think it violates that instruction. But it is a write-path decision, and `isUnscopedWriteAdmin` is the helper written for write-path questions. Using it would put the call site on the side of that guidance rather than adjacent to it.

Concretely, the change is one line, and the comment above it stops needing the "same predicate that narrowed opdivOptions" framing:

```ts
enforceCallerScope={!isUnscopedWriteAdmin(userInfo)}
```

Worth pairing with the `UserTable.tsx` comment nit from my review, since both are in the same three-line neighborhood: that comment still says options and save agree, which `2196d05` intentionally made false when it split the dropdown's assignable set from the save-time preserve boundary.

Separately, I filed #643 for the stale-caller-scope issue we discussed. That one is genuinely out of scope here since it predates this PR, but it's the last remaining way the frontend's idea of caller scope can drift from the backend's, so it seemed worth tracking rather than losing in a thread.

tarratsco:
Its a quick change, so I'm folding it in for consistency @jsos3-cms.

MackOverflow:
Pulled up one more commit from @tarratsco's fork — `bd1515f` *"gate OpDiv grant save on the unscoped-write predicate"* — addressing @jsos3-cms's optional note #1. Authorship preserved.

**What it changes:** swaps the `enforceCallerScope` predicate from `isOpDivTier(userInfo)` to `!isUnscopedWriteAdmin(userInfo)` — the direct inverse of the backend's `HasUnscopedRead`/write branch, so the frontend gates on exactly what the backend decides on. Two upshots:
- Future-proof: stays correct if a scoped *non-OpDiv* write role is ever added (`isOpDivTier` would have mis-treated such a role as unscoped).
- Safer default for note #2's edge: an unset `userInfo.role` now falls to the scoped path (no silent revocation) rather than the unscoped one.

Verified locally: `tsc --noEmit` clean and the OpDivGrantModal suite green. CI has re-triggered on the new head (`bd1515f`).

This branch now carries all four commits — the two original plus both review-driven follow-ups.

jsos3-cms (approved):
Amazing work! LGTM!

PR #644, opened 2026-07-30
Nightly sync: main -> impl (2026-07-30)

jsos3-cms:
Automated nightly sync of `main` into `impl` — **main wins every overlap**.

This branch is `impl` with `main` merged in using `-X theirs`, so the result is deterministic: main's version wins any conflict, and impl-only files that don't conflict are preserved. There should be nothing to hand-resolve.

⚠️ **Merge this with a MERGE COMMIT, not squash.** `main` and `impl` share no squash ancestry; squashing here would leave them diverged and every future nightly sync would re-conflict. A merge commit re-establishes shared ancestry and closes the gap.

- Review the net change, then merge (merge commit). Merging pushes to `impl` and triggers the impl deploy (orchestration-impl).
- To smoke-test the merged result in the impl environment first, run **Manual Deploy to IMPL** with `--ref automation/sync-main-to-impl`.

_Opened by the nightly-impl-sync workflow._

### DIFF
--- a/src/utils/systemMetadataVocab.test.ts
+++ b/src/utils/systemMetadataVocab.test.ts
@@ -145,9 +145,9 @@ describe('booleanOptions', () => {
   })
 
   it('overrides the true/false labels but keeps values and Unknown', () => {
-    expect(booleanOptions({ true: 'Not funded', false: 'Funded' })).toEqual([
-      { value: 'true', label: 'Not funded' },
-      { value: 'false', label: 'Funded' },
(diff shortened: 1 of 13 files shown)


jsos3-cms:
Skipping in favor of the Monday night sync.

Five PRs have merged to `main` since this branch was built (#642, #645, #648, #649, #656), so what's here is a stale snapshot and reviewing it would mean reviewing something we aren't going to ship. Tonight's scheduled run rebuilds the branch from current `main` and opens a single fresh PR, which will be the one to review.

Closing with zero points and no sprint assignment so the superseded run doesn't count against velocity.

PR #645, opened 2026-07-30
fix(auth): sync logout across tabs and stop the sign-in redirect loop (#606)

### LINKED ISSUE #606: Broken State Across Logouts
### Description

Currently, when you have multiple tabs for the ztmf tool open, and you log out from one of them, all other tabs will break by doing a stuttering loop. Logout should clear state across whole browser, not just tab or window.

### Steps to Reproduce

1. Navigate to `dev.ztmf.cms.gov` in one tab.
2. Open a new tab and navigate to `dev.ztmf.cms.gov`
3. On the first tab, navigate to the "log out" in the top right corner
4. On the second tab, try navigating somewhere. 

### Expected Behavior

When logging out in out tab, all tabs/sessions are logged out. 

### Actual Behavior

Tab #2 becomes stuttering and lops.

### Environment

Dev

### Acceptance Criteria

### Severity

None

### Affected Component

_No response_

### Screenshots

_No response_

### Logs / Error Messages

```shell

```

### Workaround

_No response_

costacalvin:
## Summary

Closes #606

With multiple tabs open, logging out in one tab left the others stuttering between the dashboard and the sign-in page instead of signing out.

Two independent causes. First, nothing told the other tabs: auth lives in HTTP-only cookies, so `POST /auth/logout` clears them browser-wide, but `handleLogout` only reloaded its own tab. Every other tab kept rendering the dashboard off in-memory loader data.

Second, the stutter itself was stale root loader data, and the obvious fix feeds it. Navigating to `/signin` leaves the root route matched, and react-router does not re-run a still-matched route's loader — so `loaderData` kept reporting `status === 200`, `LoginPage` trusted it and `<Navigate>`d back to the dashboard, whose mount fetches 401'd again. The interceptor's existing `router.navigate(SIGNIN)` could never break that cycle.

Frontend-only fix. The cross-tab signal is new; the loop fix corrects how the existing auth interceptor recovers stale loader state.

## Changes

* **Cross-tab logout signal.** New `src/utils/sessionSync.ts` owns a `BroadcastChannel('ztmf-auth')`. `handleLogout` calls `broadcastLogout()`; each tab registers `initLogoutListener()` once in `main.tsx`. Best-effort and fired regardless of the logout POST's result, matching the existing "even if the request fails we still drop to sign-in" contract.
* **Receiver reloads only.** It does not rewrite the hash and has no "already on `/signin`" skip. `Title` renders `LoginPage` on any route once the loader reports no session, so a hash write buys nothing — and since the questionnaire's `beforeunload` guard can veto the reload, writing the hash first would navigate the tab away even when a user cancels to rescue unsaved edits. The skip was also a race: a tab can sit on `/signin` mid-bounce with stale loader data and would stay wedged.
* **Loop fix: revalidate stale loader data.** `handleAuthError` now calls `router.revalidate()` when root `loaderData` still claims an active session, before navigating. The `isRevalidationRequired` flag survives the interrupting navigation, so the `/signin` navigation re-runs the authLoader and commits against fresh data (verified against the react-router 7.18.0 source — the version `package.json` and `yarn.lock` pin — and pinned by the integration test below rather than left as an assumption). Browser-independent, so it also covers plain session expiry and keeps browsers without `BroadcastChannel` from stuttering.
* **Redirect dedupe.** A 401 no longer re-navigates when the tab is already on (or navigating to) sign-in, so a burst of concurrent 401s cannot restart the navigation repeatedly, and a late 401 cannot clobber an already-showing `NO_ACCOUNT` terminal screen. Revalidation is likewise skipped while one is in flight.
* **`403 ACCOUNT_NOT_PROVISIONED` gets the same revalidation, and its own dedupe.** Scope stretch, flagged deliberately: it is the identical defect via the sibling trigger (an account deactivated mid-session stutters the same way), and leaving it latent three lines below the fix seemed worse. Its dedupe guard is deliberately narrower than the 401's — keyed on the sign-in location already carrying `reason === NO_ACCOUNT` **and the same message**, rather than on the route alone, so it stops NO_ACCOUNT re-navigating over itself while still overriding a stale EXPIRED redirect in either arrival order. Reusing the 401 guard here would have let EXPIRED suppress the more specific diagnosis. The message is part of the key because `LoginPage` prefers `location.state.message`: a burst whose first 403 arrived with no body falls back to the generic permission copy, and a reason-only comparison would then suppress a later 403 carrying the real "account is not set up" text, freezing the terminal screen on the wrong copy.
* **Channel construction and `postMessage` are both wrapped in try/catch.** Construction is hardening — the constructor is not specified to throw — but this module builds the channel at import time and is imported at the top of `main.tsx`, so a throw would white-screen the app before it renders, strictly worse than the bug being fixed; `'BroadcastChannel' in window` proves the interface exists, not that constructing it succeeds. `postMessage` is the one that genuinely is spec'd to throw (`InvalidStateError`, `DataCloneError`), and `handleLogout` calls it after the session cookie is already dead but *before* the hash change and reload — so an escaping throw would strand the user on a dashboard they are no longer signed in to. Both degrade to the same no-op path as an absent API.

## Why BroadcastChannel

It is the purpose-built API for this: same-origin only, delivers to every other tab but never the sender, and messages are ephemeral. Nothing sensitive crosses it — the payload is the string `logout`, and the session cookie stays `httpOnly` and untouched. Nothing is written to disk either, so the signal itself leaves no residue behind. To be precise about the scope of that last point: it is a claim about this channel, not about the app as a whole — questionnaire drafts are still persisted (encrypted) by `draftStore` and cleared only when a different user signs in. Pre-existing and untouched here; filing it separately.

The cost is browser support: it needs Safari 15.4+ (Chrome, Edge and Firefox have had it for years). Older browsers do not get the proactive teardown, but they are not left broken — the revalidation fix above is plain router state and works everywhere, so those tabs sign out cleanly at their next request instead of instantly. The channel is feature-guarded, so a browser without it no-ops rather than throwing.

## Acceptance criteria

* [x] Logging out in one tab signs out all tabs — with one deliberate exception: a tab holding unsaved questionnaire edits can veto its own reload through the browser's `beforeunload` dialog, and keeps showing the dashboard until its next request. Protecting unsaved work wins over instant teardown; that tab still signs out cleanly a moment later.
* [x] No stuttering loop after a cross-tab logout.
* [x] No stuttering loop on plain session expiry (no logout involved).
* [x] Browsers without `BroadcastChannel` still sign out cleanly, just not proactively.
* [x] Unsaved questionnaire edits are not silently discarded by a cross-tab logout.
* [x] `NO_ACCOUNT` terminal state still wins over a concurrent `EXPIRED` redirect.
* [x] Verified manually on DEV (see below).

Verified on DEV, both code paths separately, since the cross-tab signal and the loop fix are independent:

* **Cross-tab signal.** Logged out in one tab; the other tabs signed out too. No stutter.
* **Loop fix.** Deleted the session cookies in DevTools (no logout, so no broadcast — the same path a plain expiry or an older browser without `BroadcastChannel` takes), then clicked into a system: routed to the sign-in page cleanly, once.

Not walked manually, covered by code and tests instead: the unsaved-questionnaire-edits prompt and the `NO_ACCOUNT` terminal state.

## Testing

* **`authInterceptor.integration.test.ts` (new) — drives a real `createMemoryRouter`, not a mock.** This is the one that actually proves the loop stops. The mocked suite can only assert that `revalidate()` is called and ordered before `navigate()`; with no router behind the mock there is nothing to loop. It also matters because the fix leans on react-router's `isRevalidationRequired` carry-over, which is an implementation detail with no public contract — a minor version bump could regress #606 with every mocked test still green. These tests assert the observable outcome instead: after a 401 against stale loader data, the sign-in navigation commits with **fresh** root loader data, and concurrent bursts (homogeneous and mixed 401/403) settle on sign-in with a bounded number of loader runs rather than one per rejection. Litmus-verified: removing `revalidate()` fails them with the stale `status: 200`, which is exactly the value LoginPage trusted when it bounced; removing the dedupe guards takes the burst to one loader run per rejection.
* `sessionSync.test.ts` (new): `broadcastLogout` posts; a received `logout` reloads without touching the hash; it reloads *even when already on `/signin`* (the race above); non-logout messages ignored; repeat `initLogoutListener()` calls are idempotent; a failing `postMessage` is swallowed so the logging-out tab still finishes; both functions are safe no-ops when `BroadcastChannel` is absent **and** when constructing it throws. The fake channel implements `addEventListener`/`dispatch` as well as `onmessage`, so the idempotency test fails for the right reason (two reloads from stacked handlers) rather than vacuously.
* `authInterceptor.test.ts`: revalidate when loader data is stale, not when it already reports no session, not while one is in flight, still revalidate when either redirect is suppressed, ordered before navigate, plus burst / trailing-slash / casing cases. Eight cases pin the NO_ACCOUNT branch from both directions — overriding a committed *and* a pending EXPIRED redirect, not re-navigating over a showing or pending NO_ACCOUNT, correcting the copy when the message differs while still deduping an identical one, and a late 401 not clobbering a showing NO_ACCOUNT. Litmus-verified: removing the guard fails the three dedupe cases while the override cases keep passing. Every pre-existing branch (`skipAuthHandling`, `FORBIDDEN_ORIGIN`, plain 403, non-auth, network error) still asserted unchanged.
* `Title.test.tsx`: `broadcastLogout` fires on logout, including when the logout POST rejects.
* `yarn jest`: full suite green; `eslint` clean on every touched file; `tsc --noEmit` reports nothing in these files (the `EmailModal` / `QuestionnairePage` JSX-type errors are pre-existing on `main` from the `react-is` v19 vs React 18 mismatch).

### DIFF
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -10,9 +10,14 @@ import router from '@/router/router'
 import { SIGN_IN_GREETING } from '@/locales/en'
 import '@/sass/style.scss'
 import onPerfEntry from './utils/onPerfEntry'
+import { initLogoutListener } from '@/utils/sessionSync'
 import { SnackbarProvider } from 'notistack'
 // IIFE that initializes the root node and renders the application.
 ;(async function () {
+  // Cross-tab logout sync (#606): register once for this tab's lifetime so a
+  // logout in any other tab drops this one cleanly onto the sign-in page.
+  initLogoutListener()
+
   // create the root element in the DOM
   const rootElement = document.getElementById('root') as HTMLElement
 
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -14,8 +14,7 @@ import { initLogoutListener } from '@/utils/sessionSync'
 import { SnackbarProvider } from 'notistack'
 // IIFE that initializes the root node and renders the application.
 ;(async function () {
-  // Cross-tab logout sync (#606): register once for this tab's lifetime so a
-  // logout in any other tab drops this one cleanly onto the sign-in page.
+  // Once per tab lifetime, outside React so StrictMode cannot double-invoke it.
   initLogoutListener()
 
   // create the root element in the DOM
--- a/src/utils/authInterceptor.integration.test.ts
+++ b/src/utils/authInterceptor.integration.test.ts
@@ -1,0 +1,263 @@
+/**
+ * The #606 loop fix against a REAL router. authInterceptor.test.ts mocks the
+ * router, so it can pin that revalidate() runs before navigate() but not that
+ * the loop stops. That matters because the fix relies on react-router keeping
+ * its internal isRevalidationRequired flag across the interrupting navigation -
+ * public behavior of a private detail, so a version bump could regress #606
+ * with every mocked test still green. These assert the outcome instead: the
+ * /signin navigation commits with FRESH loader data.
+ */
+import type { AxiosError } from 'axios'
+import { createMemoryRouter } from 'react-router-dom'
+import { AuthCodes, SignInReasons } from '@/utils/authCodes'
+import { Routes, RouteIds } from '@/router/constants'
+
+// The interceptor imports the router singleton; point that import at whichever
+// memory router the current test built. Getters resolve lazily, so it stays live.
+const mockRouterRef: { current: ReturnType<typeof createMemoryRouter> | null } =
+  { current: null }
+
+jest.mock('@/router/router', () => ({
+  __esModule: true,
+  default: {
+    get state() {
+      return mockRouterRef.current?.state
+    },
+    navigate: (...args: unknown[]) =>
+      (
+        mockRouterRef.current?.navigate as unknown as (
+          ...a: unknown[]
+        ) => Promise<void>
+      )(...args),
+    revalidate: () => mockRouterRef.current?.revalidate(),
+  },
+}))
+
+jest.mock('@/utils/notify', () => ({
+  __esModule: true,
+  notify: jest.fn(),
+  markAuthHandled: (e: object) => Object.assign(e, { __authHandled: true }),
+}))
+
+import { handleAuthError } from './authInterceptor'
+
+// Mirrors authLoader's discriminated return.
+let sessionAlive = true
+let rootLoaderRuns = 0
+
+function rootLoader() {
+  rootLoaderRuns++
+  return sessionAlive
+    ? { status: 200, response: { userid: 'u1' } }
+    : { ok: false, reason: SignInReasons.EXPIRED, response: {} }
+}
+
+async function buildSignedInRouter() {
+  sessionAlive = true
+  rootLoaderRuns = 0
+  const router = createMemoryRouter(
+    [
+      {
+        id: RouteIds.ROOT,
+        path: Routes.ROOT,
+        loader: rootLoader,
+        children: [
+          { index: true, id: 'home' },
+          { path: RouteIds.SIGNIN, id: RouteIds.SIGNIN },
+        ],
+      },
+    ],
+    { initialEntries: [Routes.ROOT] }
+  )
+  mockRouterRef.current = router
+  router.initialize()
+  await waitForIdle(router)
+  expect(router.state.loaderData[RouteIds.ROOT]).toMatchObject({ status: 200 })
+  return router
+}
+
+/**
+ * Settle the router. It yields before checking, so `expectWorkStarted` asserts
+ * something was actually in flight - otherwise an interceptor that ever awaited
+ * before navigating would make the loader-count assertions vacuously pass.
+ */
+async function waitForIdle(
+  router: ReturnType<typeof createMemoryRouter>,
+  expectWorkStarted = false
+): Promise<void> {
+  if (expectWorkStarted) {
+    const busy =
+      router.state.navigation.state !== 'idle' ||
+      router.state.revalidation !== 'idle'
+    if (!busy) {
+      throw new Error(
+        'expected a navigation or revalidation to be in flight, but the router was idle'
+      )
+    }
+  }
+  for (let i = 0; i < 50; i++) {
+    await Promise.resolve()
+    await new Promise((r) => setTimeout(r, 0))
+    if (
+      router.state.initialized &&
+      router.state.navigation.state === 'idle' &&
+      router.state.revalidation === 'idle'
+    ) {
+      return
+    }
+  }
+  throw new Error('router did not settle')
+}
+
+function authError(
+  status: number,
+  code: string
+): AxiosError<{ error?: string; code?: string }> {
+  return {
+    config: {},
+    response: {
+      status,
+      data: { code },
+      statusText: '',
+      headers: {},
+      config: {},
+    },
+    isAxiosError: true,
+    toJSON: () => ({}),
+    name: 'AxiosError',
+    message: 'mock',
+  } as unknown as AxiosError<{ error?: string; code?: string }>
+}
+
+afterEach(() => {
+  // Optional across versions, hence the guard.
+  mockRouterRef.current?.dispose?.()
+  mockRouterRef.current = null
+})
+
+test('a 401 against a signed-in dashboard lands on sign-in with FRESH loader data (#606)', async () => {
+  const router = await buildSignedInRouter()
+
+  // Session dies out from under the tab.
+  sessionAlive = false
+  const runsBefore = rootLoaderRuns
+
+  await expect(
+    handleAuthError(authError(401, AuthCodes.UNAUTHORIZED))
+  ).rejects.toMatchObject({ __authHandled: true })
+  await waitForIdle(router, true)
+
+  expect(router.state.location.pathname).toBe(Routes.SIGNIN)
+  // The load-bearing assertion: without revalidate() this still reads 200.
+  expect(rootLoaderRuns).toBeGreaterThan(runsBefore)
+  expect(router.state.loaderData[RouteIds.ROOT]).toMatchObject({
+    ok: false,
+    reason: SignInReasons.EXPIRED,
+  })
+  expect(router.state.loaderData[RouteIds.ROOT]).not.toMatchObject({
+    status: 200,
+  })
+})
+
+test('the sign-in navigation carries the expired reason for LoginPage', async () => {
+  const router = await buildSignedInRouter()
+  sessionAlive = false
+
+  await expect(
+    handleAuthError(authError(401, AuthCodes.UNAUTHORIZED))
+  ).rejects.toMatchObject({ __authHandled: true })
+  await waitForIdle(router, true)
+
+  expect(router.state.location.state).toMatchObject({
+    reason: SignInReasons.EXPIRED,
+  })
+})
+
+test('a burst of 401s settles on sign-in once, without a redirect storm (#606)', async () => {
+  const router = await buildSignedInRouter()
+  sessionAlive = false
+  const runsBefore = rootLoaderRuns
+
+  // What a dashboard mount does: several requests rejecting together.
+  await Promise.all(
+    Array.from({ length: 4 }, () =>
+      handleAuthError(authError(401, AuthCodes.UNAUTHORIZED)).catch(() => {})
+    )
+  )
+  await waitForIdle(router, true)
+
+  expect(router.state.location.pathname).toBe(Routes.SIGNIN)
+  expect(router.state.loaderData[RouteIds.ROOT]).toMatchObject({ ok: false })
+  // Empirical bound for a homogeneous burst: lands on exactly 2, and goes to
+  // 5 without the guards. A mixed burst reaches 3 - separate test below.
+  expect(rootLoaderRuns - runsBefore).toBeLessThanOrEqual(2)
+})
+
+test('a 403 ACCOUNT_NOT_PROVISIONED burst settles on the NO_ACCOUNT terminal state', async () => {
+  const router = await buildSignedInRouter()
+  sessionAlive = false
+  const runsBefore = rootLoaderRuns
+
+  await Promise.all(
+    Array.from({ length: 4 }, () =>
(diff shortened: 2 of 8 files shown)


jsos3-cms (changes requested):
Nice work on this one. The diagnosis is right, the fix goes after the real cause instead of the symptom, and the PR body describes the react-router behavior accurately — I checked it line by line against the installed 7.18.0 source and it holds up.

I am requesting changes for **one** thing: the 403 path is missing a guard that the 401 path got. Everything else below is follow-on work, not a blocker. Happy to open those tickets or do the work myself — just say which.

## The one change I am asking for

The 401 path now skips the redirect when the tab is already on, or already heading to, the sign-in route. The 403 `ACCOUNT_NOT_PROVISIONED` path does not. It calls `router.navigate` every time.

That matters when a user is soft-deleted mid-session. A dashboard mount fires several requests at once, so every one of them comes back 403, and each one calls `navigate(SIGNIN, { replace: true })`. Each call aborts the previous one's loader run. It does settle — the last one wins and the end state is the same — so this is not a hang and not a regression. But it is the same "a burst of errors keeps restarting the navigation" problem you fixed three lines above, still sitting on the other trigger. Since the PR already reached over to cover this branch's revalidation, covering its dedupe too seems like the consistent call.

One thing to be clear about: I am **not** asking you to reuse the `alreadyOnSignIn` guard as-is here. That would break the "NO_ACCOUNT beats EXPIRED" behavior, and that behavior is correct — it should stay. What I am suggesting is a narrower guard: skip only when the tab is already on sign-in **and** the current or pending location state already says `reason === NO_ACCOUNT`. That stops NO_ACCOUNT from re-navigating over itself, while still letting it override a stale EXPIRED redirect from the same burst, in either arrival order.

## What I verified

**The revalidate-then-navigate ordering really is guaranteed, not just a happy accident.** `revalidate()` sets an internal `isRevalidationRequired` flag to true, and only `completeNavigation` clears it — a plain `navigate()` never touches it. So the flag survives the navigation that interrupts the revalidation, and the router then re-runs the still-matched root loader because of it. I also confirmed this by running a small `createMemoryRouter` probe: after `revalidate()` then `navigate('/signin')`, the committed loader data shows the fresh expired session, not the stale `status: 200`. Your test that asserts the call order is pinning something real.

**The in-flight guard cannot get stuck.** `revalidate()` flips `revalidation` to `'loading'` synchronously, before it awaits anything, so a same-tick burst correctly skips. `completeNavigation` puts it back to `'idle'`. Those are the only two values it ever takes, so the test's model of it is exact. Both guards are recomputed from `router.state` on every call, so there is no new module state that could wedge.

**No risk of the revalidation looping on itself.** `authLoader` sends `/users/current` with `skipAuthHandling: true`, so its own 401 cannot re-enter `handleAuthError`. That was the obvious way this could have gone wrong and it is already covered.

**StrictMode is fine.** `initLogoutListener()` runs in the top-level IIFE before render, outside anything StrictMode double-invokes. And `channel.onmessage = fn` overwrites rather than stacking up, so a repeat call would be harmless anyway.

**Nothing sensitive goes over the channel.** The payload is the string `logout`, and the same-origin limit is enforced by the browser rather than asserted by us. I grepped for auth material in `localStorage`/`sessionStorage` and found none — the session stays in httpOnly cookies.

**Your test numbers check out.** Full suite is 693 passed, 3 skipped, 0 failed. `eslint` is clean on all six touched files.

## Follow-ons — happy to take any of these

**1. Nothing in CI actually proves the loop stops.** The interceptor tests mock `@/router/router` completely. They do assert that `revalidate()` is called and that it comes before `navigate()`, and they would fail if the code dropped or reversed that — the branch coverage is genuinely good. But with no real router behind the mock, they cannot show that the loop breaks. Right now that rests on source-reading and your manual DEV check. It also rests on `isRevalidationRequired`, which is a react-router internal with no public contract behind it. So a minor version bump could change the behavior, the bug could come back, and every test here would still pass. One integration test against `createMemoryRouter` covering just the loop scenario would close that. I think this is the most valuable of the follow-ons.

**2. "Logging out in one tab signs out all tabs" is not quite unconditional.** The receiver reloads, and the questionnaire's `beforeunload` handler fires on that reload — the browser's native "leave site?" dialog really does interrupt. If someone cancels it to save their edits, that tab keeps showing the dashboard until its next request 401s. Your reasoning for this tradeoff is sound and I agree with it; protecting unsaved work is the right priority. It is just the acceptance-criteria line that reads stronger than the behavior. A short qualifier there or in the `sessionSync.ts` comment would do it.

**3. Consider a `try`/`catch` around the channel construction.** `sessionSync.ts` builds the channel at module load, and it is imported at the top of `main.tsx` — so anything that throws there happens before the IIFE runs and takes down app startup entirely. That would be worse than the bug being fixed. Being straight about the risk though: I went looking for a case where the constructor actually throws and could not find one. MDN lists no exceptions, and the spec's constructor just sets the channel name. In an opaque origin it constructs fine and quietly delivers nothing, which is already the fallback you want. So this is hardening, not a known crash. The reason I would still add it is that your own comment names "locked-down enterprise builds" as a case the guard covers, and `'BroadcastChannel' in window` only proves the interface exists, not that constructing it succeeds. One line closes the gap.

## Small stuff, take it or leave it

Two things are true by inspection but not asserted: that calling `initLogoutListener()` twice is idempotent, and that a channel which fails to construct leaves both functions as no-ops. The second only matters if the `try`/`catch` above goes in.

`#606` shows up 10 times in comments across the diff — four in source, six in tests. The prose reads fine without the number, and `Closes #606` in the PR body already carries the link. That said, `main` has this pattern in roughly 48 files, so you are not out of step with the repo.

`channel.onmessage =` instead of `addEventListener` is the right property for one caller. Worth a note in the comment if the channel is ever expected to carry more message types, since a second consumer would silently clobber the first.

Last one, and it is out of scope — I mention it only because the PR's kiosk framing invites the comparison. "Nothing is written to disk, so there is no logout residue" is accurate about the channel. But the app does persist questionnaire drafts to `localStorage` and IndexedDB in `draftStore.ts`, and those only get cleared when a *different* user signs in. So a shared workstation does keep draft answers after logout. Pre-existing, nothing to do with this PR — just flagging so the claim is not read more broadly than you meant it.

costacalvin:
@jsos3-cms Thanks for the detailed review — all addressed across three commits. The 403 `ACCOUNT_NOT_PROVISIONED` branch now has its own dedupe guard, narrower than the 401's as you suggested, so NO_ACCOUNT still overrides a stale EXPIRED redirect in either arrival order; I folded the message into the key too, since otherwise a burst whose first 403 has no body falls back to the generic copy and then suppresses a later one carrying the real "account is not set up" text. All three follow-ons are taken: `authInterceptor.integration.test.ts` drives a real `createMemoryRouter` and litmus-fails against the pre-fix code — writing it also surfaced that nothing pinned `revalidate()` outside the navigate guard, which would have regressed #606 with the suite still green — plus the qualified AC line for the `beforeunload` veto, and `try/catch` on both construction and `postMessage`, the latter being the one actually spec'd to throw.

jsos3-cms (approved):
Approving. The narrow guard is what I was after, and it is the right shape: keying on `reason` means a committed or pending EXPIRED redirect no longer suppresses the more specific NO_ACCOUNT diagnosis, in either arrival order. Keeping `revalidate()` outside the guard matters just as much, since a suppressed redirect still needs the stale loader data corrected, and the new assertion pins exactly that.

Folding the message into the key reads as a good call rather than scope creep. It only ever makes the skip narrower, so it cannot swallow a NO_ACCOUNT diagnosis, and it handles the body-less-first sequence you describe. One thing worth remembering: the key now depends on a backend string, so if the middleware ever varies the `ACCOUNT_NOT_PROVISIONED` message per request, the dedupe quietly degrades to one navigate per rejection. That lands back on the pre-fix behavior rather than anything worse, so it is not a concern today, just a coupling nothing on the frontend pins.

The integration test is the part I am most glad to see. It drives a real router, asserts the committed loader data is fresh and explicitly not the stale `status: 200`, and I checked it is not vacuous: reverting the 403 guard to the blanket check fails five tests, and removing `revalidate()` fails five of the six integration cases. The behavior is pinned in CI now instead of resting on a reading of react-router internals, which was the whole point. Good catch on the gap writing it surfaced, that nothing had covered `revalidate()` sitting outside the navigate guard.

Two small things, neither worth another round. The `runs <= 3` bound in the mixed-burst test also passes when runs go down, so on its own it cannot detect under-revalidation; its sibling tests cover that direction, so coverage is fine overall. And `#606` still appears in seven comments, four of them in the new integration test file, which is the one place here it is freshly authored.

Verified on my side at `98ec119`: 709 passed, 3 skipped, 0 failed; `eslint` and `tsc --noEmit` both clean; Snyk and the deploy gate green.

costacalvin:
Merge is blocked due to "Missing successful active dev deployment". I attempted to resolve this by re-running all jobs such as the ui/Deploy to the Dev environment. The deployment is currently active, however the merge is still blocked with the same requirement. This leads me to believe it is evaluating on a stale state, and although the deployment is active it did not cause a re-evaluation by the Github merge workflow. The PR's updated_at is still 2026-07-30T20:00:42Z — the moment of the approval. Nothing has touched it since, even though my re-run deployed successfully at 13:40 today which also leads me to believe it is using a stale state. 

I will attempt to remedy this by converting the PR to a draft then opening the PR again, hopefully triggering a re-evaluation. If that doesn't work, we may just need a re-approval.

costacalvin:
That sadly did not work. The head commit 98ec119 has a deployment to it whose status chain is in_progress → success with no inactive marker, and it's the newest deployment in that environment. Approved on the head commit, all three checks green, all commits signature-verified, no review threads. GitHub still says "Missing successful active dev deployment." https://github.com/CMS-Enterprise/ztmf-ui/deployments

Will attempt to rebase the PR off of main to get it up to date. Hopefully a new commit/approval will suffice and cause a re-evaluation.

costacalvin:
Merging main into the branch worked to unblock me from the "Missing successful active dev deployment" issue.

PR #437, opened 2026-06-26
Enhancement/batch opdiv grants

costacalvin:
## Summary

Closes #395. Depends on #432 (must merge first — this branch is rebased onto `enhancement/opdiv-idp-fields-selectable`).

Replaces the per-OpDiv POST/DELETE loop in `OpDivGrantModal` with a single `PUT /users/{userid}/opdivs` batch request. The backend reconciles the full desired set against current grants in one transaction and re-derives `identity_provider` once.

- **`userOpdivs.ts`**: removes `grantOpDiv` and `revokeOpDiv`; adds `setUserOpDivs(userid, opdivIds[])` → `PUT /users/{userid}/opdivs`
- **`OpDivGrantModal`**: replaces immediate per-toggle writes with local state + Save/Cancel buttons; modal only closes on success; OPDIV_ADMIN scope filter strips out-of-scope grants from the PUT body before sending (prevents backend 403 when the target user holds grants from another admin outside the caller's scope)
- **`UserTable`**: wires `setUserOpDivs` into the new-user create flow (replaces per-item `Promise.allSettled` loop from #432); fixes `handleCloseOpDivModal` to not trigger a redundant refresh on cancel

## Acceptance criteria

- [x] `OpDivGrantModal` has Save and Cancel buttons replacing the old single Close button
- [x] Changes are applied in one batch request on Save, not on each toggle
- [x] Modal stays open on error; closes only on success
- [x] OPDIV_ADMIN cannot trigger a 403 by saving a user who holds out-of-scope grants from another admin
- [x] Creating a new user with OpDivs selected uses the batch endpoint (one PUT)
- [x] Cancelling the modal fires no network requests
- [x] OpDivGrantModal saves all grant add/remove changes in one request via the new batch endpoint
- [x] Resulting grants and the post-save identity_provider refresh are unchanged from today
- [x] No per-OpDiv request loop remains in the modal

## UI
OpDivGrantModal now uses a save button in conjunction with the batch OpDiv grant network call.

image

Network request + payload `{"opdiv_ids":[7,14,21]}`

image

## Test plan

- [ ] Open Assign OpDivs modal for a user, toggle several OpDivs, click Save — verify single `PUT /opdivs` in network tab (204)
- [ ] Open modal, make no changes, click Cancel — verify no network requests fire
- [ ] Open modal, click Save — verify modal closes on success and the OpDivs column updates
- [ ] Verify error state: if PUT fails, modal stays open and shows error snackbar
[...]

PR #646, opened 2026-07-30
chore(#578): retire the hasScore score-presence proxy from ProgressCell

### LINKED ISSUE #578: Cleanup: retire the hasScore proxy from ProgressCell once ztmf#437 is deployed everywhere
## Context

## When

## What to remove

## Acceptance criteria

Refs #546, #567, ztmf#437

voidspooks:
Retires the score-presence proxy left in `ProgressCell` as graceful degradation for the ztmf#437 deploy window. `questionsanswered` becomes required on `ScoreProgress` since a #437 backend always returns it (`int32`, `COALESCE(..., 0)`, no `omitempty`), which is what lets the null-fallback branches be deleted rather than stranded as dead code. Past-call progress now renders and sorts solely from `questionsanswered` vs `questionsexpected`. Beyond the ticket's file list, `progressSortValue` carried the same proxy-era fallback (`answered == null` ranks as complete) and is retired in the same pass. Visible change: a past call with zero answers reads 0/N Incomplete instead of an em-dash. Assumes #437 is deployed everywhere per the ready-to-work move; flag if prod is still pending. `tsc`, `eslint`, and jest (52 suites / 677 tests) green locally. Snyk and deploy checks will show red as usual for fork PRs.

Closes #578

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -314,9 +314,8 @@ export type ScoreProgress = {
   questionsexpected: number
   // Distinct applicable questions with an answer at all, any status (ztmf#437).
   // The completion signal a closed call needs: imported/carried answers are
-  // answered but never "updated this cycle". Optional so the UI degrades to the
-  // prior score-presence proxy until the backend field is deployed.
-  questionsanswered?: number
+  // answered but never "updated this cycle".
+  questionsanswered: number
   questionsupdated: number
   lastupdatedat?: string | null
   updatedsincestart: boolean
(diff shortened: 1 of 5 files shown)


jsos3-cms:
@voidspooks the rehome shipped. Your change went in via #654, which was folded into the frontend code-freeze batch #674 and merged as `1d96160`. Your commit is in it with your authorship preserved, and #578 is closed.

Closing this now, as promised when it was rehomed: it was held open as a draft so it could not merge as a duplicate while the in-repo copy was in flight.

PR #647, opened 2026-07-30
fix(#639): call-scoped dashboard toggles, past-call graying, and a Data Call column

### LINKED ISSUE #639: disable "Not updated only" toggle for closed data calls
### Description

### Root cause

### Steps to Reproduce

1. Log in as an admin/owner.
4. See no rows returned

### Expected Behavior

### Actual Behavior

### Environment

Prod, Dev

### Acceptance Criteria

### Optional AC:

### Severity

### Affected Component

FismaTable.ttsx

### Screenshots

Image

Image

### Logs / Error Messages

```shell

```

### Workaround

_No response_

voidspooks:
Fixes the reported bug and implements both optional ACs plus the design agreed in the Slack feedback thread.

The bug: "Not updated only" is a current-cycle signal, and once the newest call's deadline passes it silently filtered every row and emptied the grid. Both call-scoped toggles ("Not updated only" and the new "Open data call only") now disable with an explanatory tooltip whenever the open call is not part of what the table is showing, and a slim banner (`role=status`) explains the state: "No open data call; showing each system's most recently updated call", or "The open data call is not in the selected view" when a call is open but the year picker sits on a historical group. A stored-on filter value is neutralized when the view changes, so a year switch or a refetch that closes the newest call cannot leave an invisible filter emptying the grid.

The gate is deliberately "the open call is in view", not "a call is open": the year picker can select a historical group while a newer call is open, and in that view no row is current, so a live call-scoped filter would empty the grid and graying would wash out every row. The derivation lives in `isOpenCallInView` with tests, including the pre-load `latestDataCallId = 0` case.

One divergence from the optional AC as written: "Open data call only" ships default OFF. Default-on would hide closed-call rows silently as the default view (the same failure mode as the bug), drop them from the row-selection export, and disagree with the stat tiles, which count all systems. It is an opt-in focus filter, and flipping it counts as an active filter for Clear filters.

Beyond the ticket text, per the Slack thread: rows displaying a closed call are de-emphasized while the open call is in view, and a new Data Call column names each row's displayed call so the state is text, sortable, and quick-searchable rather than color-only. De-emphasis is a background tint plus muted cell text, deliberately not whole-row opacity, which pushed the name link, checkbox, and action icons below WCAG contrast minima while leaving them clickable. The column sorts by call deadline, not name (`datacallNameComparator`, tested; names do not sort chronologically), and carries a header tooltip explaining the resolution. A new `resolveRowCallId` helper is the single source of truth for which call a row is displaying, shared by the column and the pillar-scores modal so they cannot disagree.

Known limitation, accepted for now: the disabled switches' tooltips are hover-only and the column header tooltip is a native `title` attribute, so keyboard and screen-reader users rely on the banner, which carries the same information. Worth folding into a broader a11y pass.

Visible behavior for review: the new column, the graying in mixed views, the disabled toggles plus banner between calls.

closed
focus
mixed
notupdated

Gate: `tsc` clean, `eslint` zero errors (2 pre-existing InsightsPanel warnings), jest 53 suites / 697 tests green. Snyk and deploy checks will show red as usual for fork PRs.

Closes #639

### DIFF
--- a/src/views/FismaTable/FismaTable.tsx
+++ b/src/views/FismaTable/FismaTable.tsx
@@ -53,7 +53,7 @@ import { hasSystemAccess } from '@/utils/userRoles'
 import { TIERS } from '@/utils/tierStyles'
 import { ProgressCell } from './progressColumn'
 import { progressSortValue } from './progressHelpers'
-import { resolveRowCallId } from './rowCall'
+import { resolveRowCallId, datacallNameComparator } from './rowCall'
 import { fetchOpDivs } from '@/utils/opdivs'
 import { toCategoryMap } from '@/utils/dataCenterEnvironments'
 import { parseDatacallName } from '@/utils/datacallGrouping'
@@ -62,6 +62,7 @@ import type { OpDiv } from '@/types'
 import {
   applyDashboardFilters,
   hasNoActiveFilters,
+  isOpenCallInView,
   EMPTY_DASHBOARD_FILTERS,
   type DashboardFilterState,
 } from './dashboardFilters'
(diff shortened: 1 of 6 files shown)


MackOverflow:
Moving this to #648 — fork PRs can't access the org secrets required for Snyk and the deploy CI gates. Your commits and authorship carry over unchanged. Thanks @voidspooks!

PR #648, opened 2026-07-30
fix(#639): call-scoped dashboard toggles, past-call graying, and a Data Call column

### LINKED ISSUE #639: disable "Not updated only" toggle for closed data calls

MackOverflow:
> **Note:** In-repo copy of #647 by @voidspooks. Moved from the fork so Snyk
>
> Refs #647

---

Fixes the reported bug and implements both optional ACs plus the design agreed in the Slack feedback thread.

The bug: "Not updated only" is a current-cycle signal, and once the newest call's deadline passes it silently filtered every row and emptied the grid. Both call-scoped toggles ("Not updated only" and the new "Open data call only") now disable with an explanatory tooltip whenever the open call is not part of what the table is showing, and a slim banner (`role=status`) explains the state: "No open data call; showing each system's most recently updated call", or "The open data call is not in the selected view" when a call is open but the year picker sits on a historical group. A stored-on filter value is neutralized when the view changes, so a year switch or a refetch that closes the newest call cannot leave an invisible filter emptying the grid.

The gate is deliberately "the open call is in view", not "a call is open": the year picker can select a historical group while a newer call is open, and in that view no row is current, so a live call-scoped filter would empty the grid and graying would wash out every row. The derivation lives in `isOpenCallInView` with tests, including the pre-load `latestDataCallId = 0` case.

One divergence from the optional AC as written: "Open data call only" ships default OFF. Default-on would hide closed-call rows silently as the default view (the same failure mode as the bug), drop them from the row-selection export, and disagree with the stat tiles, which count all systems. It is an opt-in focus filter, and flipping it counts as an active filter for Clear filters.

Beyond the ticket text, per the Slack thread: rows displaying a closed call are de-emphasized while the open call is in view, and a new Data Call column names each row's displayed call so the state is text, sortable, and quick-searchable rather than color-only. De-emphasis is a background tint plus muted cell text, deliberately not whole-row opacity, which pushed the name link, checkbox, and action icons below WCAG contrast minima while leaving them clickable. The column sorts by call deadline, not name (`datacallNameComparator`, tested; names do not sort chronologically), and carries a header tooltip explaining the resolution. A new `resolveRowCallId` helper is the single source of truth for which call a row is displaying, shared by the column and the pillar-scores modal so they cannot disagree.

Known limitation, accepted for now: the disabled switches' tooltips are hover-only and the column header tooltip is a native `title` attribute, so keyboard and screen-reader users rely on the banner, which carries the same information. Worth folding into a broader a11y pass.

Visible behavior for review: the new column, the graying in mixed views, the disabled toggles plus banner between calls.

closed
focus
mixed
notupdated

Gate: `tsc` clean, `eslint` zero errors (2 pre-existing InsightsPanel warnings), jest 53 suites / 697 tests green.

Closes #639

### DIFF
--- a/src/views/FismaTable/FismaTable.tsx
+++ b/src/views/FismaTable/FismaTable.tsx
@@ -53,7 +53,7 @@ import { hasSystemAccess } from '@/utils/userRoles'
 import { TIERS } from '@/utils/tierStyles'
 import { ProgressCell } from './progressColumn'
 import { progressSortValue } from './progressHelpers'
-import { resolveRowCallId } from './rowCall'
+import { resolveRowCallId, datacallNameComparator } from './rowCall'
 import { fetchOpDivs } from '@/utils/opdivs'
 import { toCategoryMap } from '@/utils/dataCenterEnvironments'
 import { parseDatacallName } from '@/utils/datacallGrouping'
@@ -62,6 +62,7 @@ import type { OpDiv } from '@/types'
 import {
   applyDashboardFilters,
   hasNoActiveFilters,
+  isOpenCallInView,
   EMPTY_DASHBOARD_FILTERS,
   type DashboardFilterState,
 } from './dashboardFilters'
@@ -475,11 +476,15 @@ export function QuickSearchToolbar(props: {
       {/* Toolbar-row captions crowd the filter cluster and wrap it (ui#639),
           so the call-scope notice renders as its own slim banner between the
           filters and the column headers. */}
+      {/* role="status": this is a standing contextual notice, not an alert —
+          MUI's default role="alert" would interrupt screen readers at mount
+          and on every year switch. It is also the keyboard/SR-reachable
+          counterpart of the disabled switches' hover tooltips. */}
       {!openCallInView && (
-        <Alert severity="info" sx={{ borderRadius: 0, py: 0 }}>
+        <Alert severity="info" role="status" sx={{ borderRadius: 0, py: 0 }}>
           {hasOpenCall
             ? 'The open data call is not in the selected view'
-            : "No open data call; showing each system's latest completed call"}
+            : "No open data call; showing each system's most recently updated call"}
         </Alert>
       )}
     </>
@@ -528,6 +533,10 @@ export default function FismaTable({
   // not-updated filter) only makes sense then; a past/closed call shows a
   // neutral Complete/Incomplete chip instead (ztmf#537). Rows without a chosen
   // call, or before latestDataCallId has loaded, keep the current rendering.
+  // Agrees with resolveRowCallId (the Data Call column) by construction:
+  // buildDashboardMaps fills chosenCallMap and systemCallMap for the same key
+  // set, so a row is grayed iff its column names a non-open call. Keep the
+  // two in step if either resolution changes.
   const isRowCurrentCall = useCallback(
     (fismasystemid: number): boolean => {
       const chosen = chosenCallMap[fismasystemid]
@@ -544,8 +553,11 @@ export default function FismaTable({
   // Without the in-view check the first toggle silently empties the grid (the
   // reported bug) and the second keeps only rows with no call data.
   const hasOpenCall = Boolean(latestDataCallId) && !latestDeadlinePassed
-  const openCallInView =
-    hasOpenCall && activeDatacallIds.includes(latestDataCallId)
+  const openCallInView = isOpenCallInView(
+    latestDataCallId,
+    latestDeadlinePassed,
+    activeDatacallIds
+  )
   const callById = useMemo(
     () => new Map(datacalls.map((d) => [d.datacallid, d])),
     [datacalls]
@@ -676,8 +688,11 @@ export default function FismaTable({
 
   // The call-scoped facets only mean something while the open call is in
   // view (their switches gray out otherwise). Drop any stored true when it is
-  // not, so a deadline passing mid-session or a switch to a historical year
-  // cannot leave an invisible filter emptying the grid (ui#639).
+  // not, so a year-picker move to a historical group, or a /datacalls refetch
+  // that flips the newest call closed, cannot leave an invisible filter
+  // emptying the grid (ui#639). (openness is evaluated when datacalls load,
+  // not continuously, so a deadline passing while the page sits open takes
+  // effect on the next fetch or reload.)
   useEffect(() => {
     if (openCallInView) return
     setFilters((prev) =>
@@ -865,7 +880,7 @@ export default function FismaTable({
       // Renders as the column-header tooltip: the resolution is not obvious
       // from the name (it is not simply "last completed call").
       description:
-        "The data call this row's score and progress are shown from; a system with no data in the selected calls shows the open call it is expected to answer.",
+        "The data call this row's score and progress are shown from: the system's most recently updated call among the selected calls, or the dashboard's active call for a system with no data in them.",
       width: 130,
       align: 'center',
       headerAlign: 'center',
@@ -879,9 +894,7 @@ export default function FismaTable({
             activeDataCallId
           )
         )?.datacall ?? '',
-      sortComparator: (a, b) =>
-        (deadlineByCallName.get(a as string) ?? 0) -
-        (deadlineByCallName.get(b as string) ?? 0),
+      sortComparator: datacallNameComparator(deadlineByCallName),
     },
     {
       // Questionnaire progress for the active data call (ztmf#299). The
@@ -1093,8 +1106,16 @@ export default function FismaTable({
           '& .MuiDataGrid-columnHeaders .MuiSvgIcon-root': {
             color: 'white',
           },
+          // De-emphasis must not dim interactive elements: whole-row opacity
+          // drops the name link, checkbox, and action icons below WCAG
+          // contrast minima while they stay clickable. Tint the row and mute
+          // only the cell text (rgba .6 on white holds AA); links, controls,
+          // and chips keep their own full-contrast colors.
           '& .past-call-row': {
-            opacity: 0.55,
+            backgroundColor: '#f5f5f5',
+          },
+          '& .past-call-row .MuiDataGrid-cell': {
+            color: 'rgba(0, 0, 0, 0.6)',
           },
         }}
       />
--- a/src/views/FismaTable/FismaTable.tsx
+++ b/src/views/FismaTable/FismaTable.tsx
@@ -476,17 +476,26 @@ export function QuickSearchToolbar(props: {
       {/* Toolbar-row captions crowd the filter cluster and wrap it (ui#639),
           so the call-scope notice renders as its own slim banner between the
           filters and the column headers. */}
-      {/* role="status": this is a standing contextual notice, not an alert —
-          MUI's default role="alert" would interrupt screen readers at mount
-          and on every year switch. It is also the keyboard/SR-reachable
-          counterpart of the disabled switches' hover tooltips. */}
-      {!openCallInView && (
-        <Alert severity="info" role="status" sx={{ borderRadius: 0, py: 0 }}>
-          {hasOpenCall
-            ? 'The open data call is not in the selected view'
-            : "No open data call; showing each system's most recently updated call"}
-        </Alert>
-      )}
+      {/* role="status" rather than MUI's default role="alert": this is a
+          standing contextual notice, and an assertive role would interrupt
+          screen readers at mount and on every year switch.
+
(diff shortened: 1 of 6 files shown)


a-dipietro:
@MackOverflow  one potential accessibility nit- disabled controls rely on hover tooltips; title attribute and hover-only tooltips are not reliably discoverable by keyboard/screen-reader users. If we're OK with that, LMK and I'll approve.

MackOverflow:
@a-dipietro Good catch, and you're right that the title/hover tooltips aren't reliably reachable by keyboard or screen reader. The reason the toggles are disabled isn't tooltip-only, though — whenever they're greyed out, a role="status" banner renders directly above the grid with the same wording ("The open data call is not in the selected view" / "No open data call; showing each system's most recently updated call"). As a polite live region it's announced to screen readers and sits in the DOM for content navigation, so the why is available without a mouse; the tooltips are just a redundant visual cue.

The gap that remains is the per-control association — a keyboard user can't focus the disabled switch itself to hear its specific reason, since the native disabled attribute drops it from the tab order. The clean fix there is aria-disabled + a no-op change handler (keeps it focusable so the tooltip fires on focus and can be wired via aria-describedby). I'd prefer to fold that into the broader a11y pass I flagged in the description rather than expand this bugfix PR, since the banner already covers the information accessibly today.

If you're good with that, I think we're OK to approve — happy to file a follow-up issue for the aria-disabled upgrade so it doesn't get lost.

jsos3-cms (approved):
Approving. The core fix is right: I traced every session flow I could think of, year switches, a `/datacalls` refetch that closes the newest call, the pre-load `0` sentinel, and could not construct a sequence that leaves a persistent invisible filter emptying the grid. `isOpenCallInView` is correct and the graying is properly gated so a historical view does not wash out every row.

There is one thing I would want in before merge. I wrote and verified the fix locally rather than pushing to your branch, so the diff is below for you to apply or adapt. Everything else here is follow-up.

## On the accessibility thread

@a-dipietro, @MackOverflow: mostly yes, with one correction that changes the answer.

The banner genuinely is discoverable by content navigation. The same `!openCallInView` condition gates both the banner and the disabled switches, so whenever a switch is greyed the banner is in the DOM to be browsed to. That half holds.

The announcement half did not. The banner renders as `{!openCallInView && <Alert role="status">...}`, so the live region is created at the same moment its text appears. A live region inserted into the DOM together with its content is generally not announced; the region has to already exist and have its contents change. So the polite-live-region behaviour offered as the reason the tooltips are only a redundant visual cue was not firing, in exactly the cases it exists to cover.

Choosing `status` over MUI's default `role="alert"` was the right call and worth keeping, since an assertive role really would interrupt on every year switch. The fix is just to keep the region mounted:

```tsx
<Box role="status" aria-live="polite">
  {!openCallInView && (
    <Alert severity="info" sx={{ borderRadius: 0, py: 0 }}>
      {hasOpenCall
        ? 'The open data call is not in the selected view'
        : "No open data call; showing each system's most recently updated call"}
    </Alert>
  )}
</Box>
```

With that in, I agree the remaining gap is the per control association, and `aria-disabled` plus a no-op handler is the right shape. Folding that into the broader accessibility pass rather than growing this PR seems right to me too.

## The one I would fix before merge

`resolveRowCallId`'s last resort is `activeDataCallId`, which is not guaranteed to be a call the user is looking at.

`activeDataCallId` is `selectedDatacall?.datacallid ?? latestDataCallId`, and `selectedDatacall` is null unless exactly one call is toggled on. Selecting a year toggles all of that year's calls on. So for any year with two or more calls, `activeDataCallId` silently becomes the open call even though the view is historical.

The sequence: pick a historical year with two calls; a system with no score rows in either is absent from both maps, since `buildDashboardMaps` only indexes systems that have score rows, but it still renders as a row; `resolveRowCallId` falls through both maps and returns the open call. The Data Call column then labels that row with the current call's name while the user is browsing a historical year. That needs only a multi-call year plus a system unscored that year, which is ordinary rather than exotic.

There is a sharper version. `PillarScoresModal` re-resolves independently: if no score matches the id it was handed, it falls back to `scoredDatacalls[0]`, the deadline-latest call the system has ever been scored in. That trigger is exactly the case above, so for a system with scores in some third call, the column can name one call while the modal header names another. That is what falsifies the "shared so they cannot disagree" claim. The modal's fallback predates this PR; the new column is what makes it visible.

Preferring the newest call actually in view fixes both, and keeps real row data winning over the view:

```ts
export function resolveRowCallId(
  fismasystemid: number,
  chosenCallMap: Record<number, number>,
  systemCallMap: Record<number, number[]>,
  datacalls: datacall[],
  activeDataCallId: number,
  activeDatacallIds?: number[]
): number {
  const chosen = chosenCallMap[fismasystemid]
  if (chosen != null) return chosen

  const scored = sortDatacallsByDeadline(
    (systemCallMap[fismasystemid] ?? [])
      .map((id) => datacalls.find((d) => d.datacallid === id))
      .filter((d): d is datacall => Boolean(d))
  )[0]?.datacallid
  if (scored != null) return scored

  if (activeDatacallIds?.length) {
    if (activeDatacallIds.includes(activeDataCallId)) return activeDataCallId
    const newestInView = sortDatacallsByDeadline(
      activeDatacallIds
        .map((id) => datacalls.find((d) => d.datacallid === id))
        .filter((d): d is datacall => Boolean(d))
    )[0]?.datacallid
    if (newestInView != null) return newestInView
  }
  return activeDataCallId
}
```

The new parameter is optional so nothing changes for a caller that does not pass the selection. Both call sites in `FismaTable.tsx`, `handleOpenPillarScores` and the column's `valueGetter`, then pass `activeDatacallIds` as the sixth argument.

The column header text needs a matching tweak: it currently describes the fallback as "the dashboard's active call", which is accurate for a single call selection and misleading for a multi call one. "or the newest selected call for a system with no data in them" matches the new behaviour.

## Worth taking while you are in here

The Data Call sort key. `deadlineByCallName` builds its map with a raw `new Date(d.deadline).getTime()`, bypassing the `deadlineMs` guard in `sortDatacallsByDeadline.ts` that exists to coerce a bad value to `-Infinity`. The comparator's `?? -Infinity` only catches a name absent from the map, not one present but mapped to `NaN`, and `NaN === NaN` is false so the equality shortcut misses it too. The result is a `NaN` comparator return, which is outside `Array.prototype.sort`'s documented contract and leaves that row's position up to the engine. Normalizing inside the comparator covers every caller:

```ts
const key = (name: string): number => {
  const value = deadlineByName.get(name)
  return value === undefined || Number.isNaN(value) ? -Infinity : value
}
```

That also makes the doc comment's claim about mirroring `sortDatacallsByDeadline`'s handling true again.

## Follow-ups, genuinely not blocking

**A one render flash to empty.** The neutralization `useEffect` runs after commit, not during render. On a year switch with a call-scoped toggle on, the render that flips `openCallInView` to false still computes `filteredRows` against the un-cleared `filters`, so the grid can paint empty for one commit before the effect corrects it. Transient and self healing, but structurally the same symptom as the original bug. Computing an effective filter value inline for `applyDashboardFilters`, while keeping the effect to clear the persisted state so the switches still uncheck, would make the invariant structural instead of eventual. I left it out of the diff above because it is a shape change to the filter flow rather than a localized fix.

**Action icons are muted.** `.past-call-row .MuiDataGrid-cell { color: rgba(0,0,0,0.6) }` cascades onto the three `GridActionsCellItem` icons, which all pass `color="inherit"`, so they do dim in past-call rows, contrary to the comment saying controls keep their own full contrast colors. The other two claims hold: the System Name link uses an inline style, which wins over a class rule, and the selection checkbox does not inherit. Contrast for the muted case is about 5.6:1, so it clears AA. This is a comment versus code discrepancy rather than a regression, so either exclude the actions cell or adjust the comment.

**No component level test.** `dashboardFilters.test.ts`, `rowCall.test.ts` and `QuickSearchToolbar.test.tsx` are solid, and I checked their assertions against the source rather than assuming. But `FismaTable.test.tsx` still only covers the download button, so nothing exercises the neutralization effect, the `getRowClassName` wiring, or the column's resolution inside the real component. Both the fallback issue above and the flash live at that wiring level, which is where the "cannot recur" proof does not currently reach.

## Verification

With the diff above applied locally: 705 passed, 3 skipped, 0 failed, which is the 697 on your head plus the eight tests I added for the new last-resort behaviour and the `NaN` sort key. `eslint` clean on the changed files and `tsc --noEmit` clean. I also confirmed against the Data Grid v6 docs that `valueGetter` receives the params object and `sortComparator` receives the `valueGetter` output, so the name-keyed comparator is the correct pairing, and that MUI's `Alert` does default to `role="alert"`, so the override was well reasoned.

Happy to push the commit to your branch if you would rather not retype it.

jsos3-cms (changes requested):
Converting my earlier approval to a request for changes. This is not a change of position, it is making the mechanics match what that review already said twice: the fallback fix in `resolveRowCallId` needs to be in before this merges.

The reason for the switch is that an approval does not carry a condition. Mine is currently the only review on this PR, so it satisfies the one-approval requirement and the last thing standing between this and `main` is the deployment gate, which we routinely clear with `--admin`. On this repo merging is the prod deploy, so an admin merge would ship the wrong-label behaviour straight to production. A request for changes is a real gate rather than a note someone has to remember to re-read.

Nothing new to look at. The finding, the reasoning, and a working diff are all in my earlier review: `resolveRowCallId`'s last resort is `activeDataCallId`, which collapses to the open call for any multi-call year, so the Data Call column can name a call that is not in the selected view, and `PillarScoresModal` can then re-resolve to a third call for the same row.

Happy to push the commit myself if you would rather not retype it, just say so. Otherwise re-request review once it is in and I will approve straight away. Everything else in that review stays follow-up, and the core filter fix is still correct as written.

voidspooks:
@jsos3-cms go ahead and push your verified commit, that beats a round trip through the fork and I appreciate you writing it up. If the always-mounted live region wrapper isn't already in it, please fold that in too since it's two lines and settles the a11y thread. I'll pick up the flash-to-empty, the icon comment discrepancy, and the component-level test as a follow-up issue. @MackOverflow heads up since it's your branch.

jsos3-cms (approved):
Clearing my earlier request for changes. The fallback fix is in and I have verified it on `cdd010f` rather than taking the commit message for it.

`resolveRowCallId` now prefers the newest call that is actually in the selection, with `activeDatacallIds` threaded to both call sites, the pillar-scores handler and the column's `valueGetter`, so the column and the modal can no longer name different calls for the same row. The comparator normalizes a present-but-`NaN` deadline to the same sentinel an unknown name gets, so it can no longer return `NaN` and leave a row's position up to the engine. And the `role="status"` region is now a permanently mounted wrapper with only its contents gated, so it can actually announce rather than being created at the same moment as its text.

I also rebased the branch onto `main`. It was three commits behind, and since dev deploys from the PR branch, anyone testing this on dev was getting a bundle without #642 and #645, which meant testing these dashboard changes against pre-fix auth behaviour. It is now current with no conflicts, the net diff is still only the six `src/views/FismaTable/` files, and @voidspooks remains the author on all four original commits.

Checks on the rebased head: FismaTable suites 70 passed, `tsc --noEmit` clean, `eslint` clean on the touched directory.

## Needs an approval from someone else

@costacalvin @a-dipietro, could one of you take this one? It needs a second pair of eyes rather than mine specifically.

`main` has `require_last_push_approval` on, which means the most recent push has to be approved by someone other than whoever pushed it. I pushed both the fix and the rebase, so my approval does not satisfy that rule, and @MackOverflow cannot approve as the author. That is a consequence of my applying the fix directly rather than handing back a diff, and it is on me.

The change is dashboard-only. It disables the two call-scoped toggles when the open data call is not in the selected view, de-emphasizes rows showing a closed call, and adds a Data Call column naming each row's source cycle. My full review is above if it is useful, and the follow-ups I raised there are non-blocking.

a-dipietro (approved):
lgtm, Jono, considering we already discussed the tooltips. Thanks!

voidspooks:
Filed the three non-blocking items from your review as #673: the one-paint flash to empty, the action-icon color cascade contradicting its own comment, and the missing component-level coverage in `FismaTable.test.tsx`. Confirmed all three still hold on current main after #665 reworked the file.

PR #650, opened 2026-07-31
Use fresh caller scope in the Assign OpDivs modal

### LINKED ISSUE #643: Assign OpDivs modal uses a stale caller scope: session-old grants can silently revoke a target's OpDiv on save
### Description

The Assign OpDivs modal decides what to preserve on save using the acting admin's own OpDiv grants, read from `userInfo.assignedopdivids`. That value comes from the route loader's `/users/current` response and is never refreshed for the rest of the session, so an `OPDIV_ADMIN` whose own grants change mid-session is working from a stale scope.

This matters more than it did before #642. That PR made `callerGrantIds` the save-time preserve boundary — the set that decides which of the target's grants survive the `PUT /users/{userid}/opdivs`. A stale value now has a data-integrity consequence, not just a cosmetic one.

Two directions, with different severity:

- **Caller lost a grant** (stale-wide): the PUT includes an id the caller no longer holds, and the backend's pre-gate rejects the whole request with a 403 (`usersopdivs.go`, the `IsAssignedOpDiv` loop before the tier-ceiling check). Loud failure, no data loss, but the admin sees an unexplained 403 on a modal they did not change.
- **Caller gained a grant** (stale-narrow): the new id is absent from `callerScope`, so the save strips it from the PUT. The backend still sees `IsAssignedOpDiv = true` for it, so it lands in `toRemove` and the target's grant is silently revoked. Same silent-revocation class as #600.

Neither direction is a regression from #642 — `opdivOptions` is derived from the same stale `userInfo`, so the narrow-scope path predates it. #642 closed the last *derivation* gap between the frontend's idea of caller scope and the backend's; this is the last *freshness* gap.

Refs #600. Precedent: #574 fixed the same class of staleness for the Assign Systems picker in the same file.

### Steps to Reproduce

1. Sign in as an `OPDIV_ADMIN` who holds a grant to OpDiv A. Navigate to Users and leave the tab open.
2. In a second session, an `OWNER` or `HHS_ADMIN` grants that same admin an additional OpDiv B.
3. Grant OpDiv B to some target user as well, so the target now holds both A and B.
4. Back in the first tab, without reloading, open Assign OpDivs for that target user.
5. Change nothing. Click Save.

### Expected Behavior

The target keeps both grants. The admin now holds B, so B is inside their backend scope and must be preserved on save.

### Actual Behavior

The PUT body carries only A. `callerScope` was built from the session's stale `assignedopdivids`, which predates the B grant, so B is filtered out of `idsToSave`. The backend sees `IsAssignedOpDiv(B) = true`, places B in `toRemove`, and the target's grant to B is revoked with no error and no confirmation.

Navigating away from `/users` and back re-runs the loader and self-heals the scope, which is why this is easy to miss.

### Environment

Any browser. Reproduces against `main` at ztmf-ui `2196d05` (the head of #642) and every commit before it. Requires the backend's scoped-admin reconciliation path, so `OPDIV_ADMIN` only — unscoped `OWNER`/`HHS_ADMIN` callers skip the filter entirely and are unaffected.

### Acceptance Criteria

- [ ] An `OPDIV_ADMIN` whose own grants change mid-session gets the current scope on the next Assign OpDivs open, without a page reload or a navigation round-trip.
- [ ] The stale-narrow repro above preserves the target's grant to B instead of silently revoking it.
- [ ] The stale-wide direction no longer produces an unexplained 403 on an unmodified modal.
- [ ] A regression test covers the divergence: caller scope changes between two modal opens, and the second open uses the newer set.
- [ ] The dropdown's assignable set stays derived from the refreshed scope, so `callerGrantIds` remains a superset of `assignableIds`. That invariant is what keeps the save from stripping a legitimately-held grant.
- [ ] No extra `/users/current` fetches on unrelated interactions.

### Severity

Low

### Affected Component

Users page — Assign OpDivs modal (`src/views/OpDivGrantModal/OpDivGrantModal.tsx`, `src/views/UserTable/UserTable.tsx`)

### Workaround

Reload the page, or navigate away from `/users` and back, before editing OpDiv grants in a session where your own grants may have changed. Both re-run the route loader and refresh `userInfo`.

tarratsco:
## Summary

Closes #643. Builds on #642.

The Assign OpDivs modal decided what to preserve on save from the acting admin's own OpDiv grants, read from `userInfo.assignedopdivids`. That value comes from the route loader's `/users/current` once per session and is never refreshed, so an `OPDIV_ADMIN` whose own grants change mid-session works from a stale scope. #642 made that scope the save-time preserve boundary, so a stale value now has a data-integrity consequence, not just a cosmetic one:

* **Stale-narrow (caller gained a grant B the target also holds):** B isn't in the stale scope, so the save strips it from the PUT, and the backend's `toRemove` gate (pure grant membership) still sees `IsAssignedOpDiv(B)=true` and silently revokes the target's grant.
* **Stale-wide (caller lost a grant A):** the PUT still carries A and the backend's pre-gate 403s the whole unmodified save.

This is the last freshness gap between the frontend's idea of caller scope and the backend's (#642 closed the last derivation gap).

## Changes

* **Fresh scope on open.** The modal now fetches the caller's own current grants on open (scoped callers only), alongside the target's grants in the same `Promise.allSettled`. Both the dropdown's assignable set and the save-time preserve boundary derive from that fresh set, keeping the `callerScope ⊇ assignableIds` invariant.
* **UserTable** passes the full active/non-parent OpDiv list plus the caller's user ID, instead of a caller-narrowed, session-stale list. Its own `opdivOptions` (and the inline `EditOpDivCell`) are left untouched.
* **Hardening (from review):** a caller-scope fetch failure - or an empty-but-fetched scope - blocks Save, so an incomplete scope can never filter the PUT; a single error toast when both fetches fail; and a self-row open (`userid === callerUserId`) reuses the one grants request instead of a duplicate GET.

## Acceptance criteria

* [x] An `OPDIV_ADMIN` whose own grants change mid-session gets the current scope on the next Assign OpDivs open, without a reload or navigation round-trip.
* [x] The stale-narrow case preserves the grant instead of silently revoking it.
* [x] The stale-wide case no longer 403s an unmodified modal.
* [x] A regression test covers the mid-session change between two opens, and the second open uses the newer set.
* [x] The dropdown's assignable set stays within scope (`callerScope ⊇ assignableIds`).
* [x] No extra `/users/current` fetches - the modal hits the scoped `assignedopdivs` endpoint on open, not `current`.

## Testing

* `OpDivGrantModal` suite reworked and extended to cover fresh-scope save behavior, the mid-session regression, the caller-fetch-failure / empty-scope / self-row / single-toast / caller-401 paths, plus the existing coverage.
* Verified the load-bearing FE/BE contract by hand against the backend: `enforceCallerScope` (`!isUnscopedWriteAdmin`) is the exact match for `IsUnscoped()` for every role that can reach the modal (`OWNER` / `HHS_ADMIN` unscoped, `OPDIV_ADMIN` scoped).
* `yarn jest`: full suite green (723 passing); `tsc --noEmit` and lint clean (pre-existing unrelated warnings only).
* Behavior-only fix; no visual changes.

## Notes / out of scope

* **A seconds-wide open-to-save TOCTOU remains by design.** A concurrent change to the caller's own grants during an active edit isn't caught until the next open. It's unclosable client-side (a save-time refetch only narrows it), and the backend grant-membership gate is the final authority. Documented at the fetch site.
* `EditOpDivCell` (the inline grid edit) is not affected by this class. It only calls `setEditCellValue` (writes nothing), and `setUserOpDivs` has a single call site on the new-user (`isNew`) path: a new user has no prior grants to revoke, and the existing-user save PUTs only `email/fullname/role`. So stale caller scope cannot cause a silent revocation there. (Separately, pre-existing and out of scope, inline-editing an existing user's OpDivs is silently discarded on save while still showing "Saved", which is worth its own bug ticket, unrelated to freshness.)

### DIFF
--- a/src/views/OpDivGrantModal/OpDivGrantModal.test.tsx
+++ b/src/views/OpDivGrantModal/OpDivGrantModal.test.tsx
@@ -31,11 +31,13 @@ const mock = new MockAdapter(axiosInstance)
 
 const USER_ID = 'aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa'
 const USER_ID_B = 'bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb'
+const CALLER_ID = 'cccccccc-cccc-4ccc-8ccc-cccccccccccc'
 
-// Represents the caller's grantable scope (children, active, and — for an
-// OPDIV_ADMIN — limited to their own OpDivs). OpDiv 99 is intentionally absent
-// so scope-filter tests can verify it is stripped from the PUT body.
-const opdivOptions: OpDiv[] = [
+// The full assignable universe: active, non-parent OpDivs, NOT narrowed to the
+// caller. The modal narrows this against the caller's fresh grants itself. 99 is
+// intentionally absent (parent/inactive) so it can only ever appear as a
+// current grant, never as a selectable option.
+const assignableOpDivs: OpDiv[] = [
   {
     opdiv_id: 1,
     code: 'AAA',
@@ -56,13 +58,20 @@ const opdivOptions: OpDiv[] = [
 
 // Full label source (incl. the non-assignable OpDiv 99, e.g. a parent/inactive
 // division), so the modal can label a grant to it even though it is absent from
-// opdivOptions. Id 77 is intentionally absent to exercise the "OpDiv #{id}"
+// assignableOpDivs. Id 77 is intentionally absent to exercise the "OpDiv #{id}"
 // fallback.
 const opdivLabelMap: Record<number, { code: string; name: string }> = {
   1: { code: 'AAA', name: 'Division A' },
   2: { code: 'BBB', name: 'Division B' },
   99: { code: 'ZZZ', name: 'Parent Division' },
 }
+
+// The acting caller's own current grants, served fresh by the modal's on-open
+// fetch. Mutable so a test can change it between two opens (the staleness case).
+let callerGrants: number[] = [1, 2]
+// Status the caller-scope fetch returns; non-200 exercises the failure paths
+// (500 = generic error guard, 401 = auth redirect).
+let callerFetchStatus = 200
 
 function renderModal(
   overrides: Partial<React.ComponentProps<typeof OpDivGrantModal>> = {}
@@ -73,19 +82,37 @@ function renderModal(
       handleClose={jest.fn()}
       userid={USER_ID}
       userName="Test User"
-      opdivOptions={opdivOptions}
+      assignableOpDivs={assignableOpDivs}
       opdivLabelMap={opdivLabelMap}
       enforceCallerScope={true}
-      callerGrantIds={[1, 2]}
+      callerUserId={CALLER_ID}
       onChanged={jest.fn()}
       {...overrides}
     />
   )
 }
 
+// Waits for the modal's on-open fetches to settle (Save enabled = not loading).
+// Fetch-count-agnostic, so it survives the extra caller-scope fetch.
+const waitForReady = () =>
+  waitFor(() =>
+    expect(screen.getByRole('button', { name: /^save$/i })).toBeEnabled()
+  )
+
 beforeEach(() => {
   mock.reset()
   mockedNavigate.mockReset()
+  callerGrants = [1, 2]
+  callerFetchStatus = 200
+  // The caller-scope fetch is read at request time, so mutating callerGrants
+  // (or callerFetchStatus) between opens changes what the next open sees.
+  mock
+    .onGet(`/users/${CALLER_ID}/assignedopdivs`)
+    .reply(() =>
+      callerFetchStatus === 200
+        ? [200, { data: callerGrants }]
+        : [callerFetchStatus, {}]
+    )
 })
 
 // Scoped caller (OPDIV_ADMIN, enforceCallerScope=true): the save strips
@@ -99,7 +126,7 @@ test('scoped caller: PUT body excludes grants the target holds outside caller sc
   mock.onPut(`/users/${USER_ID}/opdivs`).reply(204)
 
   renderModal({ enforceCallerScope: true })
-  await waitFor(() => expect(mock.history.get).toHaveLength(1))
+  await waitForReady()
 
   await userEvent.click(screen.getByRole('button', { name: /^save$/i }))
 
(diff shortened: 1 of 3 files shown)


jsos3-cms (approved):
Approving. I traced both failure modes from #643 against the backend's gate semantics and both resolve, including the half that is easy to get backwards: in the stale-wide case, dropping A from the PUT is what *preserves* it, because the `toRemove` gate only revokes ids the caller holds.

One thing worth more credit than the description gives it. `assignableIds` is now `assignableOpDivs.filter((id) => callerScope.has(id))`, so the `callerScope ⊇ assignableIds` invariant holds by construction rather than by the parent remembering to pass a pre-narrowed list. That closes a class of drift, not just the freshness gap.

The hardening is genuinely fail-closed: either fetch rejecting disables the picker and Save, a fetched-but-empty scope blocks Save, the `cancelled` flag guards both paths, and close resets all four pieces of state. `tsc` and lint clean, and the touched suites pass.

## One correction, in your favour

The note says `EditOpDivCell` is still on stale scope, "the same class on a different surface," left for a follow-up. It is not the same class, and I do not think that follow-up is needed.

`EditOpDivCell` writes nothing; it is a `renderEditCell` that only calls `setEditCellValue`. And `setUserOpDivs` has exactly one call site in `UserTable` (line 326), inside the `newRow.isNew` branch. A new user has no prior grants to revoke, and the existing-user branch PUTs only `email`, `fullname`, and `role`. So the silent-revocation class cannot reach that surface.

What is there instead is a different, pre-existing bug. The `opdivs` column is `editable: isAdmin` and is not gated to new rows, so an admin can inline-edit an existing user's OpDivs, `processRowUpdate` takes the else branch and discards the edit, and the row still gets `STATUS_MESSAGES.saved` plus an optimistic state update. The change appears to apply, nothing persists, and it reverts on refresh. Not yours and not in scope, but worth pointing the follow-up ticket at that rather than at stale scope.

## Note on the test suite

My full-suite run showed 4 failures, so I checked before raising it: `origin/main` at `5ca9858` without this PR shows **18** failures across 8 suites, a completely disjoint set, all timeout-shaped and all passing in isolation. That is environment, not this change. Your "full suite green" holds.

Both accepted costs read as the right calls to me. The extra scoped GET per open is the price of freshness, and memoizing it would reinstate exactly the staleness this fixes. And the open-to-save TOCTOU really is unclosable client-side, so documenting it at the fetch site and leaving the backend as final authority is the correct treatment.

jsos3-cms:

Issue #619, opened 2026-07-27
A. Introduce a data-fetching hook layer over axiosInstance


voidspooks:
Child A of the epic: https://github.com/CMS-Enterprise/ztmf-misc/issues/319

The foundation slice. Everything else in the epic waits on this, so it ships the library, the config, the docs, and one worked example, and nothing more.

## Scope

Install `@tanstack/react-query` and its devtools.

Add `src/queryClient.ts` exporting a configured `QueryClient`: `staleTime: 60_000`, `retry: 1`, `refetchOnWindowFocus: false`. This app is admin workflows, so aggressive refetching is noise.

Wrap `<App>` in `main.tsx` with `QueryClientProvider`, and gate the devtools on `import.meta.env.DEV`.

Add `src/api/keys.ts` (or equivalent) naming each endpoint once. This is what fixes the third problem on the epic: `GET /users/:id` currently appears in five places across three files, and paths are rooted inconsistently, `fismasystems/${id}` in one file against `/fismasystems/${id}/questions` in another.

Write `docs/data-fetching.md` covering query-key conventions, when to reach for `useQuery` against `useMutation` against neither, and the interop notes below.

## The worked example

Pick a single-call surface: `dataCenterEnvironments` or `SystemEnrichmentCard`.

Not `Home`. It is a `Promise.all` fan-out over `activeDatacallIds` for two endpoints, merged by `buildDashboardMaps`, which maps to `useQueries` and teaches the wrong first lesson.

## Interop that has to survive

The transport layer stays exactly as it is. Axios remains the transport, and the interceptor, `parseApiError`, and `notify` are unchanged. Each of these gets a test here:

- A 401 raised inside a `queryFn` still redirects to login through `authInterceptor.ts`.
- The `isAuthHandled` short-circuit is honored in `onError`, so a handled auth error does not double-toast.
- `queryFn({ signal })` is threaded into the axios config, so cancellation works through the library rather than the ad-hoc `AbortController` usage in `Home` and `Title`.
- `skipAuthHandling: true` still bypasses the interceptor at all five call sites: `authLoader.ts:42`, `authLookup.ts:73`, `delegates.ts:83`, `SystemEnrichmentCard.tsx:124`, `Title.tsx:304`.

`src/router/authLoader.ts` keeps its raw axios call and is not migrated. The reasoning is recorded on the epic, and child J allowlists it in the lint rule.

## Decided, so it does not need re-litigating

[...]

PR #654, opened 2026-07-31
chore(#578): retire the hasScore score-presence proxy from ProgressCell (rehome of #646)

### LINKED ISSUE #578: Cleanup: retire the hasScore proxy from ProgressCell once ztmf#437 is deployed everywhere

jsos3-cms:
> **Note:** In-repo copy of #646 by @voidspooks (Cameron Testerman). Moved from the fork so Snyk
> and the deploy CI gates can run, since those jobs need org secrets that fork PRs cannot
> access. Commits and authorship are unchanged.
>
> Refs #646

---

Closes #578.

`ProgressCell` decided between a neutral "Complete" chip and an em dash for a past call by asking whether the system had any score row at all, passed down as `hasScore`. That proxy existed because `ScoreProgress` had no count of answered questions. ztmf#437 added one, so the cell can compare answered against expected directly and the proxy can go.

`questionsanswered` moves from optional to required on `ScoreProgress`, the `answered == null` fallback branches come out of `progressColumn.tsx` and `progressHelpers.ts`, and `FismaTable.tsx` stops passing `hasScore`. A past call now reads as "Complete" when fully answered and as an `answered/total` fraction with an "Incomplete" chip when partially answered, rather than collapsing both into "Complete".

## Why retiring the proxy is safe

`hasScore` was already redundant. `buildDashboardMaps` derives `systemIdxs` from the score map, and the chosen call is always one the system has scores in, so `scoreMap[sys]` is populated whenever `progressMap[sys]` is. The `!hasScore` branch inside the `answered == null` case was therefore unreachable in normal operation; it only did anything during the ztmf#437 deploy window when the backend genuinely omitted the field.

It is also a small improvement against a known failure mode. `hasScore` only asked whether a score row existed, and a system whose answers point at another environment's functions still produces one, so it read as "Complete" regardless. `questionsanswered` is computed through an INNER JOIN on `functionoptions` with the function's environment re-matched against the system's, so those answers are excluded from the numerator instead of counted.

## Added on top of the original commit

One extra commit, `73f3e25`, fixing a docblock paragraph the original left behind. The `States:` list was updated to the answered/expected rule, but the paragraph above it still said a past call with a score for that system was treated as completed, so the two contradicted each other in the same comment block. Comment-only, no behaviour change.

## Test plan

- [x] `npx jest src/views/FismaTable src/views/Home` — 75 passing
- [x] `tsc --noEmit` clean, `eslint` clean, `prettier --check` clean on the changed files
- [x] `0/0` still renders the neutral N/A chip, since `hasNoQuestionnaire()` short-circuits ahead of the past-call branch
- [ ] Dashboard spot check on dev: a fully answered past call reads Complete, a partially answered one reads the fraction with Incomplete

## Known follow-up, not addressed here

`questionsanswered` being required is enforced only by TypeScript. `Home.tsx` casts the response with `as ScoreProgress[]` rather than validating it, so if a row ever arrived without the field the cell would render the literal string `undefined/41` and `progressSortValue` would return `NaN`, leaving that row's sort position undefined. ztmf#437 has been live in prod since 2026-07-17, so the assumption holds today, but the last runtime guard against a contract violation is gone. A `Number.isFinite` check at ingestion that logs rather than renders would close it. Worth its own ticket rather than widening this one.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -334,9 +334,8 @@ export type ScoreProgress = {
   questionsexpected: number
   // Distinct applicable questions with an answer at all, any status (ztmf#437).
   // The completion signal a closed call needs: imported/carried answers are
-  // answered but never "updated this cycle". Optional so the UI degrades to the
-  // prior score-presence proxy until the backend field is deployed.
-  questionsanswered?: number
+  // answered but never "updated this cycle".
+  questionsanswered: number
   questionsupdated: number
   lastupdatedat?: string | null
   updatedsincestart: boolean
--- a/src/views/FismaTable/FismaTable.tsx
+++ b/src/views/FismaTable/FismaTable.tsx
@@ -931,7 +931,6 @@ export default function FismaTable({
         <ProgressCell
           entry={progress?.[params.row.fismasystemid]}
           isCurrentCall={isRowCurrentCall(params.row.fismasystemid)}
-          hasScore={Boolean(scores[params.row.fismasystemid])}
         />
       ),
     },
--- a/src/views/FismaTable/progressColumn.test.tsx
+++ b/src/views/FismaTable/progressColumn.test.tsx
@@ -102,6 +102,25 @@ describe('progressSortValue', () => {
       progressSortValue(pastComplete, false)
     )
   })
+
+  it('returns a usable rank for a past-call row missing the answered count', () => {
+    // The type requires the field, so absence can only be reached at runtime
+    // against a response that omits it, and the cast is how the test expresses
+    // that. Without coalescing this returned NaN, which compares equal to every
+    // other rank and drops the row at an arbitrary position instead of ranking
+    // it. Zero is the honest rank: nothing recorded as answered.
+    const missingAnswered = {
+      ...untouchedEntry,
+      questionsanswered: undefined,
+    } as unknown as ScoreProgress
+    const rank = progressSortValue(missingAnswered, false)
+    expect(Number.isNaN(rank)).toBe(false)
+    expect(rank).toBe(0)
+    // Ranks with the genuinely unanswered rather than above a partial call.
+    expect(rank).toBeLessThan(
+      progressSortValue({ ...untouchedEntry, questionsanswered: 10 }, false)
+    )
+  })
 })
 
 describe('progressTooltip', () => {
@@ -201,6 +220,22 @@ describe('ProgressCell', () => {
     expect(screen.getByText('0/41')).toBeInTheDocument()
     expect(screen.getByText('Incomplete')).toBeInTheDocument()
     expect(screen.queryByText('Not updated')).not.toBeInTheDocument()
+    expect(screen.queryByText('Complete')).not.toBeInTheDocument()
+  })
+
+  it('renders 0/total on a past call when the answered count is missing', () => {
+    // Reachable only at runtime against a response that omits the field, which
+    // the cast expresses. The fraction interpolates the value directly and React
+    // drops undefined, so without coalescing this rendered a bare "/41" beside a
+    // warning chip. This is the one input where retiring the score-presence
+    // proxy changes what a reader sees, so it is pinned here.
+    const missingAnswered = {
+      ...untouchedEntry,
+      questionsanswered: undefined,
+    } as unknown as ScoreProgress
+    render(<ProgressCell entry={missingAnswered} isCurrentCall={false} />)
+    expect(screen.getByText('0/41')).toBeInTheDocument()
+    expect(screen.getByText('Incomplete')).toBeInTheDocument()
     expect(screen.queryByText('Complete')).not.toBeInTheDocument()
   })
 
--- a/src/views/FismaTable/progressColumn.test.tsx
+++ b/src/views/FismaTable/progressColumn.test.tsx
@@ -307,18 +307,40 @@ describe('ProgressCell', () => {
     expect(screen.queryByText('Not updated')).not.toBeInTheDocument()
   })
 
-  it('keeps the legacy Not updated wording when questionsanswered is absent', () => {
-    // The field is now required on ScoreProgress, so absence can no longer be
-    // expressed through the type. It remains reachable at runtime against a
-    // backend that predates the field, which is exactly what the cell's
-    // fallback guards, so the cast is deliberate: it exercises the runtime
-    // path the type alone can no longer describe.
+  it('reads Not started on the current call when the answered count is absent', () => {
+    // The field is required on ScoreProgress, so absence is reachable only at
+    // runtime against a response that omits it, and the cast is how the test
+    // expresses that. Both branches now coalesce, so this reads as nothing
+    // answered rather than falling back to the retired "Not updated" wording.
     const withoutAnswered = {
       ...untouchedEntry,
       questionsanswered: undefined,
     } as unknown as ScoreProgress
     render(<ProgressCell entry={withoutAnswered} isCurrentCall={true} />)
-    expect(screen.getByText('Not updated')).toBeInTheDocument()
+    expect(screen.getByText('Not started')).toBeInTheDocument()
+    expect(screen.queryByText('Not updated')).not.toBeInTheDocument()
+  })
+
+  it('no longer renders the retired Not updated wording in any state', () => {
+    // The label had one emit site, so this is the guard against it coming back
+    // through a future conditional rather than a restatement of the cases above.
+    for (const entry of [
+      updatedEntry,
+      untouchedEntry,
+      { ...untouchedEntry, questionsanswered: 41 },
+      {
+        ...untouchedEntry,
+        questionsanswered: undefined,
+      } as unknown as ScoreProgress,
+    ]) {
+      for (const isCurrentCall of [true, false]) {
+        const { unmount } = render(
+          <ProgressCell entry={entry} isCurrentCall={isCurrentCall} />
+        )
+        expect(screen.queryByText('Not updated')).not.toBeInTheDocument()
+        unmount()
+      }
+    }
   })
 
   it('describes the carried-forward state in the tooltip', () => {
--- a/src/views/FismaTable/progressColumn.test.tsx
+++ b/src/views/FismaTable/progressColumn.test.tsx
@@ -6,6 +6,7 @@ import { progressSortValue, progressTooltip } from './progressHelpers'
 const updatedEntry: ScoreProgress = {
   fismasystemid: 1,
   questionsexpected: 41,
+  questionsanswered: 12,
   questionsupdated: 12,
   lastupdatedat: '2026-07-01T15:30:00Z',
   updatedsincestart: true,
@@ -14,6 +15,7 @@ const updatedEntry: ScoreProgress = {
 const untouchedEntry: ScoreProgress = {
   fismasystemid: 2,
   questionsexpected: 41,
+  questionsanswered: 0,
   questionsupdated: 0,
   lastupdatedat: null,
   updatedsincestart: false,
@@ -52,6 +54,7 @@ describe('progressSortValue', () => {
     const empty: ScoreProgress = {
       fismasystemid: 3,
       questionsexpected: 0,
+      questionsanswered: 0,
       questionsupdated: 0,
       updatedsincestart: false,
     }
@@ -69,6 +72,7 @@ describe('progressSortValue', () => {
     const empty: ScoreProgress = {
       fismasystemid: 3,
       questionsexpected: 0,
+      questionsanswered: 0,
       questionsupdated: 0,
       updatedsincestart: false,
     }
@@ -127,6 +131,7 @@ describe('progressTooltip', () => {
     const empty: ScoreProgress = {
       fismasystemid: 3,
       questionsexpected: 0,
+      questionsanswered: 0,
       questionsupdated: 0,
       updatedsincestart: false,
     }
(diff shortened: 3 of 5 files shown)


jsos3-cms:
@voidspooks could you review this one? It is your change, rehomed in-repo so the Snyk and deploy gates can actually run, and your authorship is preserved on the original commit.

Two things worth your eyes specifically.

I added one commit on top of yours, `73f3e25`, rather than sending you back for a two-line fix. The `States:` list in the `progressColumn.tsx` docblock was correctly updated to the answered/expected rule, but the paragraph five lines above it still described the retired score-presence behaviour, saying a past call with a score for the system was treated as completed. The two contradicted each other in the same comment block, so I rewrote the paragraph to match. Comment-only, no behaviour change, and it is a separate commit so it is easy to see and easy to object to.

Second, and this is a genuine compliment rather than a nit: retiring the proxy turns out to be slightly better than a cleanup. `hasScore` only asked whether a score row existed, and a system whose answers point at another environment's functions still produces one, so it read as "Complete" regardless. `questionsanswered` is counted through an INNER JOIN on `functionoptions` with the environment re-matched, so those answers drop out of the numerator instead. That is a small win against the class of bug where a system scores exactly 1.00 and renders as a legitimate lowest tier.

I also traced why the proxy was safe to remove at all: `buildDashboardMaps` builds `systemIdxs` from the score map and the chosen call is always one the system has scores in, so `scoreMap[sys]` is populated whenever `progressMap[sys]` is. The `!hasScore` branch was unreachable outside the ztmf#437 deploy window.

One follow-up I did not fold in, noted in the description: `questionsanswered` being required is a TypeScript guarantee only, and `Home.tsx` casts rather than validates, so a row arriving without it would render `undefined/41` and give that row a `NaN` sort key. ztmf#437 has been live since 2026-07-17 so it holds today, but the last runtime guard is gone. Happy to open that ticket unless you would rather handle it here.

The original #646 stays open as a draft until this merges, then I will close it pointing here.

jsos3-cms:
Rebased onto current `main` now that #665 has landed, and force-pushed. Head is `886d35b`. Cameron's authorship and the signatures are intact on both commits, and the reconciliation below is folded into his commit rather than added as a third, so each commit passes its own tests instead of leaving the first one red.

Flagging one thing for whoever reviews, because a piece of the diff will look unmotivated without it.

## The rebase applied clean and still broke three tests

Git found no conflicts, which was misleading. This branch makes `questionsanswered` required on `ScoreProgress` and adds it to the two shared fixtures in `progressColumn.test.tsx`. #665 had meanwhile split the current-call wording so that zero answers reads "Not started" and carried answers reads "Awaiting confirmation". The moment the shared fixtures started carrying `questionsanswered: 0`, every test reusing them stopped taking the legacy "Not updated" branch. The hunks never overlapped, so there was nothing for git to report.

Three expectations were stale as a result, and I corrected them to match the fixtures rather than reverting the newer wording:

- The test that claimed to cover "a carried-over system" was asserting "Not updated" against a fixture with zero answers, which is the "Not started" case. Renamed it to say what it actually covers, since the carried-answers case has its own test.
- The current-call-versus-past-call contrast test now expects "Not started" and also asserts the past-call "Incomplete" chip is absent, so it pins the contrast in both directions.
- The legacy-wording test is the one worth explaining. It pins the fallback for an **absent** `questionsanswered`, which this branch's type change makes impossible to express through the type. I kept the test and forced the absence with a cast, because the fallback is still reachable at runtime against a backend that predates the field. That cast is deliberate and commented, not an oversight.

## A question this raises, and I do not think it is mine to answer

This branch now declares `questionsanswered` required, while `progressColumn.tsx` still keeps a runtime fallback for its absence and emits the legacy "Not updated" when it is missing. Those two statements disagree with each other, and my cast is what lets the test express the disagreement.

Retiring that fallback is arguably the natural completion of this change: the field ships from the backend, the type now says it is always there, and the second proxy is the same kind of thing this PR exists to remove. I deliberately left it alone because it is #665's code and added for a stated deployment reason, so widening this PR into it felt like the wrong call to make unilaterally.

Happy to do it here if reviewers prefer one coherent change, or to file it as a follow-up. Either way the fallback and the required type should not stay in disagreement indefinitely.

## Verification

`tsc --noEmit` clean, full jest suite 798 passed with 0 failures, `eslint ./src` reporting 0 errors and the 2 pre-existing warnings in a file this PR does not touch. The `hasScore` occurrences that remain in `aggregateScores.ts` are a different local variable that arrived with the not-updated undercount fix, unrelated to the ProgressCell prop being retired here.

Needs a fresh approval since the force-push dismissed the previous one.

voidspooks (approved):
Reviewed both commits plus the rebase reconciliation, and re-ran the gate locally on 886d35b: tsc clean, eslint 0 errors with the 2 pre-existing warnings, jest 798 passing. The three corrected expectations now say what the fixtures actually exercise, and the deliberate cast in the legacy-wording test is the right way to keep the runtime fallback pinned once the type can no longer express absence. 73f3e25 is a fix, the paragraph and the States list really did contradict each other. Thanks also for the functionoptions trace, an environment-mismatched answer dropping out of the numerator is a better argument for this change than the one I wrote in the description.

On your two questions my answer to both is a follow-up, and I think they are one ticket rather than two, because they are the same question: what a row without questionsanswered should do at runtime now that the type says it cannot happen. For the guard, Home.tsx casting instead of validating is not specific to this field, every axios site in the app casts, so a one-off guard here would be an island. ui#619 already proposes the fetch-hook layer where response validation belongs, so the ticket should point that half there. For the legacy "Not updated" fallback, I agree it is the same species of proxy this PR retires, and its stated reason has aged out with ztmf#437 live everywhere since 7/17. But it is #665's code with a deployment rationale on it, so it should go in its own small PR with Danny's ack rather than a widening here that force-pushes away another approval on a merge that deploys to prod. I will file the follow-up ticket and link it back here.

Approving as-is.

voidspooks:
Filed the follow-up as #672, covering both halves: the response validation, scoped to #619 rather than a one-off guard in Home.tsx, and the contradiction between the required type and the legacy-wording fallback.

One correction while checking it. The row would not render `undefined/41`, React drops `undefined` out of the fraction, so a past-call row with the field absent renders a bare `/41` next to the warning Incomplete chip. `progressSortValue` does return NaN, same class as the `deadlineByCallName` sort key you fixed on #648. Verified both against 886d35b with a throwaway test rather than by reading. The current-call branch is unaffected because your fallback is still there, so only the past-call branch you rewrote is exposed.

jsos3-cms:
Pushed `abb236c`, which hardens the branch this PR rewrote, and a correction to something I got wrong earlier in this thread.

## Correction: my `buildDashboardMaps` argument was wrong

I wrote above that `buildDashboardMaps` derives `systemIdxs` from the score map so `scoreMap[sys]` is populated whenever `progressMap[sys]` is, and that the old `!hasScore` branch was therefore unreachable outside the deploy window. That is not true, and this repo's own test says so: `aggregateScores.test.ts` has "includes a system present only in progress (no score row)", which asserts `scoreMap[7]` is undefined while `progressMap[7]` is populated. The code's comment in that file is explicit that the property is upheld by a backend invariant rather than by the frontend logic. So that branch was reachable, and nobody should lean on my reasoning as a second line of defence.

The removal is still correct, but for the other reason I gave separately: `questionsanswered` is served and has been deployed since 2026-07-17, so a never-scored system now arrives with a real `0` and lands in the new "0/N Incomplete" branch, which is the intended outcome. Wrong reasoning, right conclusion, and worth correcting since it was load-bearing in how I described the change.

## What `abb236c` fixes

The past-call branch read `questionsanswered` with no null check, while the current-call branch three lines below kept one. So for a response omitting the field, the branch this PR rewrote was the exposed one:

- the fraction interpolates the value directly and React drops `undefined`, so the cell rendered a bare `/41` next to a warning "Incomplete" chip;
- `progressSortValue` returned `NaN`, because `Math.max(undefined, 0)` is `NaN`.

The NaN is the more interesting half. A NaN sort key makes MUI's number comparator return NaN, which `Array.prototype.sort` treats as "equal", so the row does not sort high or low — it settles wherever its all-equal comparisons leave it, depending on its pre-sort position and the array size. Driving the comparator directly, a row starting at index 25 of 100 finished at 50, and one starting at 75 finished at 95. That defeats the triage ordering the column's own docblock describes, and it is implementation-defined rather than merely undesirable.

The fix coalesces to zero in both the cell and the sort key, so a missing count reads as nothing answered, which is exactly what the branch already renders for a genuinely unanswered row.

Two tests come with it, and both fail without the source change and pass with it, verified by reverting the source while keeping the tests rather than assuming: one pins `0/41` plus Incomplete for the missing-field render, and one pins a real numeric rank instead of NaN and asserts it ranks below a partially-answered call.

This also closes the coverage gap that made the retirement invisible to the suite. Before this, every past-call test supplied a concrete count, so no test could tell the old proxy implementation from the new one — the suite would have passed with `hasScore` restored. The missing-field case is the one input where the two genuinely differ, and it is now pinned.

Scope note: this restores parity with the guard the sibling branch still has, on code this PR itself rewrote. #672 still owns the architectural question of validating the response and reconciling the required type with the legacy fallback, and nothing here preempts it.

## Verification

`tsc --noEmit` clean, `eslint ./src` 0 errors with the 2 pre-existing warnings in an untouched file, full jest suite 800 passed and 0 failed. Cameron's authorship and the signatures are intact across all three commits.

jsos3-cms:
@danielbowne one question for you while you have this open, because answering it here may remove a follow-up entirely.

`abb236c` coalesces the answered count on the past-call branch. The current-call branch still keeps the `answered == null` fallback you added in #665, which emits the legacy "Not updated" wording when the field is missing. So the file now has a required type, one branch that coalesces, and one that branches on absence.

#672 tracks settling that, and @voidspooks' preference there is to retire the fallback so the code agrees with the type. I did not touch it: it changes user-visible wording on the degraded path, since a backend without `questionsanswered` would read "Not started" instead of "Not updated", and it is your code with a stated deployment reason. That felt like your call rather than something to fold into a rehome silently.

If you are happy to retire it, I will do it in this PR and close #672 as subsumed, so we are not carrying a ticket for a one-branch deletion. If you would rather keep it, say so and #672 stays as the record that the asymmetry is deliberate. Either answer is quick and neither blocks your review of the rest.

The response-validation half of #672 is out of scope either way: that is the app-wide casting pattern, 72 `axiosInstance` call sites, tracked with #619.

jsos3-cms:
Pushed `1b814fe`, which retires the `answered == null` fallback so both branches read the count the same way. That completes the reconciliation rather than leaving it as a follow-up, and #672 is now fully covered by this PR.

@danielbowne I folded this in for ease rather than waiting, since the gap was two lines from the change already in front of you and splitting it across two PRs would have cost you a second review of the same file. **It rolls back cleanly if you would rather keep the fallback** — it is one isolated commit and nothing else in the PR depends on it, so say the word and I will drop it.

The one user-visible consequence is on the degraded path: against a backend that does not serve `questionsanswered`, the current-call chip now reads "Not started" instead of "Not updated". Nothing changes at all against a backend that serves the field, which has been every deployed environment since 2026-07-17.

The reasoning for retiring rather than keeping: the fallback described a state the wording already covers. A response with no answered count and a system with zero answers are the same thing to a reader, and "Not started" already says it. Keeping a separate label for the first case meant the file asserted a required type in `types.ts`, trusted it on one branch, and guarded against it on the other. That asymmetry is what let the past-call branch ship without a guard, so removing the disagreement is worth more than preserving a distinct label for an input no deployed backend produces.

`Not updated` now has no emit site anywhere in `src/`. A test sweeps every entry shape across both call modes to keep it from coming back through a future conditional, since a single-site label is easy to reintroduce by accident. The legacy-wording test became its inverse: the absent-count case now pins "Not started" and asserts the retired label is gone.

Both new tests fail against the previous source and pass with it, verified by reverting the source while keeping the tests.

Unrelated and staying put: the "Not updated only" filter checkbox in `FismaTable.tsx` is a different string on a different control, and the wording mismatch there is its own copy question rather than part of this.

## Verification

`tsc --noEmit` clean, `eslint ./src` 0 errors with the 2 pre-existing warnings in an untouched file, full suite 801 passed and 0 failed. Four commits, all signed, Cameron's authorship intact on the first.

danielbowne:
Retire it, keep 1b814fe. The fallback was mine but its reason expired: every deployed environment has served questionsanswered since 2026-07-17, and a backend old enough to omit it predates scores.status too, so the confirm flow the wording was protecting degrades anyway. The asymmetry already produced one real bug on the past-call branch, so the type and the code agreeing is worth more than preserving a label for an input no deployed backend produces.

One fix before this merges: abb236c's comment on the past-call branch (progressColumn.tsx, the "Coalesce so a response that omits the count" block) still says the current-call branch below keeps its own null check with a distinct legacy wording. 1b814fe removed exactly that, so the file now contradicts itself three lines apart. Drop that sentence.

On the wording itself, worth saying it was considered rather than missed: "Not started" is a stronger claim than "Not updated" was. If the field ever went missing while answers exist, the old label was vague but true and the new one is false. I am fine with that trade on a path that requires a pre-July backend, and the sweep test means we will notice if anyone tries to bring the old label back.

When #672 closes as subsumed, point the response-validation half at #619 explicitly so it does not fall between the two tickets.

jsos3-cms:
@danielbowne fixed in `b69eb48`. You were right and it was my error: I wrote that sentence in `abb236c` when the current-call branch did still hold its own null check, then `1b814fe` removed exactly what it pointed at without my updating it. Dropped the clause; the coalescing rationale stands without needing to describe the other branch at all, which is also why it went stale in the first place.

Thanks for spelling out the wording trade rather than just waving it through. "Not started" being a stronger claim than "Not updated" is the part I had not stated plainly: if the count went missing while answers existed, the old label was vague but true and the new one is false. That only happens against a backend predating July, and as you say such a backend predates `scores.status` too, so the confirm flow the wording protected is already degraded there. Worth having on the record for whoever reads this next, since the sweep test enforces the decision without explaining it.

`#672` will name #619 explicitly for the response-validation half when it closes as subsumed, rather than leaving it implied.

`tsc` clean, `eslint` 0 errors, FismaTable suite 76 passed. Head is now `b69eb48`.

One thing still open that is not code: this PR reads `REVIEW_REQUIRED` because @voidspooks is not a collaborator, so his approval does not count toward the required one. If you are satisfied, a formal approve is what unblocks it, and it is also the gate for folding this into the freeze batch, which only takes approved PRs.

danielbowne (approved):
b69eb48 resolves the one thing I flagged, and it is the exact deletion asked for. Re-ran the FismaTable and Home suites plus the typecheck on the branch to be safe: 88 passing, clean. The retirement question is settled above, 1b814fe stays. Good to merge.

jsos3-cms:
Shipped in #674, the frontend code-freeze batch, merged as `1d96160`. All five commits are in it, @voidspooks' retirement with his authorship intact plus the coalescing, the wording-fallback retirement, and the comment fix @danielbowne asked for.

Closing by hand: the batch squash-merged, so GitHub cannot tell this content landed. #578 was closed by the keyword declared on #674.

PR #657, opened 2026-08-03
Count score-less systems in "Not updated only" filter

### LINKED ISSUE #639: disable "Not updated only" toggle for closed data calls

### LINKED ISSUE #655: [P1]"Not updated only" undercounts: systems with no score rows for the call are invisible to the dashboard filter
### Description

On the Home dashboard, the "Not updated only" toggle undercounts badly for the current (open) data call. It returns 901 systems when the true number of not-started systems for that call is 1,425. The 524 systems it hides are exactly the ones that have not started at all, which are the ones that most need chasing.

The cause is that a system with no score rows for any active call never gets a progress entry in the dashboard's maps, so the filter treats it as "no progress data" rather than "not updated". The backend is fine: `/scores/progress?datacallid=N` returns a correct row for these systems (`questionsexpected: 40`, `questionsupdated: 0`). The frontend drops the row on the floor while building its maps.

Verified against a clone of production data:
- backend progress query returns 1,436 live systems for the current call
- 1,425 of those have `questionsexpected > 0` and `questionsupdated = 0`, so they are genuinely not started
- the dashboard's "Not updated only" facet shows 901

### Root cause

`src/views/Home/Home.tsx` (lines 32-63) fetches `/scores/aggregate?datacallid=N` and `/scores/progress?datacallid=N` for each active call, then hands both arrays to `buildDashboardMaps`.

`src/views/Home/aggregateScores.ts` builds its `systemIdxs` map from the scores responses only (lines 53-62):

```ts
// The call indices each system appears in (scores drive the table), newest
// first since callIds is newest first.
const systemIdxs = new Map<number, number[]>()
scoreByCall.forEach((m, idx) => {
  for (const sys of m.keys()) {
    ...
  }
})
```

Everything downstream is driven by that one map. The loop at lines 69-90 is what populates `progressMap`:

```ts
const progress = progressByCall[chosen]?.get(sys)
if (progress) progressMap[sys] = progress
```

A system with zero score rows in every active call never enters `systemIdxs`, so that loop never runs for it and `progressMap` never gets its entry, even though `progressByCall` holds a perfectly valid row for it.

From there:
- `src/views/FismaTable/progressHelpers.ts` line 44: `progressSortValue(undefined)` returns 2, the "no progress data" rank, instead of -1, the "not updated, has a questionnaire" rank. So these systems also sort to the bottom of the Data Call Progress column instead of the triage top.
- `src/views/FismaTable/dashboardFilters.ts` lines 87-93: `isNotUpdated` only matches rank -1, so the "Not updated only" facet rejects every one of them.

The "scores drive the table" comment documents the assumption this rests on, that any system in a data call has at least one score row. That held when every call carried answers forward. It no longer holds now that a call can contain systems with no carried-forward answers at all, and those are precisely the not-started systems the filter exists to surface.

### Not a duplicate of #639

Closed issue #639 was a different mechanism: on a closed data call, the `isRowCurrentCall` short-circuit made `isNotUpdated` return false for every row and the toggle emptied the grid. That is fixed, and the affected calls were closed ones.

This issue is on the current, open call, and the rows are not being rejected by a current-call check at all. They never make it into `progressMap` in the first place, one step earlier, during map construction in `buildDashboardMaps`. A system missing from `chosenCallMap` even gets `isRowCurrentCall` = true by the default on line 552 of `FismaTable.tsx`, so the #639 code path passes it through cleanly and it is still dropped.

### Steps to Reproduce

1. Log in as an admin or owner on an environment where the newest data call is still open.
2. Open the Home dashboard with the current data call selected in the year picker.
3. Toggle "Not updated only" on and note the row count.
4. Compare against `GET /api/v1/scores/progress?datacallid=<current call id>`, counting rows with `questionsexpected > 0` and `questionsupdated = 0`.
5. The dashboard count is far lower than the API count. The missing systems are the ones with no rows in `GET /api/v1/scores/aggregate?datacallid=<current call id>`.

### Expected Behavior

"Not updated only" returns every system that the backend reports as having a questionnaire and zero updates for the current call, whether or not that system has any score rows yet. With the production data above that is 1,425 systems. A never-started system should also sort to the top of the Data Call Progress column with the other laggards, not to the bottom as unknown.

### Actual Behavior

"Not updated only" returns 901 systems, only those that happen to have score rows. The other 524 not-started systems are invisible to the facet and are ranked 2 ("no progress data") in the progress column sort, so they fall to the bottom of the triage order.

### Environment

Prod, Dev. Reproduced against a clone of production data. Browser independent, the logic is client side.

### Suggested fix

Build `systemIdxs` from the union of the keys of `scoreByCall` and `progressByCall` instead of from scores alone. Systems that have a progress row but no scores then get their `progressMap` entry, their chosen call, and their correct -1 rank, and the score lookup stays guarded by the existing `if (score)` check so nothing gains a fake score.

Worth removing or rewriting the "scores drive the table" comment at the same time, since that assumption is what broke.

### Acceptance Criteria

- [ ] `buildDashboardMaps` includes systems that appear only in the progress response, covered by a unit test in `aggregateScores` with a system present in progress and absent from scores
- [ ] "Not updated only" count matches the backend count of `questionsexpected > 0` and `questionsupdated = 0` for the current call
- [ ] Never-started systems sort with the other laggards in the Data Call Progress column, not as "no progress data"
- [ ] #639 behavior on closed data calls is unchanged

### Severity

High - Significantly impacts usage

### Affected Component

Home dashboard, src/views/Home/aggregateScores.ts
ImageImage

tarratsco:
## Summary

Closes #655.

On the Home dashboard, the "Not updated only" toggle badly undercounted the current (open) data call. It hid exactly the systems that have not started at all, which are the ones that most need chasing.

Root cause: `buildDashboardMaps` (`src/views/Home/aggregateScores.ts`) built its `systemIdxs` map from the `/scores/aggregate` responses only, and everything downstream is driven by that map. A never-started system (one with a `/scores/progress` row where `questionsexpected > 0` and `questionsupdated = 0`, but no score row yet) never entered `systemIdxs`, so it got no `progressMap` entry despite a perfectly valid progress row being available. From there it ranked 2 ("no progress data") instead of -1 ("not updated"), so it sorted to the bottom of the Data Call Progress column, and the `isNotUpdated` facet (which only matches -1) rejected it entirely.

The stale "scores drive the table" comment documented the assumption that broke: it held when every call carried answers forward, but not once a call can contain systems with no carried-forward answers at all. This is frontend-only. The backend `/scores/progress` returns the correct rows, while the frontend dropped them during map construction.

Distinct from the closed #639 (an `isRowCurrentCall` short-circuit on closed calls), this is one step earlier, in map construction, on the open call.

## Changes

* **`aggregateScores.ts`**: build `systemIdxs` from the union of each call's score and progress rows instead of scores alone, iterated per call index so a system present in both a call's scores and progress lists that call once (no duplicate call IDs in `systemCallMap`). The score lookup stays behind the existing `if (score)` guard, so a progress-only system gains no fake score; its progress row now lands in `progressMap` with the correct -1 rank.
* Corrected the now-inaccurate `systemCallMap` docs ("calls the system has scores in" → "appears in (score or progress)") in `aggregateScores.ts` and `rowCall.ts`, and dropped a redundant `lastUpdatedMs` recompute in the chosen-call loop.

## Acceptance criteria

* [x] Bug no longer occurs: never-started systems appear in "Not updated only".
* [x] No regression: a scored system's chosen call cannot flip onto a scoreless call (a progress-only call has null `lastupdatedat`, so it never wins chosen), covered by a test.
* [x] `buildDashboardMaps` includes systems present only in progress (unit test in `aggregateScores`).
* [x] "Not updated only" count matches the backend's `expected > 0` / `updated = 0` count.
* [x] Never-started systems sort with the laggards (rank -1), not as "no progress data".
* [x] #639 closed-call behavior unchanged (separate `isRowCurrentCall` path).

## Testing

* **`aggregateScores.test.ts`**: added a progress-only-system test (present in progress, absent from scores → lands in `progressMap`, no fake score), a per-call dedupe test, and a no-regression test (a never-started progress row in another call cannot blank a scored system's score). Litmus: the progress-only test throws on the pre-fix code.
* Verified against a clone of production data via the API: for the open call, `scored = 0`, `notStarted = 27`, and `hiddenByBug = 27`. Every not-started system was dropped by the old build. The fixed dashboard's "Not updated only" now shows all 27, whereas the buggy build shows ~0.
* `yarn jest`: full suite green (749 passed); `tsc --noEmit` and lint clean (pre-existing unrelated warnings only).
* Behavior-only fix; no visual change beyond the corrected counts/ordering.

### DIFF
--- a/src/views/FismaTable/rowCall.ts
+++ b/src/views/FismaTable/rowCall.ts
@@ -39,8 +39,8 @@ export function datacallNameComparator(
  * a row is on.
  *
  * Resolution order: the call chosen by most-recently-updated in
- * buildDashboardMaps; else the newest-by-deadline call the system has scores
- * in; else a call that is actually in the current view.
+ * buildDashboardMaps; else the newest-by-deadline call the system appears in;
+ * else a call that is actually in the current view.
  *
  * That last rung matters. activeDataCallId collapses to the newest call
  * whenever more than one call is toggled on, and selecting a year toggles all
@@ -52,7 +52,8 @@ export function datacallNameComparator(
  * the label inside what the user is looking at.
  * @param {number} fismasystemid - The row's system id.
  * @param {Record<number, number>} chosenCallMap - System id -> displayed call id.
- * @param {Record<number, number[]>} systemCallMap - System id -> call ids with scores.
+ * @param {Record<number, number[]>} systemCallMap - System id -> call ids the
+ *   system appears in (score or progress rows).
  * @param {datacall[]} datacalls - All known data calls.
  * @param {number} activeDataCallId - The dashboard's active call id.
  * @param {number[]} [activeDatacallIds] - Call ids selected in the year picker.
--- a/src/views/FismaTable/rowCall.ts
+++ b/src/views/FismaTable/rowCall.ts
@@ -39,8 +39,8 @@ export function datacallNameComparator(
  * a row is on.
  *
  * Resolution order: the call chosen by most-recently-updated in
- * buildDashboardMaps; else the newest-by-deadline call the system appears in;
- * else a call that is actually in the current view.
+ * buildDashboardMaps; else the newest-by-deadline call the system has scores
+ * in; else a call that is actually in the current view.
  *
  * That last rung matters. activeDataCallId collapses to the newest call
  * whenever more than one call is toggled on, and selecting a year toggles all
@@ -52,8 +52,7 @@ export function datacallNameComparator(
  * the label inside what the user is looking at.
  * @param {number} fismasystemid - The row's system id.
  * @param {Record<number, number>} chosenCallMap - System id -> displayed call id.
- * @param {Record<number, number[]>} systemCallMap - System id -> call ids the
- *   system appears in (score or progress rows).
+ * @param {Record<number, number[]>} systemCallMap - System id -> call ids with scores.
  * @param {datacall[]} datacalls - All known data calls.
  * @param {number} activeDataCallId - The dashboard's active call id.
  * @param {number[]} [activeDatacallIds] - Call ids selected in the year picker.
--- a/src/views/Home/aggregateScores.test.ts
+++ b/src/views/Home/aggregateScores.test.ts
@@ -50,22 +50,69 @@ describe('buildDashboardMaps', () => {
     expect(scoreMap[7]).toBeUndefined() // no fake score
     expect(progressMap[7].questionsexpected).toBe(40)
     expect(progressMap[7].questionsupdated).toBe(0)
-    expect(systemCallMap[7]).toEqual([38])
+    // systemCallMap is score-derived: a never-started system lists no calls, so
+    // export provenance and the call picker don't treat it as having real data.
+    expect(systemCallMap[7]).toEqual([])
     expect(chosenCallMap[7]).toBe(38)
+  })
+
+  it('keeps the newest call for a never-started system present in multiple calls', () => {
+    // System 7 is never-started in BOTH calls with no score anywhere, so every
+    // idx ties at -1. With no scored call to prefer, the tie-break leaves the
+    // newest-first fallback intact (idxs[0] = 38) and systemCallMap stays empty.
+    const { scoreMap, progressMap, systemCallMap, chosenCallMap } =
+      buildDashboardMaps(
+        CALL_IDS,
+        [[], []],
+        [[prog(7, 0, null)], [prog(7, 0, null)]]
+      )
+    expect(scoreMap[7]).toBeUndefined()
+    expect(progressMap[7].questionsupdated).toBe(0)
+    expect(chosenCallMap[7]).toBe(38) // newest-first fallback
+    expect(systemCallMap[7]).toEqual([])
   })
 
   it('keeps a scored call as chosen when a never-started progress row exists in another call', () => {
     // System 1 is scored + completed in the newer call 38, and has a
     // never-started progress row (0 updated, null lastupdated) in the older call
-    // 3. The union now adds call 3 to its idxs, but that never-started row must
-    // not win `chosen` (its lastupdated is null -> -1) and blank the real score.
-    const { scoreMap, chosenCallMap } = buildDashboardMaps(
+    // 3. The union adds call 3 to its idxs, but that never-started row must not
+    // win `chosen` and blank the real score, and systemCallMap must stay
+    // score-derived ([38], not the union [38, 3]).
+    const { scoreMap, chosenCallMap, systemCallMap } = buildDashboardMaps(
       CALL_IDS,
       [[agg(1, 88)], []], // score only in call 38
       [[prog(1, 40, '2025-09-01')], [prog(1, 0, null)]]
     )
     expect(scoreMap[1].score).toBe(88) // scored call kept
     expect(chosenCallMap[1]).toBe(38)
+    expect(systemCallMap[1]).toEqual([38])
+  })
+
+  it('keeps the score when the older scored call has no progress rows', () => {
+    // Score only in the older call 3, but its /scores/progress fetch failed, so
+    // call 3 has no progress row -> lastUpdatedMs -1, tying with call 38's
+    // never-started row. Without the score-preference tie-break, chosen lands on
+    // 38 and the real score is lost.
+    const { scoreMap, chosenCallMap } = buildDashboardMaps(
+      [38, 3],
+      [[], [agg(1, 88)]],
+      [[prog(1, 0, null)], []]
+    )
+    expect(chosenCallMap[1]).toBe(3)
+    expect(scoreMap[1]?.score).toBe(88)
+  })
+
+  it('keeps the score when the scored call row has a null lastupdatedat', () => {
+    // Score in the older call 3, whose progress row has a null lastupdatedat (an
+    // imported call carries no edit events) -> -1, tying with call 38's
+    // never-started row. The tie-break must still keep the scored call.
+    const { scoreMap, chosenCallMap } = buildDashboardMaps(
+      [38, 3],
+      [[], [agg(1, 88)]],
+      [[prog(1, 0, null)], [prog(1, 12, null)]]
+    )
+    expect(chosenCallMap[1]).toBe(3)
+    expect(scoreMap[1]?.score).toBe(88)
   })
 
   it('lists a call once for a system in both its scores and progress', () => {
--- a/src/views/Home/aggregateScores.test.ts
+++ b/src/views/Home/aggregateScores.test.ts
@@ -36,6 +36,49 @@ describe('buildDashboardMaps', () => {
     expect(systemCallMap[2]).toEqual([3])
   })
 
+  it('includes a system present only in progress (no score row)', () => {
+    // A never-started system in the current call: /scores/progress returns a row
+    // (40 expected, 0 updated) but there is no aggregate score row yet. It must
+    // still land in progressMap so the "Not updated only" facet can see it -
+    // keying off scores alone dropped these systems entirely.
+    const { scoreMap, progressMap, systemCallMap, chosenCallMap } =
+      buildDashboardMaps(
+        CALL_IDS,
+        [[agg(1, 80)], []], // system 7 has no score row in either call
+        [[prog(1, 40, '2025-09-01'), prog(7, 0, null)], []]
+      )
+    expect(scoreMap[7]).toBeUndefined() // no fake score
+    expect(progressMap[7].questionsexpected).toBe(40)
+    expect(progressMap[7].questionsupdated).toBe(0)
+    expect(systemCallMap[7]).toEqual([38])
+    expect(chosenCallMap[7]).toBe(38)
+  })
+
+  it('keeps a scored call as chosen when a never-started progress row exists in another call', () => {
+    // System 1 is scored + completed in the newer call 38, and has a
+    // never-started progress row (0 updated, null lastupdated) in the older call
+    // 3. The union now adds call 3 to its idxs, but that never-started row must
+    // not win `chosen` (its lastupdated is null -> -1) and blank the real score.
+    const { scoreMap, chosenCallMap } = buildDashboardMaps(
+      CALL_IDS,
+      [[agg(1, 88)], []], // score only in call 38
+      [[prog(1, 40, '2025-09-01')], [prog(1, 0, null)]]
+    )
+    expect(scoreMap[1].score).toBe(88) // scored call kept
+    expect(chosenCallMap[1]).toBe(38)
+  })
+
+  it('lists a call once for a system in both its scores and progress', () => {
+    // System 1 appears in the score AND progress list for call 38; its call list
+    // must not double-count that call (the union is deduped per call index).
+    const { systemCallMap } = buildDashboardMaps(
+      [38],
+      [[agg(1, 80)]],
+      [[prog(1, 40, '2025-09-01')]]
+    )
+    expect(systemCallMap[1]).toEqual([38])
+  })
+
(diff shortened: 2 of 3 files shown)


danielbowne (changes requested):
Pulled this down and ran it locally. Full suite passes (749) and tsc is clean, and the union approach is the right call for the bug. One thing needs fixing before it goes in.

Building `systemIdxs` from the union means a system's `idxs` can now include a call it only has a progress row in, while `chosen` is still picked purely on `lastupdatedat`. The new comment says that's safe because a scoreless call always has a null `lastupdatedat`, which is true, but it misses the tie. `lastUpdatedMs` also returns -1 when the progress row is missing entirely, or when a real row's `lastupdatedat` is null. So a scored call can tie at -1 with a progress-only call, and the strict `>` leaves `chosen` at `idxs[0]`, the newest call. If the newest call is the progress-only one, the system's real score never lands in `scoreMap`.

Concrete path: a system scored in the older call 3 and never started in the newer call 38. If `/scores/progress?datacallid=3` fails, the per-request catch in Home.tsx returns `[]`, so call 3's progress lookup is undefined and ranks -1. Call 38's never-started row is also -1. `chosen` becomes 38, `scoreMap[1]` is never set, and a scored system reads on the dashboard as unscored and not updated. No fetch failure needed either if the scored call's progress row happens to carry a null `lastupdatedat`.

Two cases that pass on main and fail on this branch:

```ts
it('keeps the score when the older scored call has no progress rows', () => {
  const { scoreMap, chosenCallMap } = buildDashboardMaps(
    [38, 3],
    [[], [agg(1, 88)]], // score only in the older call 3
    [[prog(1, 0, null)], []] // call 3 progress missing (fetch failed)
  )
  expect(chosenCallMap[1]).toBe(3) // gets 38
  expect(scoreMap[1]?.score).toBe(88) // undefined
})

it('keeps the score when the scored call row has a null lastupdatedat', () => {
  const { scoreMap, chosenCallMap } = buildDashboardMaps(
    [38, 3],
    [[], [agg(1, 88)]],
    [[prog(1, 0, null)], [prog(1, 12, null)]]
  )
  expect(chosenCallMap[1]).toBe(3) // gets 38
  expect(scoreMap[1]?.score).toBe(88) // undefined
})
```

Neither could happen before the change, since `idxs` only ever held calls the system had a score row in, so this is new behavior rather than something pre-existing.

Smallest fix is to break the tie toward a call that actually has a score, which leaves the newest-first preference intact everywhere else:

```ts
let chosen = idxs[0]
let bestT = -Infinity
let bestHasScore = false
for (const idx of idxs) {
  const t = lastUpdatedMs(progressByCall[idx]?.get(sys))
  const hasScore = scoreByCall[idx]?.has(sys) ?? false
  if (t > bestT || (t === bestT && hasScore && !bestHasScore)) {
    bestT = t
    chosen = idx
    bestHasScore = hasScore
  }
}
```

Both cases above are worth adding to `aggregateScores.test.ts`. The existing no-regression test only covers the mirror situation where the scored call is the newer one, and that passes with or without the fix.

Everything else reads well. The per-call dedupe is correct, dropping the `lastUpdatedMs` seed really is a no-op since the loop re-reads `idxs[0]` on its first pass, and the doc corrections match what the map actually holds now.

danielbowne:
Second pass turned up another one, and this is the bigger of the two. Also a must fix.

Widening `systemIdxs` widens `systemCallMap` along with it, and two consumers read that map as "calls this system has real data in," not "calls it has a row of any kind in."

`/scores/progress?datacallid=X` returns a row for every system the caller can see in call X, and `Home.tsx` fetches it for every id in `activeDatacallIds`. Selecting a year toggles all of that year's calls, so multi-call is the normal state, not an edge case. Under the union, a system now lists every selected call it has a progress row in, which in a year view is most of them.

Two things break downstream:

**Download gets disabled.** `FismaTable.tsx:106-120` collects the selected rows' call ids and blocks the export when the set has more than one entry, since there is no single call to target. A system that previously listed exactly one call, and exported fine, now lists two or more, so `exportCallId` goes null and `exportBlocked` goes true. In a multi-call year view that is close to every selection.

**The #467 call picker fires when it should not.** `FismaTable.tsx:959-984` opens the picker when `rowCallObjs.length > 1` instead of navigating straight to the questionnaire. Rows that used to open directly now stop at a picker, and the picker lists calls the system never participated in.

The blast radius here is wider than it looks, because of the numbers in the PR description. For the open call, `scored = 0` and `notStarted = 27`, so every system with a progress row in the open call is progress-only there. Any of those that also carry a score in another call in the same year picks up an extra call id. That is the common case, not a corner.

The PR's own new test already shows the change, it just does not assert on it. In "keeps a scored call as chosen when a never-started progress row exists in another call", `systemCallMap[1]` is `[38, 3]` on this branch and `[38]` on main. Worth adding that assertion either way, so the map's contract is pinned by a test rather than by the docstring.

Fix is to stop overloading one map for two jobs. Keep `systemCallMap` score-derived, which is what export provenance and the row actions mean by it, and use the union only where it is actually needed, for `progressMap` and the rank. Building a separate `scoreIdxs` alongside the union covers it without touching either consumer.

Worth noting the doc change in this PR ("calls the system has scores in" to "appears in (score or progress)") makes the widening intentional-looking. If the map really is meant to carry both, then `FismaTable.tsx:106-120` and `:959-984` both need to filter down to score-bearing calls themselves, which is more moving parts than keeping the two sets separate at the source.

danielbowne (dismissed):
Both fixes look right. Repulled at a38db28 and verified against the branch rather than just reading the diff.

The tie-break does what it needs to. The two cases I described, a missing progress row for the scored call and a scored call whose row carries a null `lastupdatedat`, both used to blank the score and now keep it. The updated comment is also more honest than the one it replaces, since it names the actual triggers instead of asserting the tie cannot happen.

Splitting `scoreIdxs` out is the cleaner half of the fix. Checked the case that actually matters for export rather than just the never-started one: a system scored in a single call but carrying progress rows in both selected calls now resolves to one call id, so `exportCallId` stays non-null and Download works. Scored in both still gives both. Never-started gives an empty list while keeping its `chosenCallMap` entry and its -1 rank, so #655 stays fixed. Reverting the `rowCall.ts` docstring was the right call too, the map means what its consumers think it means again.

Checks on the branch: `yarn jest` 752 passed, 3 skipped, 0 failed. `tsc --noEmit` clean. Lint 0 errors, prettier clean. The two remaining warnings are in `InsightsPanel.tsx` and predate this work.

Approving.

One thing to be aware of before you merge, and it is your call what to do with it.

For a system with a progress row but no score in the selected call, the Data Call column and the questionnaire icon can resolve to different calls. The column goes through `resolveRowCallId` and lands on `chosenCallMap` (`FismaTable.tsx:899`). The icon reads `systemCallMap`, which is now correctly empty for that system, so it falls through to `rowCallObjs[0] ?? datacalls.find(d => d.datacallid === activeDataCallId)` (`FismaTable.tsx:988-989`). Those agree whenever the open call is the newest selected call, which covers the #655 scenario this PR is about. They diverge in a past-year view, where `activeDataCallId` points at the open call rather than anything on screen, so the column names the past call and clicking through opens the current one. That is the exact failure `rowCall.ts`'s own docstring warns about, the column and the thing it opens naming different calls.

Worth being clear that this is not something this PR introduced. It behaves the same way on main, because before your change those systems had no `chosenCallMap` entry either and the column fell through to the same `activeDatacallIds` rung. Impact is also limited, since `isNotUpdated` returns false for past calls, so the filter never surfaces these rows in the view where the divergence shows.

So it is genuinely pre-existing and low impact, and I checked, there is no open issue covering it in any of the ZT repos. Fixing it here would mean giving the icon the same `chosenCallMap` fallback the column uses, which is small but is scope this PR did not sign up for. Happy either way, fix it now or file it as a follow-up, as long as it does not get lost.

stu-aq:
One data-provenance note for the tie-break, since the test comments frame the null `lastupdatedat` as an edge case (a failed fetch, an imported call). From the load side: every FY23-25 ZTM score row entered through the bulk load rather than the app's write path, so no row in those calls has edit events, and their `lastupdatedat` is null by construction. The 0048 migration codified the same rule (imported provenance never counts as updated). So in any year view that includes a historical call, the -1 tie is the normal state for every system, not a corner, and the score-preference tie-break is the hot path for what those systems display. Worth keeping that in mind before any future refactor of the chosen-call loop, and worth the two tests staying pinned exactly as they are.

danielbowne:
Went and checked this against the backend, and it holds up. Two independent confirmations:

`scoreprogress.go:24` says history imported from outside the app "never had any (no events to backfill)". And the only thing feeding `lastupdatedat` is the `LEFT JOIN LATERAL` over `events` at `scoreprogress.go:214-221`, so with no event rows the `MAX(le.createdat)` is NULL. Migration 0048 lines up with both: the backfill only flips `status` to `done` for `action IN ('created', 'updated')`, so bulk-loaded rows stay `not_started` and report 0 updated.

So the framing in the test comments, mine included, undersold this. In any year view that includes a historical ZTM call, every system ties at -1 for that call, which means the score-preference branch is what decides the displayed call rather than a fallback that rarely fires. Agreed the two tests should stay pinned exactly as they are.

One thing worth adding, because it makes the "before any future refactor" warning sharper than it already is.

The two backend files disagree about whether imported data carries events at all. `0048scoresstatus.go:41-43` says data loaded outside the app "is attributed with 'imported' provenance events so it carries a who/when". `scoreprogress.go:24` says there were no events to backfill. Grepping the backend, nothing writes an `imported` action anywhere, so `scoreprogress.go` is the accurate one and 0048's comment describes an intent that was never implemented.

That matters here because the two queries filter differently. 0048's backfill is scoped to `action IN ('created', 'updated')`. The `lastupdatedat` lateral has no action filter at all, it takes the newest event of any action for that scoreid. So if anyone ever starts writing those `imported` provenance events, `status` correctly stays `not_started`, but `lastupdatedat` quietly becomes non-null for every historical row. The -1 ties disappear, this tie-break stops firing, and which call each system displays changes, with no frontend change and nothing failing.

Not something to handle in this PR. Probably worth a small backend issue to make the two agree, either by correcting 0048's comment or by adding the same action filter to the lateral, so the invariant this code now leans on is written down on the side that owns it.

tarratsco:
@danielbowne - thanks. Addressed the invariant note directly in code, and captured the two out-of-scope items as follow-ons so they don't get lost or hold the merge:

- In code - added a comment on the chosen-call tie-break stating the guarantee rests on the backend invariant "scoreless -> null lastupdatedat," with a note on where it would break if that changes.
- FE follow-on - questionnaire icon and the Data Call column naming different calls for a score-less system in a past-year view (pre-existing, low impact): CMS-Enterprise/ztmf-ui#662
- BE follow-on - the lastupdatedat lateral has no action filter while status/migration 0048 exclude imported events, so that invariant can drift on the backend side: CMS-Enterprise/ztmf#513

Both tickets are pre-existing and out of scope for #655.

danielbowne (approved):
Delta since a38db28 (the state I verified on-branch) is comment-only - five lines documenting that the chosen-call tie-break rests on the backend invariant "scoreless means null lastupdatedat," and naming where a score would get blanked if that ever changes. That matches what the backend check turned up (scoreprogress.go's lateral over events, migration 0048's provenance rule), so the comment is accurate, not aspirational. Follow-ons #662 and the backend issue capture the out-of-scope items cleanly. Approving at 52ba054.

danielbowne:

Issue #651, opened 2026-07-31
OpDivs cell is inline-editable for existing users but the edit is discarded and reported as saved


jsos3-cms:
## Description

The user grid's OpDivs column is inline-editable for existing users, but the row-commit handler never writes grants for an existing user. The edit is discarded, and the UI reports success anyway.

`src/views/UserTable/UserTable.tsx` declares the column with `editable: isAdmin`, which is not gated to newly-added rows, so an admin can open the inline editor on any existing user's OpDivs cell and change the selection.

`processRowUpdate` has two branches. The `newRow.isNew` branch does call `setUserOpDivs` (line 326) to persist grants for a user being created. The existing-user branch does not: it PUTs `email`, `fullname`, and `role` only. `setUserOpDivs` has exactly one call site in this file, in the create branch, so there is no path that writes grants for an existing row.

What makes it worse than a silent no-op is the feedback. After the PUT the handler sets `STATUS_MESSAGES.saved` (line 371) and then writes `updatedRow` into local state (line 379), so the grid optimistically shows the edited grants. The admin sees their change applied and a success message, and nothing was persisted. It reverts on the next refresh or navigation.

Found while reviewing #650, which touches the same cell. Not caused by that PR, and pre-dates it.

## Steps to Reproduce

1. Sign in as an admin who can edit users.
2. In the user table, double-click the OpDivs cell on an existing user.
3. Change the selection, adding or removing an OpDiv.
4. Commit the row edit.
5. Observe the success message and the grid showing the new value.
6. Refresh the page, or navigate away and back.

## Expected Behavior

Either the inline edit persists the grants, or the OpDivs cell is not editable for existing rows and the Assign OpDivs modal is the only way to change them.

Given the modal already handles caller-scope narrowing, the save-time preserve boundary, and the fail-closed behaviour when scope cannot be fetched (see #642 and #650), making the cell read-only for existing rows is the smaller and safer change. Reproducing that logic in an inline editor would duplicate a permissions boundary in a second place.

## Actual Behavior

The edit is silently dropped. The success message and the optimistic row update both fire, so the UI states that a change was saved which never reached the backend.

## Environment

Any browser. Present on `main` as of 2026-07-31, independent of #650.

## Acceptance Criteria

- [ ] An inline OpDivs edit on an existing user either persists, or is not offered.
- [ ] No success message is shown for a change that was not written.
- [ ] The grid does not optimistically display grants that were not persisted.
[...]

Issue #511, opened 2026-07-13
Questionnaire: let the data-call selector override a picker-opened call


danielbowne:
### Type
Enhancement (UX)

### Context
#501 / #506 made a picker-opened data call persist through in-survey navigation (Next, Back, sidebar, canonical redirect), which fixed the survey reverting to the latest call. A side effect: once you enter the questionnaire via the dashboard picker, the global Title data-call selector can no longer switch the questionnaire's call, because the persisted route state wins over the selector branch in the load effect.

### Current behavior (intended for now)
A call opened via the picker is authoritative on the questionnaire page. The Title selector remains a dashboard control. To view a different call for that system, re-open via the picker.

### Possible future behavior
Let an explicit change of the Title data-call selector switch the questionnaire's active call mid-survey, while still persisting the call through internal navigation.

### Why it isn't trivial
The load effect would need to distinguish a deliberate, user-initiated selector change from the selector's default value. Without that, on first load the default selector value could immediately override the call the user just picked. So it needs a small amount of "was this an explicit change since entry" tracking.

### Priority
Low / future. The picker covers the common need.

PR #663, opened 2026-08-03
fix(#661): unlock the ISSO Name control for write admins

### LINKED ISSUE #661: Unlock the ISSO Name control for write admins
### Description

The ISSO Name control is display-only for every role, so a system whose stored `isso_name` is stale cannot be corrected through the UI at all.

`fieldConfig.ts` marks the field `readOnly: true`. That single flag does two things: it renders the control disabled on both the System Details edit view and the admin Edit System modal, and it excludes the key from `EXTENDED_METADATA_KEYS`, which is the derived write list both PUT paths iterate. So the field is not merely disabled, it is absent from the request body.

The comment on that entry explains the original reasoning: the backend resolves the name, so the form edits `issoemail` instead and lets the name follow. That holds only where `isso_name` is NULL. Where the onboarding import populated it, the stored value shadows the email lookup and editing the email has no effect on the displayed name.

This depends on the backend carve-out and must not ship before it deploys. Unlocking the control while the backend still discards the write would produce a field that reports success and loses the change, which is the same defect as #651.

### Steps to Reproduce

1. Sign in as an admin who can edit systems
2. Open a system belonging to an OpDiv whose `isso_name` was seeded by the onboarding import
3. Click Edit and scroll to Extended Metadata
4. Observe ISSO Name rendered disabled with no way to change it
5. Correct the ISSO Email instead and save
6. Observe the ISSO Name still showing the previous person

### Expected Behavior

ISSO Name is editable by write admins, which is the existing `isWriteAdminRole` set: `OWNER`, `HHS_ADMIN`, `OPDIV_ADMIN`. The field is included in the write payload when edited. `ISSO`, `ISSM`, `SYSTEM_DELEGATE`, and the read-only admin tiers cannot change it.

Clearing the field is a supported action that restores the email-derived name, and the UI says so.

### Actual Behavior

The control is disabled for every role and the key never reaches the request body.

### Environment

all

### Acceptance Criteria

- [ ] ISSO Name renders as an editable text control for write admins on both the System Details edit view and the Edit System modal
- [ ] The field is included in the write payload when changed, and a cleared field sends the value that maps to NULL
- [ ] `ISSO`, `ISSM`, and `SYSTEM_DELEGATE` cannot edit it. The existing page-level `isAdmin` gate already covers this, so verify rather than add a second gate
- [ ] Read-only admin tiers cannot edit it
- [ ] Help text at the point of edit explains that a value entered here overrides the name derived from the ISSO's user record, and that clearing the field restores the derived name
- [ ] The mechanism that prevented a display-only field from silently rendering-but-not-saving still holds for any field that is genuinely read-only, so removing this one entry does not remove the guard for others
- [ ] Tests cover: the control is enabled for each write-admin role and disabled or absent for the non-admin tiers; an edited value appears in the PUT payload; a cleared value sends the clear signal

### Severity

Medium - Impacts usage but has a workaround

### Affected Component

Frontend

### Screenshots

_No response_

### Logs / Error Messages

```shell

```

### Workaround

None in the UI. Correcting the value requires a data-side change per system.

jsos3-cms:
Closes #661

## Merge ordering

Ready for review and for CI. The one constraint is on the **merge**, not on the review: the backend carve-out in CMS-Enterprise/ztmf#511 has to be deployed first. Merging this ahead of it would give an OpDiv admin an editable field, a success toast, and a discarded write, which is the same defect already tracked as #651. `#511` is approved and green, so the sequence is its merge and deploy, then this one.

Reviewing and deploying this branch to dev in the meantime is useful rather than just harmless. Dev's backend is currently running `#511`'s branch, since dev deploys from whichever PR branch last landed there, so the whole flow can be exercised end to end on dev right now: correct a name as an OpDiv admin, confirm it persists, clear it, confirm the derived name comes back. That holds until some other backend PR deploys to dev and replaces it.

One behavior change here is not scoped to this field and is worth a reviewer's attention: the post-save refetch below fires on **every** system save, not only an ISSO Name edit. It replaces an in-place state update with a list refetch, so each save now costs one additional GET.

## What was wrong

`fieldConfig.ts` marked the `isso_name` entry `readOnly: true`. That one flag did two things: it rendered the control disabled on both the System Details edit view and the Edit System modal, and it excluded the key from `EXTENDED_METADATA_KEYS`, the derived write list both PUT paths iterate. So the field was absent from the request body, not merely greyed out, and a stale ISSO name could not be corrected through the UI by anyone.

The original reasoning was sound when written: the backend resolves the name, so the form edited `issoemail` and let the name follow. That holds only where the stored column is NULL. Where the onboarding import populated it, the stored value shadows the email lookup and editing the email changes nothing visible.

## The change

Drop the flag from that one entry and add help text at the point of edit, since the semantics are not guessable: a name typed here becomes a stored override of the name derived from the ISSO's user record, and clearing the field is what restores the derived name.

`readOnly` stays a supported concept on `FieldConfig` and in the write-list derivation, so a genuinely backend-owned field added later still gets the same protection. No entry currently uses it, which is worth knowing if someone adds one.

**No new permission gate.** The whole detail-page edit surface is already gated on `isAdmin`, which is `OWNER` / `HHS_ADMIN` / `OPDIV_ADMIN` plus the legacy `ADMIN`, so the read-only admin tiers, `ISSO`, `ISSM`, and `SYSTEM_DELEGATE` cannot reach an editable system form at all. Tests pin that per role rather than adding a second gate that could drift from the first.

Worth knowing for anyone extending this: `EditSystemModal` has exactly one production render, in `Title.tsx`, with `mode` hardcoded to `create`. There is no production edit-mode caller, so it governs what can be set at system creation and edits go through the detail page.

## Second commit: the post-save refetch

A review pass found a defect this unlock exposed, worth reading as its own change.

The page wrote its own edited draft into shared state after a successful PUT. That is accurate for a field whose stored value is the value sent, which was every extended field until now. `isso_name` is the first one the backend resolves: clearing it stores NULL and the list read then derives the name, so the draft held an empty string where the saved system carries a name. The read view rendered the cleared field as an em dash, the exact opposite of what the new help text promises, and it affected `OWNER` and `HHS_ADMIN` immediately since they were never behind the backend gate.

The page now refetches instead. The PUT returns no body, and the single-system read returns the raw column rather than the derived name, so the list read is the only source that reflects what was saved. The refetch carries the caller's decommissioned view mode so saving does not change which systems the dashboard lists, and the fallback guard is cleared so a system excluded by that mode is re-added rather than leaving the page empty.

Two tests pin this. Both fail with the fix reverted and pass with it, which was verified rather than assumed.

## Known gap, deliberately not fixed here

For a **decommissioned** system viewed with the toggle off, clearing ISSO Name still displays an em dash. The refetch drops the system from the active-only list and the fallback re-populates it from the single-system read, which returns the raw column.

This is not a regression: before this branch the same scenario echoed the draft and rendered the same em dash. It is one path the fix does not resolve, and it cannot be resolved cleanly on the frontend, because no read available to the client returns a derived name for a system outside the current list view. Refetching with `decommissioned=true` returns only decommissioned systems, so it would change the dashboard as a side effect of saving.

The fix belongs in CMS-Enterprise/ztmf#510, where the read split is tracked and this is now recorded as evidence. Scope: needs a decommissioned system, the toggle off, and specifically clearing the field rather than typing a value, since a typed value displays correctly on that path. The stored data is correct throughout; only the view is stale and it self-corrects on reload.

## Test plan

- `npx tsc --noEmit`: clean
- 80 tests pass across `src/views/SystemDetailPage/` and `src/views/EditSystemModal/`, including 15 in a new `SystemDetailPage.issoName.test.tsx` covering all ten roles, the three write signals, and the refetch behavior
- eslint and prettier clean on the touched files
- the husky pre-commit ran the full suite: 763 passed, 3 skipped, no failures

Note for local runs: the full jest suite is flaky on some machines under parallel worker load, independent of any change. A bare `main` control run has produced more failures than a feature branch, disjoint and all timeout-shaped. Run the touched suites alone for the real signal, and control against `main` before attributing a failure to a change.

### DIFF
--- a/src/utils/systemMetadataVocab.test.ts
+++ b/src/utils/systemMetadataVocab.test.ts
@@ -223,6 +223,24 @@ describe('buildExtendedDiff', () => {
     expect(buildExtendedDiff(base, base, KEYS)).toEqual({})
   })
 
+  // isso_name is the one extended field whose displayed value can be resolved
+  // by the backend rather than stored, so the form is seeded with a name that
+  // is not in the column. Omitting it while untouched is what stops an
+  // unrelated edit from persisting the resolved name as a stored override.
+  it('omits an untouched isso_name that was resolved rather than stored', () => {
+    const keys = [...KEYS, 'isso_name'] as (keyof FismaSystemType)[]
+    const loaded = {
+      ...base,
+      isso_name: 'Resolved From User Record',
+    } as unknown as FismaSystemType
+    const edited = { ...loaded, fips: 'High' } as FismaSystemType
+
+    const diff = buildExtendedDiff(edited, loaded, keys)
+
+    expect(diff).toEqual({ fips: 'High' })
+    expect(diff).not.toHaveProperty('isso_name')
+  })
+
   it('includes only changed fields, at their typed value', () => {
     const edited = { ...base, fips: 'High', hva: false } as FismaSystemType
     expect(buildExtendedDiff(edited, base, KEYS)).toEqual({
--- a/src/utils/systemMetadataVocab.ts
+++ b/src/utils/systemMetadataVocab.ts
@@ -167,6 +167,13 @@ export function formatList(v: string[] | null | undefined): string {
  * array []) passes through as the user's clear. On create, pass an empty
  * baseline so only the fields the user set are sent.
  *
+ * Omission is what protects a field the backend resolves rather than stores:
+ * the System Details form is seeded from a read that returns a resolved
+ * isso_name, so an untouched field must stay out of the payload or saving an
+ * unrelated edit would persist the resolved name as a stored override. A
+ * missing baseline defeats that, since every field then counts as changed;
+ * pass an empty system on create and a real one on update, never null.
+ *
  * @param edited - The edited system.
  * @param baseline - The system to diff against (the loaded system, or an empty
  *   one when creating); a missing baseline treats every field as unset.
--- a/src/views/EditSystemModal/EditSystemModal.test.tsx
+++ b/src/views/EditSystemModal/EditSystemModal.test.tsx
@@ -140,6 +140,64 @@ test('an untouched extended field is omitted from the save (dirty-diff)', async 
   expect(putBody).not.toHaveProperty('cloud_service_model')
 })
 
+test('an edited ISSO Name is sent in the save payload', async () => {
+  mock.onGet('/systemattributes').reply(200, { data: [] })
+  let putBody: Record<string, unknown> | undefined
+  mock.onPut(/fismasystems\/42$/).reply((config) => {
+    putBody = JSON.parse(config.data)
+    return [200, {}]
+  })
+  const user = userEvent.setup()
+
+  renderWithProviders(
+    <EditSystemModal
+      title="Edit"
+      open
+      onClose={jest.fn()}
+      system={{ ...SYSTEM, isso_name: 'Conan Antonio Motti' }}
+      mode="edit"
+      datacenterEnvironments={[]}
+    />
+  )
+
+  const input = await screen.findByRole('textbox', { name: 'ISSO Name' })
+  await user.clear(input)
+  await user.type(input, 'Firmus Piett')
+  await user.click(screen.getByRole('button', { name: 'Save' }))
+
+  await waitFor(() => expect(putBody).toBeDefined())
+  expect(putBody).toHaveProperty('isso_name', 'Firmus Piett')
+})
+
+test('clearing ISSO Name saves an empty string, which restores the derived name', async () => {
+  mock.onGet('/systemattributes').reply(200, { data: [] })
+  let putBody: Record<string, unknown> | undefined
+  mock.onPut(/fismasystems\/42$/).reply((config) => {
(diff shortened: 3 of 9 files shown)


danielbowne (dismissed):
Verified the two things that could bite: the dirty-diff baseline is the resolved-name prefill, so an untouched field compares equal and is omitted from the PUT - a derived name cannot be persisted by accident, which keeps #510's edit-form constraint honored without a second gate. And the refetch-over-echo change is correct now, not just for this field - both pinned tests genuinely fail with it reverted.

One update to the "known gap" section: it is already closed, by CMS-Enterprise/ztmf#514. That PR (stacked on #511) gives the single-system GET the same COALESCE resolution the list applies, opt-in for display reads only, so the decommissioned-fallback path here starts returning the derived name once it deploys - the em dash case disappears with no frontend change. The "no read available to the client returns a derived name" premise goes stale at that point; nothing in this PR needs to change for it.

Merge order stands as you wrote it: #511 deploy, then this. #514 rides behind and closes the gap.

danielbowne:

Issue #515, opened 2026-07-13
InsightsPanel: migrate hardcoded hex colors to the CMS/MUI theme


danielbowne:
### Type
Tech debt (follow-up from #514 review)

### Context
The ZTMF Insights panel (`src/views/QuestionnairePage/InsightsPanel/InsightsPanel.tsx`) uses hardcoded hex values (e.g. `#f8f9fe`, `#5666b8`, `#b02a37`) rather than the app's MUI/CMS theme palette. It came from a prototype-derived trial component, which is why.

### Why it matters
We stick to the CMS light theme (no dark mode planned), so this is not urgent, but the hardcoded values can drift from the CMS palette as the theme evolves, and contrast is managed by hand rather than by the theme. If the rest of the app themes via the MUI palette, this panel will diverge over time.

### Ask
Post-trial, migrate the panel's colors to theme tokens / the CMS palette so it tracks the rest of the app. Low priority.

### Note
Do this on the impl branch (where Insights is actively iterated) so it converges back to main on the next refresh, rather than diverging main from impl.

danielbowne:
Correction on the note above: with Insights now shipped and the working model settled, routine insights UI work happens in the main/prod line, not impl. So this can just be done here on a normal feature branch off main. Disregard the 'do on impl to converge' bit.

PR #664, opened 2026-08-03
fix(#659): carried-forward badges, explicit confirm, Complete summary, progress wording

### LINKED ISSUE #659: [P0] Carried-forward answers look answered but never register - ISSOs believe they finished when they haven't
**What happened**

An ISSO (a CMS system) reviewed all 40 questions, agreed with the pre-filled answers, and reported completing the questionnaire. The dashboard shows 14/40. Event logs confirm both halves: 160+ questionnaire views, exactly 14 saves.

**Why**

FY2026 pre-fills each question with the prior year's answer (`status = 'not_started'`). A question only counts as updated once it is re-saved this cycle (`status = 'done'`) - that part is by design.

The gap sits on top of #413. Before #413, merely paging through a questionnaire recorded the viewer as an editor (`last_edited_by`) on every question they passed, so #413 correctly made Next skip the save when nothing changed. The side effect: "review and agree" - the most common action during a data call - now performs no write at all, and the UI gives no indication the question is still unconfirmed. The user's only feedback is a progress number that appears to contradict what they just did.

Two hard requirements are in tension, so neither can simply be reverted:

- Browsing must stay write-free (#413's guarantee - viewers must never show up as editors).
- Agreeing must produce a write, or progress never moves.

The resolution is that agreement has to be an explicit act, distinct from navigation.

This is the second report in one day caused by this gap (see also the Complete-button loop report). Expect it to repeat across most of the ~900 systems with carried-forward answers.

**Fix direction**

1. **Visible state per question.** Carried-forward, unconfirmed questions get a clear badge ("Carried forward - not yet confirmed") in both the sidebar and the question view, so the remaining work is visible at a glance instead of only as a count.

2. **Explicit confirm, inline - not a dialog.** On a carried-forward question, show a primary action alongside the answer: "Confirm this answer is still accurate." Clicking it saves the unchanged answer (flips `status` to `done`) and swaps the badge to a confirmed state. Next remains pure navigation and never writes - preserving #413.
   - Inline button, not a per-question "are you sure?" modal: with 40 questions, interrupting dialogs add ~40 clicks of pure friction and train users to dismiss without reading. The confirmation IS the button - one deliberate click, immediate visible state change as feedback.
   - Changing the answer keeps today's behavior: an edit is already an explicit act and saves via Next as it does now.
   - Button must be keyboard-reachable and the state change announced (aria-live) for 508.

3. **End-of-questionnaire summary.** On the last question / Complete, if unconfirmed questions remain, show a summary ("12 of 40 confirmed - 28 carried-forward answers still need confirmation") with links to jump to each. This is the single "are you sure" moment, placed where it has context, instead of 40 small ones. Ties into the Complete-button issue.

4. **Progress wording.** Distinguish "answered last year, awaiting confirmation" from "unanswered" so 14/40 stops reading as data loss.

**Where**

`src/views/QuestionnairePage/QuestionnairePage.tsx` - the unchanged-skip from #413 and the Next/save path. Backend counting (`scoreprogress.go`) is correct and should not change.

costacalvin:
Closes #659

## Problem

Carried-forward answers looked answered but never registered. Browsing is write-free by design (#413 — viewers must never be stamped as editors), so "review and agree" performed no write, and nothing in the UI showed the outstanding work. The reporting ISSO reviewed all 40 questions and saw 14/40 with no explanation; view events confirmed full coverage and exactly 14 saves.

## Changes, mapped to the issue's four items

**1. Visible state per question.** Unconfirmed carried answers get a warning chip ("Carried forward — not yet confirmed") in the question view and a "Not yet confirmed" marker in the sidebar; rows counted as updated show the confirmed state. Everything derives from the persisted `scores.status` — the same fact the Data Call Progress fraction counts — through one pure classifier (`confirmState.ts`), so the badges, the button, the summary, and the dashboard number cannot disagree. Open call only: historical rows are legitimately `not_started` forever. Read-only users see the badges (that's the triage information) but never the button.

**2. Explicit confirm, inline — not a dialog.** A review strip sits between the content and the navigation — the same zone where the CMS prior-response flow surfaces its "review before continuing" message, so every system shares one review area. The strip pairs the state chip with a small outlined "Confirm this answer is still accurate" button: one deliberate click issues `PUT scores/{id}/confirm` and the chip swaps in place (`role="status"`, so the change is announced — 508). The button yields the moment the question is dirty (an edit is its own explicit act and Next saves it, flipping status server-side), while a required prior-response review is pending, and for read-only sessions. **Next remains pure navigation and never writes on an untouched question** — pinned by a dedicated regression test.

The CMS insights flow keeps its shape (Insert only edits the field, exactly like the suggestion card's Insert) — but a resolved required review now routes through the confirm endpoint instead of the identical-body answer PUT the backend's no-op guard silently dropped. The whole insert-and-continue ritual finally counts.

**3. End-of-questionnaire summary.** Complete on an open call opens a summary — "N of M answers counted as updated" — listing carried-forward-unconfirmed and unanswered questions as separate groups with jump links, instead of silently looping back to question 1 (also retires the Complete-button loop report). When nothing remains it's the success state. Closed calls keep the old wrap-around (harmless paging for historical viewers).

**4. Progress wording.** The dashboard's Data Call Progress chip splits "Awaiting confirmation" (answers exist, none updated) from "Not started" (no answers), with matching tooltips, so `0/40` stops reading as data loss. The legacy "Not updated" wording is preserved when the backend doesn't serve `questionsanswered`.

## Degrade behavior

Against a backend that doesn't serve `scores.status` yet: no badges, no Confirm button, no summary counts (row-presence facts like "unanswered" still render), and the CMS prior-review falls back to its previous save path. Nothing misrenders during a staggered deploy.

## Out of scope

The dashboard em-dash for systems with **no score rows at all** (523 live systems) is #655's fix — its root cause is `buildDashboardMaps` building its universe from scores only, and the map change carries its own regression surface (Data Call column, per-row call picker). This PR delivers the wording half of the related comment; #655 delivers the rendering half.

## Testing

- New pure-helper suite (`confirmState.test.ts`, 16 tests): classification × status × call-state × row presence, degrade cases, summary bucketing in sidebar order
- `QuestionnairePage.test.tsx`: badge in view + sidebar; Confirm issues exactly one dedicated PUT and no answer write, chip flips after refetch; **Next write-free on untouched carried questions (#413 guard)**; button yields when dirty / read-only / pending review; no treatment on closed calls; CMS insert→navigation lands via confirm with no identical-body PUT; Complete summary with working jump links; success state with no loop and no manufactured writes
- `progressColumn.test.tsx`: three-way wording, absent-field fallback, tooltip split
- Full suite: 773 passing; lint clean; no new type errors

## UI

**Review strip on a carried-forward question (non-CMS):** the state chip and the explicit confirm action, between the justification field and the navigation.

image (18)

**After confirming / saving:** the chip flips to "Updated this data call" in place, and the sidebar marker follows.

image (16)

**CMS insights flow, unchanged shape:** the prior-response card still gates with "Review required"; the strip below shows the same carried-forward state every other system gets.

image (15)

**Complete summary:** the single "are you sure" moment — carried-forward-unconfirmed and unanswered as separate groups with jump links.

image (17)

**Dashboard progress wording:** "Awaiting confirmation" instead of "0/40 Not updated" for carried systems, with the matching tooltip.

image (19)

## Acceptance criteria

From the issue body:

- [x] **Item 1 — visible state per question**: "Carried forward — not yet confirmed" chip in the question view + "Not yet confirmed" sidebar marker, both driven by `scores.status` through one classifier (`confirmState.ts`), text-bearing rather than color-only
- [x] **Item 2 — explicit confirm, inline, not a dialog**: one click saves the unchanged answer (`status` → `done`) and swaps the badge to its confirmed state in place; native button (keyboard-reachable); the chip is a `role="status"` live region so the change is announced (508); changing the answer keeps today's behavior (edit saves via Next)
- [x] **Item 2 — Next remains pure navigation and never writes**: pinned by a dedicated regression test (`keeps Next write-free on an untouched carried-forward question (#413)`)
- [x] **Item 3 — end-of-questionnaire summary**: Complete shows "N of M answers counted as updated" with carried-forward-unconfirmed and unanswered groups, each entry a jump link to the question
- [x] **Item 4 — progress wording**: "Awaiting confirmation" (answered last year, awaiting confirmation) distinguished from "Not started" (unanswered), so 0/40 stops reading as data loss

From the acceptance comment:

- [x] **Retro-serves affected users with no data migration**: the badge renders from nothing beyond the existing `not_started` state; the live 26-row case badges immediately on deploy
- [x] **Confirm flips rows to `done` with the confirming user's own identity and timestamps** (via the backend's events hook; pinned by `TestConfirmScoreIntegration`)
- [~] **Dash renders as 0/40**: deferred to #655 (its root cause is the dashboard map construction, out of this PR's scope); the carried-forward-unconfirmed vs not-started distinction from the same comment is delivered here

## Deploy order

Requires CMS-Enterprise/ztmf#515 (scores.status on reads + the confirm endpoint) — deploy backend first.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -233,13 +233,7 @@ export type QuestionScores = {
   status?: ScoreStatus
   // Present when fetched with ?include=functionoption; functionid maps the
   // row back to its question.
-  functionoption?: {
-    functionoptionid: number
-    functionid: number
-    score: number
-    optionname: string
-    description: string
-  }
+  functionoption?: QuestionOption
   last_edited_at?: string | null
   last_edited_by?: LastEditedBy | null
   notes_is_ai_summary?: boolean
--- a/src/types.ts
+++ b/src/types.ts
@@ -213,6 +213,14 @@ export type LastEditedBy = {
   role?: UserRole
 }
 
+/**
+ * The answer's review state for its data call (scores.status): 'not_started'
+ * = carried forward and untouched this cycle; 'done' = saved or confirmed
+ * this cycle. The same persisted fact the Data Call Progress count reads, so
+ * badges derived from it cannot disagree with the fraction.
+ */
+export type ScoreStatus = 'not_started' | 'done'
+
 export type QuestionScores = {
   scoreid: number
   fismasystemid: number
@@ -220,6 +228,18 @@ export type QuestionScores = {
   notes: string
   functionoptionid: number
   datacallid: number
+  // Optional: a backend that does not serve it degrades to no carried-forward
+  // badges rather than badging everything.
+  status?: ScoreStatus
+  // Present when fetched with ?include=functionoption; functionid maps the
+  // row back to its question.
+  functionoption?: {
+    functionoptionid: number
+    functionid: number
+    score: number
+    optionname: string
+    description: string
+  }
   last_edited_at?: string | null
   last_edited_by?: LastEditedBy | null
   notes_is_ai_summary?: boolean
--- a/src/views/FismaTable/progressColumn.test.tsx
+++ b/src/views/FismaTable/progressColumn.test.tsx
@@ -261,4 +261,49 @@ describe('ProgressCell', () => {
     expect(screen.queryByText('Complete')).not.toBeInTheDocument()
     expect(screen.queryByText('Not updated')).not.toBeInTheDocument()
   })
+
+  // --- Carried-forward wording on the current call ---
+
+  it('reads Awaiting confirmation for a current-call system with carried answers and zero updates', () => {
+    // The answers exist (carried forward) but none count as updated —
+    // "0/41 Not updated" read as data loss; the chip says the actual state.
+    render(
+      <ProgressCell
+        entry={{ ...untouchedEntry, questionsanswered: 41 }}
+        isCurrentCall={true}
+      />
+    )
+    expect(screen.getByText('0/41')).toBeInTheDocument()
+    expect(screen.getByText('Awaiting confirmation')).toBeInTheDocument()
+    expect(screen.queryByText('Not updated')).not.toBeInTheDocument()
+  })
+
+  it('reads Not started for a current-call system with no answers at all', () => {
+    render(
+      <ProgressCell
+        entry={{ ...untouchedEntry, questionsanswered: 0 }}
+        isCurrentCall={true}
+      />
+    )
+    expect(screen.getByText('0/41')).toBeInTheDocument()
+    expect(screen.getByText('Not started')).toBeInTheDocument()
+    expect(screen.queryByText('Not updated')).not.toBeInTheDocument()
+  })
+
+  it('keeps the legacy Not updated wording when questionsanswered is absent', () => {
+    // ztmf#437 may not be deployed everywhere; without the field the split
+    // would be a guess, so the cell keeps saying what it always said.
+    render(<ProgressCell entry={untouchedEntry} isCurrentCall={true} />)
+    expect(screen.getByText('Not updated')).toBeInTheDocument()
+  })
+
+  it('describes the carried-forward state in the tooltip', () => {
+    expect(progressTooltip({ ...untouchedEntry, questionsanswered: 41 })).toBe(
+      'Answers carried forward from a previous data call — not yet confirmed'
+    )
+    // No answers at all keeps the plain no-updates line.
+    expect(progressTooltip({ ...untouchedEntry, questionsanswered: 0 })).toBe(
+      'No updates this data call'
+    )
+  })
 })
--- a/src/views/FismaTable/progressColumn.tsx
+++ b/src/views/FismaTable/progressColumn.tsx
@@ -103,6 +103,19 @@ export function ProgressCell({
     )
   }
   const updated = entry.questionsupdated > 0
+  // Zero updates splits two materially different states: carried-forward
+  // answers awaiting confirmation vs no answers at all — a blanket "Not
+  // updated" read as data loss to ISSOs who had just reviewed everything.
+  // questionsanswered may be absent until ztmf#437 is deployed everywhere;
+  // keep the legacy wording then rather than guessing.
+  const answered = entry.questionsanswered
+  const label = updated
+    ? 'Updated'
+    : answered == null
+      ? 'Not updated'
+      : answered > 0
+        ? 'Awaiting confirmation'
+        : 'Not started'
   return (
     <Tooltip title={progressTooltip(entry)}>
       <Box
@@ -117,7 +130,7 @@ export function ProgressCell({
         </span>
         <Chip
           size="small"
-          label={updated ? 'Updated' : 'Not updated'}
+          label={label}
           color={updated ? 'success' : 'warning'}
           variant="outlined"
         />
--- a/src/views/FismaTable/progressColumn.tsx
+++ b/src/views/FismaTable/progressColumn.tsx
@@ -26,8 +26,9 @@ import { hasNoQuestionnaire, progressTooltip } from './progressHelpers'
  *     "N/A" chip, not an orange "Not updated" one - it is not a laggard,
  *     there is nothing to nudge;
  *   - current call, any genuine edit: "Updated" (green);
- *   - current call, otherwise: "Not updated" (orange). The chip is derived from
- *     questionsupdated so it can never disagree with the fraction.
+ *   - current call, zero updates: "Awaiting confirmation" (carried answers
+ *     exist) or "Not started" (none do), both warning-colored laggards;
+ *     legacy "Not updated" when questionsanswered is not served.
  * @param {object} props - Component props.
  * @param {ScoreProgress | undefined} props.entry - The system's progress row;
  *   undefined renders an em-dash (progress fetch failed or not covered).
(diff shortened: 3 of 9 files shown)


jsos3-cms (approved):
## Verdict

This holds up. I verified the load-bearing invariant rather than taking the test's word for it, and it is sound: Next never writes on an untouched question, and neither does any other exit path. Nothing here blocks. Everything below is a follow-up.

## What I verified, not just read

**The write-free guarantee survives on every path out of a question, not only the one the test covers.** Next and Complete both funnel through `saveResponse()`, which returns without a request when the answer is unchanged. Back and the sidebar jump are structurally different: on a clean question they navigate without ever calling `saveResponse()`, and on a dirty one the confirm dialog discards the draft rather than saving it. A write is not reachable from either handler. In-survey navigation uses `replace: true`, so browser Back exits the questionnaire instead of stepping back a question, and the unmount effect writes only to localStorage. The one path that does write on an otherwise untouched answer is the CMS resolved-prior-review flow, and it goes through the confirm endpoint rather than an identical-body answer PUT, which is the right call.

**The chip is a real live region.** The gate at `QuestionnairePage.tsx:1813` admits both `unconfirmed` and `updated`, so the `role="status"` node at `:1824` stays mounted across the confirm transition and only its label and color change. That is the pattern that actually announces. A conditionally mounted live region does not, and this avoids it. The chips carry text rather than relying on color.

**Failure does not lie.** The optimistic flip to `done` sits inside the `try` and after the `await`, so a 403 or 404 leaves the chip in its prior state and surfaces an error. Claiming success on a failed write would have been worse than the bug this PR fixes, and it does not happen.

**The type contract is exhaustive.** `ScoreStatus` matches the database CHECK constraint on `scores.status` exactly, so the union cannot miss a value the backend can emit. Since the column is `NOT NULL`, the only reachable degrade case is the field being absent against a backend that predates ztmf#515, and that path correctly produces no badges rather than badging everything.

**Test numbers check out.** 773 passing on this branch against 746 on bare main, so +27 with no regressions, and no flaky failures in either run. `tsc --noEmit` clean, `eslint` reporting only two pre-existing warnings in a file this PR does not touch. Worth stating explicitly because this repo runs no lint or test CI on pull requests, so those numbers are otherwise unverified by anything reviewers can see.

## Follow-ups, none blocking

**1. The hardening commit moved three guards without tests.** 8b5840a5 added the Complete-flow re-seed that stops a second Complete click from double-posting, the `pastCall` parameter that suppresses the confirmation tooltip line on a past call, and `jumpToSummaryEntry`'s early return when the target is the question already open. All three are correct by inspection. None has a test, and `pastCall` appears twice in `progressHelpers.ts` and zero times in `progressColumn.test.tsx`. These are the edges most likely to regress quietly later, since a reader has no failing test to tell them the guard mattered.

**2. The filter label will disagree with the new chip wording.** This PR reserves "Not updated" for the legacy fallback and splits the live wording into "Awaiting confirmation" and "Not started". The filter checkbox in `FismaTable.tsx:427` is still labeled "Not updated only". Once this and #657 both land, checking that filter surfaces rows whose chips no longer say "Not updated". No code conflict, and #657's counting change is in a file this PR does not touch. It is a copy change, but it lands in the same place a confused user is already looking.

**3. `confirmState.ts:24` classifies anything that is not `not_started` as updated.** Equivalent to `=== 'done'` today, because the CHECK constraint permits exactly two values. It is worth naming only because migration 0048's own comment says a third per-answer state was deliberately deferred and that varchar plus CHECK was chosen specifically to make widening easy. If that state arrives, this branch counts it as updated while the progress query counts only `done`, which reintroduces exactly the badge-versus-count disagreement this PR exists to remove. Comparing against `'done'` directly costs nothing and removes the trap.

**4. One accuracy note on the PR body.** The "Degrade behavior" section reads as though the whole change is gated on ztmf#515. The dashboard wording is not: it is driven by `questionsanswered`, which has been live since ztmf#437. Only the questionnaire badge, the Confirm button, and the summary counts depend on #515. Narrowing that sentence would make the deploy-order note easier to act on.

## Also worth knowing

The double-click guard on Confirm combines a `confirming` flag check with `disabled={confirming}`. No test exercises rapid double invocation, and the backend's confirm is idempotent on an already-`done` row, so the worst case is a redundant audit restamp rather than a correctness problem. Low priority.

Merge coordination: this merges clean against all four other open PRs in this repo, verified by squash-then-merge simulation on current heads, so there is no ordering dependency inside ztmf-ui. The only ordering that matters is the backend one already noted in the PR body.

danielbowne:
Tested this locally against the paired frontend branch, with the API container rebuilt on this branch and a full carry-forward fixture (~38.8k not_started rows in the open call).

Exercised as two different ISSOs on two different systems - one CMS system on the CMS-Cloud-AWS question edition, one HRSA system on the MAG edition - covering both the API directly and the UI flow. Confirm flips the row to done, the status now surfaces on the score read, progress moves off 0/40, the dashboard chip flips from Not updated to Updated, and the carried-forward score is preserved rather than blanked. Attribution landed on the confirming user in both cases.

Looks good end to end. The Confirm audit-guard point above is the only thing I would hold for.

danielbowne:

PR #670, opened 2026-08-04
Adjust required fields for system edit form to allow saving with blank values 

### LINKED ISSUE #667: System edit form requires fields the backend does not, blocking edits on every non-CMS system
### Description

Editing any single field on a system re-validates the entire record against the form's required-field rules. For systems created by the onboarding load rather than through the UI, those rules were never satisfied, so an unrelated edit is blocked until the user invents values for fields that may not apply to them.

On current production-shaped data this blocks **every live non-CMS system** — 1,196 of 1,196. CMS is largely unaffected at 21 of 234.

Reported by an OpDiv admin reassigning ISSOs: changing the ISSO on a system also demanded `component` and `datacallcontact` before it would save.

### Where the requirement lives

Frontend only. `src/views/SystemDetailPage/fieldConfig.ts` marks seven fields `required: true`. The backend's `FismaSystem.validate()` requires exactly one — `fismauid` — and only format-checks `issoemail` / `datacallcontact` when they are present.

So the form is stricter than the system of record, and stricter than the data it is editing. The onboarding load satisfied the real contract; the UI added its own on top.

### Fill rates on live systems

| field | non-CMS (1196) | CMS (234) |
|---|---|---|
| `component` | 148 (12%) | 230 (98%) |
| `datacallcontact` | 0 (0%) | 216 (92%) |
| `issoemail` | 1158 (97%) | 232 (99%) |
| `fismaname` / `fismaacronym` / `fismauid` | 1196 (100%) | 234 (100%) |
| `datacenterenvironment` | 1196 (100%) | 234 (100%) |

`datacallcontact` blocks all 1,196 on its own — it was never collected outside CMS. `component` is a CMS organizational concept (OIT, CM, ...) with no HHS OpDiv equivalent, so its 12% fill is not sparse data; it is a field that does not apply.

### Proposed fix

**Keep hard-required (no change to these):** `fismaname`, `fismaacronym`, `fismauid`, `datacenterenvironment`.

`datacenterenvironment` in particular should stay required. The questions endpoint inner-joins through the environment mapping, so a system without one renders an empty questionnaire. It is 100% filled today and should stay that way.

**Change — to be explicit, this means removing `required: true` from the field's entry in `fieldConfig.ts`. No field is being deleted. Every field below stays on the form, stays editable, stays saved, and stays displayed. The only thing that changes is that a blank value no longer blocks the save.**

- `component` — remove `required: true`. Decided: not conditional. Gating it on the system's OpDiv would mean hardcoding a specific agency into the form, and this repo is agency-neutral by design — the tool is open source for any agency, so tenant literals do not belong in it. The field remains on the form for anyone who wants to fill it in; it just stops blocking saves.
- `datacallcontact` — remove `required: true`. Field stays.
- `issoemail` — keep it required when creating a system, but a missing stored value must not block an edit to an unrelated field. Field stays.

The underlying rule worth encoding: **required-on-create is not the same as required-to-edit.** The form currently conflates them, which is what makes every onboarding-loaded system uneditable.

### Impact

Every non-CMS OpDiv admin, on every system edit, for the duration of the data call. There is a manual workaround — fill in the missing fields — but it asks people to invent values for fields that do not apply to their OpDiv, which puts wrong data in the record.

tarratsco:
Summary

Closes #667.

The system edit form re-validated the entire record against its create-time required-field rules on every single-field edit. Systems onboarded outside CMS arrive without a Component, Data Call Contact, or ISSO Email, so those blank stored values failed validation and froze the whole record: an OPDIV admin could not change the ISSO without first backfilling fields the system was never given. This blocked edits on every non-CMS system.

Root cause: the required-field rules live in two places that both have to agree before Save enables. fieldConfig.ts drives the edit view's rendering (error state, which validator runs), and a separate hardcoded formValid object in SystemDetailPage.tsx is the actual save gate (Object.values(formValid).every(...)). component, datacallcontact, and issoemail were required in both, so a blank value in any of them disabled Save regardless of what the user actually edited.

The distinction the fix encodes: required-to-create is not the same as required-to-edit. A system already exists by the time it reaches this form, so a value it was onboarded without must not freeze an unrelated edit.

Changes

- fieldConfig.ts: set component, datacallcontact, and issoemail to required: false. The two email fields switch to the optional validator (blank is valid, a typed value is still format-checked) and lose the required styling. Documented the required flag as "required to edit, not to create."
- SystemDetailPage.tsx: drop the same three keys from the formValid save gate (initial state and edit-mode seed), so a blank stored value no longer disables Save. The four fields the backend needs on every record stay required.

Acceptance criteria

- [x] fismaname, fismaacronym, fismauiemain required to edit.
- [x] component and datacallcontact no longer block the save; they stay on the form, editable, saved,
and displayed.
- [x] A missing stored issoemail no longer blocks an edit, but a typed value is still format-checked.
- [x] Creating a system still requires modal is a separate hardcoded surface,used in production only in create mode, and is untouched).
- [x] Editing an unrelated field on a ut backfilling the blank fields.

Testing

- New SystemDetailPage.requiredFields.e with all three fields blank; editingone field saves without filling the blanks; clearing a hard-required field still blocks; a malformed ISSO Email still blocks (optional is n
- Full SystemDetailPage suite green (79 passed), tsc --noEmit and eslint clean.
- Behavior-only fix; no visual change r rendering as required.

### DIFF
--- a/src/views/SystemDetailPage/SystemDetailPage.requiredFields.test.tsx
+++ b/src/views/SystemDetailPage/SystemDetailPage.requiredFields.test.tsx
@@ -1,0 +1,180 @@
+// A blank Component, Data Call Contact, or ISSO Email must not block editing an
+// unrelated field on a system onboarded without those values; the four fields
+// the backend needs on every record still gate the save.
+
+// axiosConfig reads import.meta.env at module load and throws under @swc/jest.
+// Swap in a bare axios instance the MockAdapter can drive.
+jest.mock('@/axiosConfig', () => {
+  const axios = require('axios').default
+  return { __esModule: true, default: axios.create({ baseURL: '/api/v1/' }) }
+})
+jest.mock('@/utils/notify', () => {
+  const actual = jest.requireActual('@/utils/notify')
+  return { ...actual, notify: jest.fn() }
+})
+
+// The page reads its shared state via useContextProp (useOutletContext).
+let mockCtx: Record<string, unknown>
+jest.mock('../Title/Context', () => ({
+  useContextProp: () => mockCtx,
+}))
+
+import { screen, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { Route, Routes } from 'react-router-dom'
+import MockAdapter from 'axios-mock-adapter'
+import SystemDetailPage from './SystemDetailPage'
+import axiosInstance from '@/axiosConfig'
+import { renderWithProviders } from '@/test-utils/renderWithProviders'
+import type { FismaSystemType, userData } from '@/types'
+
+const mock = new MockAdapter(axiosInstance)
+
+// The four required fields are present; Component, Data Call Contact, and ISSO
+// Email were never collected.
+const NON_CMS_SYSTEM = {
+  fismasystemid: 42,
+  fismaname: 'Rebel Alliance Fleet Command',
+  fismaacronym: 'RAFC',
+  fismauid: 'UID-42',
+  component: '',
+  datacenterenvironment: 'CMS-Cloud-AWS',
+  issoemail: '',
+  datacallcontact: '',
+  opdiv_id: 1,
+  decommissioned: false,
+  sdl_sync_enabled: false,
+  isso_name: '',
+  hva: null,
+  fips: null,
+  system_type: null,
+  cloud_system: null,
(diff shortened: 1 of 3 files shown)


MackOverflow (approved):
lgtm!

jsos3-cms:

Issue #669, opened 2026-08-04
Questionnaire sidebar hides most questions and gives no sense of place


costacalvin:
### Description

The questionnaire sidebar lists roughly 40 questions in a narrow left column capped at `calc(100vh - 240px)` — about 10 rows on a typical window. Three compounding problems make the rest effectively invisible and leave users without a sense of place:

1. **It hides its own scrollbars** on every engine (`msOverflowStyle: 'none'`, `::-webkit-scrollbar { display: none }`, and `scrollbarWidth: 'none'`), so nothing indicates there is more content below the fold. Users only discover the remaining ~30 questions by scrolling on a surface that does not look scrollable.
2. **The selection never follows navigation.** Paging with Next/Back updates the highlight but never scrolls it into view — there is no `scrollIntoView` call anywhere in the repo — so after a few questions the active row is off-screen and the sidebar stops reflecting where the user is.
3. **The selected row is nearly indistinguishable**, carrying only the theme's `rgba(0, 0, 0, 0.04)` tint.

Net effect: users cannot tell how many questions exist, where they are in the set, or how much work remains — which also undercuts the per-question confirmation markers added in #659.

### Fix

All within the sidebar block of `QuestionnairePage.tsx`.

1. **Visible scrollbar.** Drop the three hiding rules; add a thin styled scrollbar scoped to this list (8px `::-webkit-scrollbar` with a rounded thumb that darkens on hover and a transparent track, plus `scrollbarWidth: 'thin'` and `scrollbarColor`). Both rule sets must stay visually identical — Chrome 121+ honors `scrollbarColor` and then ignores the `-webkit-` rules. Kept local rather than theme-level, since the theme has no scrollbar rules today.

2. **Auto-scroll the active question into view.** Give each row a stable DOM id and add one effect keyed on the selected question: center the target on the first run so a deep link or a Complete-summary jump link opens with it visible, then use `block: 'nearest'` so Next/Back nudge without scroll-jacking a list the user may be browsing. Rows get a `scroll-margin-top` so a scrolled-to row is never parked under the sticky pillar header. The `scrollIntoView` call must be optional-chained: jsdom does not implement it and the existing suites would throw.

3. **Stronger selected state.** A per-instance style on the row button (more specific than the theme rule, so no global regression): a primary-colored left accent bar, a real background tint, and heavier primary text, with a matching hover. The unselected rows reserve the same border width so selecting causes no layout shift, and the keyboard focus ring is preserved.

[...]

PR #671, opened 2026-08-04
fix(#668): say what makes a carried-forward answer count

### LINKED ISSUE #668: Questionnaire does not state what makes a carried-forward answer count
### Description

FY2026 pre-fills every question with the prior cycle's answer, and #659 made confirming an explicit act. But the page never tells the user that acting is required, so a pre-filled answer that looks complete silently fails to count toward Data Call Progress. The two questionnaire variants fail differently:

- **Non-CMS questions** (most systems) render the carried answer pre-selected with its justification already filled in, plus a "Confirm this answer is still accurate" button. Nothing on the page explains that skipping that button leaves the answer uncounted for the cycle.
- **CMS / ZTMF Insights questions** do gate on review — the justification field is blanked and Next is disabled until the user inserts the prior response or dismisses it — but the message under the field ("Review the previous response before continuing.") never names the actions that actually satisfy the gate, so users can't tell how to proceed.

Left alone this generates support requests from ISSOs whose progress did not move despite reviewing every question.

### Fix

Copy only, in parallel phrasing so both variants read as one system. **No behavior change:** the CMS pending-review gate stays exactly as it is, and Next stays enabled on unconfirmed non-CMS questions (it writes nothing, preserving #413's guarantee that browsing never stamps a viewer as an editor).

1. **CMS / Insights message** (`JustificationField/JustificationField.tsx`) — replace `"Review the previous response before continuing."` with:

   > Review the previous response and insert it, or write a new justification, to continue.

   Keeps `id="previous-response-review-message"` (the textarea's `aria-describedby` target), `role="status"`, and the "Checking the previous response…" initializing branch untouched.

2. **Non-CMS helper line** — add one caption in the review strip (`QuestionnairePage.tsx`), same sentence shape with the verb and ending matched to what that variant actually does:

   > Review the carried-forward answer and confirm it, or write a new justification, before continuing.

   Two deliberate differences from the CMS string: it says "carried-forward answer" rather than "previous response" because in this variant there is no separate prior-response card — the answer is pre-filled directly into the field; and it ends with "before continuing" rather than "to continue" because Next is not blocked here, so the line reads as guidance on what to do rather than a claim that the user is stopped.

   Rendered only while the question is carried-forward-unconfirmed, and suppressed when a prior-response card is present (that variant already carries its own explanation — two guidance lines on one question is worse than none).

3. **Accessibility** — the helper is plain text, not a second live region (the state chip's `role="status"` should announce only the state flip, not re-read a static sentence). It is wired to the Confirm button via `aria-describedby`, applied conditionally so the reference never dangles when the helper is absent.

### Acceptance Criteria

- [ ] On a carried-forward question without a prior-response card, the page states in plain language that the answer needs to be confirmed or updated before moving on. Advisory rather than absolute: Next is not blocked on this variant, so the copy must not imply a hard gate
- [ ] That guidance disappears once the answer is confirmed or edited, together with the Confirm button
- [ ] It does not appear on questions that already show a prior-response review card, nor on any question that is not carried-forward-unconfirmed
- [ ] The CMS / Insights pending-review message names both accepted actions (insert the prior response, or write a new justification)
- [ ] A screen reader reaching the Confirm button hears why it exists; no dangling `aria-describedby` when the helper is not rendered; the static sentence is not announced twice
- [ ] No behavior change: Next remains enabled and write-free on an unconfirmed non-CMS question, and the CMS gate still disables Next until Insert or Dismiss
- [ ] The existing test pin on the old message string is updated so its negative assertion cannot pass vacuously; new tests cover the helper's presence, absence, and accessible description
- [ ] Full suite, lint, and typecheck clean

### Affected Component

`src/views/QuestionnairePage/JustificationField/JustificationField.tsx`, `src/views/QuestionnairePage/QuestionnairePage.tsx`

### Severity

Medium - Confusing UX that generates support load; the underlying data behavior is correct.

costacalvin:
Closes #668

## Problem

A pre-filled answer looks complete, so nothing on the page told users that leaving it untouched keeps it out of the data call's count. The two questionnaire variants failed differently, and both misled:

- **Carried-forward questions without a prior-response card** showed a state chip and a Confirm button but no instruction — no indication that skipping the button leaves the answer uncounted.
- **CMS / ZTMF Insights questions** did gate on review, but the message ("Review the previous response before continuing.") never named the actions that clear the gate. Worse, users who reasonably tried writing their own justification found Next still disabled with no explanation: typing never resolves the review, because only the card's **Insert into response** and **Dismiss** buttons call `onPriorReview` (`JustificationField.tsx:534`, `:545`).

## Changes

Copy only, positioned and styled identically in both variants — directly under the justification field, so the instruction lives in one place regardless of which flow the system gets.

**CMS / Insights** (`JustificationField.tsx`):

> Review the previous response and insert it, or dismiss it and write a new justification, to continue.

Both branches are accurate: Insert and Dismiss each unlock Next. Keeps `id="previous-response-review-message"` (the textarea's `aria-describedby` target), `role="status"`, and the "Checking the previous response…" initializing branch untouched.

**Carried-forward without a prior-response card** (`QuestionnairePage.tsx`):

> Review the carried-forward answer and confirm it, or write a new justification, before continuing.

The verb and the ending differ from the CMS string on purpose, because the behavior differs. There is no card to insert from — the answer is pre-filled into the field — so it says "confirm" rather than "insert". And Next is **not** blocked on this variant, so the line ends "before continuing": it reads as guidance on what to do next rather than a claim that the user is stopped, which is what a "to continue" ending would wrongly imply. Both variants stay one sentence in the same position and voice, so they read as one system.

Suppressed when a prior-response card is present (that card carries its own message — two guidance lines on one question is worse than none) and for read-only sessions (no action to take). Rendered as plain text rather than a second live region, since the state chip already owns the announcement, and tied to the Confirm button with `aria-describedby` so the reason is announced with the action.

## No behavior change

Next remains enabled and write-free on an unconfirmed carried-forward answer (#413's guarantee that browsing never stamps a viewer as an editor), and the CMS pending-review gate is untouched — it still disables Next until Insert or Dismiss. This PR adds no gating and changes no save path.

## Acceptance criteria

- [x] On a carried-forward question without a prior-response card, the page states in plain language that the answer needs to be confirmed or updated before moving on. Advisory rather than absolute: Next is not blocked on this variant, so the copy must not imply a hard gate (AC amended on #668 from an earlier "to count for this data call" phrasing, which read as a stronger claim than the behavior supports)
- [x] That guidance disappears once the answer is confirmed or edited, together with the Confirm button
- [x] It does not appear on questions that already show a prior-response review card, nor on any question that is not carried-forward-unconfirmed
- [x] The CMS / Insights pending-review message names both accepted actions (insert the prior response, or dismiss it and write your own)
- [x] A screen reader reaching the Confirm button hears why it exists; no dangling `aria-describedby` when the helper is not rendered; the static sentence is not announced twice
- [x] No behavior change: Next remains enabled and write-free on an unconfirmed non-CMS question, and the CMS gate still disables Next until Insert or Dismiss
- [x] The existing test pin on the old message string is updated so its negative assertion cannot pass vacuously; new tests cover the helper's presence, absence, and accessible description
- [x] Full suite, lint, and typecheck clean

## Testing

New coverage:
- `JustificationField.test.tsx` — the pending-review message names both actions (positive assertion; the pre-existing negative assertion was updated so it can no longer pass vacuously against a string that had been replaced)
- `QuestionnairePage.test.tsx` — the helper renders for an unconfirmed carried answer and the Confirm button carries it as its accessible description; the helper retires with the button after confirming; it is absent for read-only sessions; and it is absent when a prior-response card is showing (that path asserts the CMS message instead)

Results: 271 questionnaire tests, 802 full suite, lint clean, no new type errors.

Manual, against dev (~38.8k `not_started` rows in FY2026 ZTM): non-CMS systems (**SPS**, **NGSC**) show the helper under the field and it clears on Confirm; CMS systems (**ACO-OS**, **ABR**) show the new gate message, and Next stays disabled until Insert or Dismiss, then enables — including the previously confusing path where a user writes their own justification first.

## Follow-up

#669 covers the sidebar half of this feedback (hidden scrollbar, selection that never follows navigation, no sense of remaining work). Independent of this PR.

### DIFF
--- a/src/views/QuestionnairePage/JustificationField/JustificationField.test.tsx
+++ b/src/views/QuestionnairePage/JustificationField/JustificationField.test.tsx
@@ -108,6 +108,17 @@ describe('JustificationField', () => {
     ).toBeInTheDocument()
   })
 
+  it('names both ways to satisfy the pending review', () => {
+    // "Review the previous response" alone left users guessing what action
+    // clears the gate; the message must name the two that do.
+    render(<Harness />)
+    expect(
+      screen.getByText(
+        'Review the previous response and insert it, or dismiss it and write a new justification, to continue.'
+      )
+    ).toBeInTheDocument()
+  })
+
   it('copies the previous response into the submitted field after acceptance', () => {
     render(<Harness />)
     fireEvent.click(
@@ -120,7 +131,9 @@ describe('JustificationField', () => {
     ).toHaveValue(insight.last_score_notes)
     expect(screen.getByText('Added to response')).toBeInTheDocument()
     expect(
-      screen.queryByText('Review the previous response before continuing.')
+      screen.queryByText(
+        'Review the previous response and insert it, or dismiss it and write a new justification, to continue.'
+      )
     ).not.toBeInTheDocument()
   })
 
--- a/src/views/QuestionnairePage/JustificationField/JustificationField.tsx
+++ b/src/views/QuestionnairePage/JustificationField/JustificationField.tsx
@@ -610,7 +610,7 @@ export default function JustificationField({
         >
           {priorInitializing
             ? 'Checking the previous response…'
-            : 'Review the previous response before continuing.'}
+            : 'Review the previous response and insert it, or dismiss it and write a new justification, to continue.'}
         </Typography>
       )}
     </Box>
--- a/src/views/QuestionnairePage/QuestionnairePage.test.tsx
+++ b/src/views/QuestionnairePage/QuestionnairePage.test.tsx
@@ -1152,13 +1152,29 @@ describe('carried-forward confirmation', () => {
   const DEVICES_LINK =
     '/questionnaire/ssd-ex/FY2026_Q1/devices/imperial-device-management'
 
+  // The guidance sentence, asserted by exact text so a copy change has to be
+  // deliberate.
+  const HELPER_COPY =
+    'Review the carried-forward answer and confirm it, or write a new justification, before continuing.'
+
   // Serves a mutable scores list and, like the real backend, flips the
   // targeted row to done when the confirm endpoint is hit — so the refetch
   // after a confirm returns the confirmed row instead of resurrecting the
-  // original fixture.
-  function installScoreMocks(scores: Array<{ scoreid: number }>) {
+  // original fixture. Insights/OpDiv rows default to empty (the non-CMS
+  // variant); pass them to exercise the prior-response card.
+  function installScoreMocks(
+    scores: Array<{ scoreid: number }>,
+    {
+      insightRows = [] as unknown[],
+      opdivRows = [] as unknown[],
+    }: { insightRows?: unknown[]; opdivRows?: unknown[] } = {}
+  ) {
     let rows = scores
     axios.get.mockImplementation((url: string) => {
+      if (url === '/opdivs')
+        return Promise.resolve({ data: { data: opdivRows } })
+      if (url === 'insights')
+        return Promise.resolve({ data: { data: insightRows } })
       if (url.includes('/questions'))
         return Promise.resolve({ data: { data: QUESTIONS } })
       if (url.startsWith('scores'))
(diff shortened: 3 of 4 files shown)


MackOverflow (commented):
Solid, but I have a note on usability that might need to be addressed first: 

⚠️ One real gap — helper doesn't retire on edit (contradicts AC #2).
AC #2 states the guidance should disappear "once the answer is confirmed or edited, together with the Confirm button." But:

The button disappears on edit — showConfirmButton includes dirty: selectQuestionOption !== initQuestionChoice || notes !== initNotes.
The helper does not — showCarryForwardHelper depends only on currentCarryState, which is derived from currentSavedScore and stays 'unconfirmed' until a save round-trips.
So when a user starts typing their own justification (the exact "confused user" path the PR is trying to help), the Confirm button vanishes but the helper line — "…confirm it, or write a new justification, before continuing" — lingers, telling them to do the thing they're already doing. Harmless for aria (button is gone, so the description is moot), but it's stale UX and it diverges from the PR's own AC and from the "retires with the button" framing.

Suggested fix — align the helper with the button so they appear/retire together:

const showCarryForwardHelper = showConfirmButton
(showConfirmButton already implies unconfirmed && !isReadOnly via canConfirmCarryForward, and !currentPriorResponse is subsumed by the priorReviewBlocked gate — worth confirming, but functionally this collapses the divergence.) If you'd rather keep them independent, add the same dirty condition to showCarryForwardHelper.

@costacalvin if this is moot, let me know and i will re-review

costacalvin:
Thanks Mack, working on verifying my fix now. Will ping you once it is ready for review.

costacalvin:
Verified @MackOverflow — fixed in `11de480`. `showCarryForwardHelper` now derives from the button, so the sentence and the action it describes cannot drift apart:

```ts
const showCarryForwardHelper = showConfirmButton && !currentPriorResponse
```

On your "worth confirming" note: you were right that `priorReviewBlocked` does the duplication work — the card's message only shows while the review is pending, and that same state blocks the button. So the second term isn't a duplication guard; it's a deliberate choice to let the card own the explanation on insights questions. Safe because a resolved review makes Next itself land the confirm (`saveResponse`, `:782`), so the quiet strip cannot strand an answer.

Hiding on edit is also covered: the draft alert takes over in that state with *"click Next or Complete to save permanently"* (`:2006`).

Tests: the dirty case now asserts the helper retires with the button (fails against the old code), plus a guard for the insights suppression. 803 passing, lint clean, no new type errors.

MackOverflow (approved):
lgtm!

jsos3-cms:

Issue #672, opened 2026-08-04
Reconcile the required questionsanswered type with the runtime fallback that still guards its absence


voidspooks:
## Problem

#654 makes `questionsanswered` required on `ScoreProgress`, which is what lets the score-presence
proxy come out rather than sit there as dead code the type claims is reachable. Nothing enforces
that at runtime. `Home.tsx` fetches `/scores/progress` and casts with `res.data.data as
ScoreProgress[]`, no validation, and nothing between the fetch and the cell normalizes the field:
`buildDashboardMaps` never touches it. That casting is the app-wide pattern rather than anything
specific to this field, at 72 `axiosInstance` call sites across 23 files today. Meanwhile
`progressColumn.tsx` keeps the `answered == null` fallback #665 added on 2026-08-03 (99fb8513),
emitting the legacy "Not updated" wording when the field is missing. So `types.ts` says the field
is always there and the cell that consumes it still guards against it not being.

The disagreement is not symmetric, and the side #654 rewrote is the exposed one. Verified against
886d35b with a throwaway test rather than by reading: with the field absent, a past-call row
renders a bare `/41` next to a warning "Incomplete" chip, because React drops `undefined` out of
the fraction, and `progressSortValue` returns `NaN`, whose comparator return is engine-defined
sort order. That is the same class of bug @jsos3-cms fixed in #648 with the `deadlineByCallName`
sort key. The current-call branch degrades cleanly to "Not updated" because the fallback is still
there, so today only the past-call branch would misrender.

ztmf#437 has been live since 2026-07-17, so this does not reproduce against any deployed backend.
What makes it worth a ticket is that #654 deletes the past-call guard (82c84bb, from #567) while
leaving the current-call one in place, so the file ends up with a required type, one branch that
trusts it, and one that does not.

## Proposal

Two halves, both small, and they can land independently.

Validate the response instead of casting it. Every `axiosInstance` site in the app casts, so a
one-field guard in `Home.tsx` would be an island and would age badly. #619 already proposes the
fetch-hook layer, and response validation belongs in whatever that ticket picks, so this half
should be scoped there rather than duplicated here.

Then settle the contradiction in `progressColumn.tsx`. Either retire the `answered == null`
fallback so the code agrees with the required type, or keep it and give the past-call branch the
same guard plus a NaN-safe default in `progressSortValue`. I prefer retiring it: the field ships
from the backend, the type now says it is always present, and it is the same species of proxy
[...]

PR #674, opened 2026-08-04
Code freeze batch: approved frontend fixes (OpDiv scope, system edit required fields, confirm guidance)

### LINKED ISSUE #578: Cleanup: retire the hasScore proxy from ProgressCell once ztmf#437 is deployed everywhere

### LINKED ISSUE #643: Assign OpDivs modal uses a stale caller scope: session-old grants can silently revoke a target's OpDiv on save

### LINKED ISSUE #667: System edit form requires fields the backend does not, blocking edits on every non-CMS system

### LINKED ISSUE #668: Questionnaire does not state what makes a carried-forward answer count

jsos3-cms:
Frontend release train for the code freeze. **The queue is complete: every approved frontend PR is in, and this is now open for review and merge.** The backend counterpart is CMS-Enterprise/ztmf#521.

Closes #643
Closes #667
Closes #668
Closes #578

## What is in it

Approved PRs only, cherry-picked in the order they cleared review so each commit's entry point is auditable. Original authorship is preserved on every commit.

| Source | Author | Change | Closes |
| --- | --- | --- | --- |
| #650 | @tarratsco | Fresh caller scope in the Assign OpDivs modal | #643 |
| #670 | @tarratsco | Required fields on the system edit form allow saving with blank values | #667 |
| #671 | @costacalvin | State what makes a carried-forward answer count | #668 |
| #654 | @voidspooks, with follow-on commits from @jsos3-cms | Retire the hasScore score-presence proxy from ProgressCell, coalesce the answered count on the past-call branch, and retire the legacy wording fallback | #578 |

The first three touch disjoint files (`OpDivGrantModal` and `UserTable`, `SystemDetailPage`, `QuestionnairePage` and `JustificationField`), so their ordering is for readability rather than to avoid conflicts. Verified that reordering produces a byte-identical tree. #654 touches `FismaTable` and `types.ts`, also disjoint from the other three, and appending it was a fast-forward.

#654 carries five commits rather than one: @voidspooks' retirement, then a docblock correction, the past-call coalescing, the wording-fallback retirement, and a fix for a comment that the retirement made self-contradicting. @danielbowne reviewed the retirement decision and asked for that last fix specifically before merge.

## Disposition of everything else

Recorded in full, because four approved PRs sitting open alongside this batch invites the question of whether something was missed. Nothing was: every approved frontend PR is in.

**The four source PRs stay open and will close without merging.** #650, #670, #671, and #654 are all folded in here. That is why their `Closes` keywords are declared on this PR instead: a PR that closes unmerged never fires its own. After this merges, confirm each of #643, #667, #668, and #578 actually closed rather than assuming, since exactly this pattern left two issues open after an earlier combined PR.

**#646** is @voidspooks' fork original that #654 rehomes. Held as a draft so it cannot merge as a duplicate. It closes once this merges, pointing at #654.

**#584** is a dependency bump with no approval, so it stays out. It joins only if it gets one.

**#672 is deliberately absent from the `Closes` list above, and that is not an omission.** Its second half is resolved here, by the coalescing and wording-retirement commits in #654. Its first half, validating the response instead of casting it, is not: that is the app-wide casting pattern at 72 `axiosInstance` call sites and it belongs with **#619**. So #672 needs a manual close naming #619 for the remaining half, rather than an auto-close that would drop it. @danielbowne asked for that pointer explicitly.

**#619** stays open and is where the validation half lands. Nothing here changes it.

**ztmf#521** is the backend counterpart batch, tracked separately. No dependency in either direction.

## Verification on the combined branch

`tsc --noEmit` clean. `eslint ./src` 0 errors, with 2 pre-existing warnings in a file none of these PRs touch. Full jest suite **815 passed, 0 failed**, 3 skipped, across 58 suites. Every commit is signed.

The point of verifying the combination rather than trusting the individual PRs is that three green PRs can still interact once stacked. They do not here, but that is a result rather than an assumption.

## How this branch is maintained

Additions are appended, never reordered. Appending keeps the push a fast-forward, avoids re-hashing other contributors' commits, and keeps the "what joined when" history intact. Each addition gets a comment recording which PR joined and the verification result at that point.

Two things worth knowing while this stays open:

Any push dismisses an approval on this PR, so approvals are best collected now that the batch is final rather than per addition.

**A skipped check is not a pass.** While this was a draft both the Snyk job and the `ui` deploy were gated on `draft == false` and reported `skipping`, which reads as green at a glance. The runs that count are the ones from after the conversion. Older `skipping` entries stay visible on the checks tab because check-runs attach to the commit, so expect duplicate names with different results.

Noted for next time: the conversion did fire a fresh run here, and on ztmf#521 as well. The concern that marking ready would not spawn a run for an already-seen head did not materialise in either repo today, so the empty-commit workaround was not needed.

## After merge

The source PRs will close without merging, so their own `Closes` keywords will not fire. That is why all four tracking issues are declared on this PR instead. Any issue not listed here needs closing by hand, which currently means #672 and #646.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -334,9 +334,8 @@ export type ScoreProgress = {
   questionsexpected: number
   // Distinct applicable questions with an answer at all, any status (ztmf#437).
   // The completion signal a closed call needs: imported/carried answers are
-  // answered but never "updated this cycle". Optional so the UI degrades to the
-  // prior score-presence proxy until the backend field is deployed.
-  questionsanswered?: number
+  // answered but never "updated this cycle".
+  questionsanswered: number
   questionsupdated: number
   lastupdatedat?: string | null
   updatedsincestart: boolean
--- a/src/views/FismaTable/FismaTable.tsx
+++ b/src/views/FismaTable/FismaTable.tsx
@@ -931,7 +931,6 @@ export default function FismaTable({
         <ProgressCell
           entry={progress?.[params.row.fismasystemid]}
           isCurrentCall={isRowCurrentCall(params.row.fismasystemid)}
-          hasScore={Boolean(scores[params.row.fismasystemid])}
         />
       ),
     },
--- a/src/views/FismaTable/progressColumn.test.tsx
+++ b/src/views/FismaTable/progressColumn.test.tsx
@@ -102,6 +102,25 @@ describe('progressSortValue', () => {
       progressSortValue(pastComplete, false)
     )
   })
+
+  it('returns a usable rank for a past-call row missing the answered count', () => {
+    // The type requires the field, so absence can only be reached at runtime
+    // against a response that omits it, and the cast is how the test expresses
+    // that. Without coalescing this returned NaN, which compares equal to every
+    // other rank and drops the row at an arbitrary position instead of ranking
+    // it. Zero is the honest rank: nothing recorded as answered.
+    const missingAnswered = {
+      ...untouchedEntry,
+      questionsanswered: undefined,
+    } as unknown as ScoreProgress
+    const rank = progressSortValue(missingAnswered, false)
+    expect(Number.isNaN(rank)).toBe(false)
+    expect(rank).toBe(0)
+    // Ranks with the genuinely unanswered rather than above a partial call.
+    expect(rank).toBeLessThan(
+      progressSortValue({ ...untouchedEntry, questionsanswered: 10 }, false)
+    )
+  })
 })
 
 describe('progressTooltip', () => {
@@ -201,6 +220,22 @@ describe('ProgressCell', () => {
     expect(screen.getByText('0/41')).toBeInTheDocument()
     expect(screen.getByText('Incomplete')).toBeInTheDocument()
     expect(screen.queryByText('Not updated')).not.toBeInTheDocument()
+    expect(screen.queryByText('Complete')).not.toBeInTheDocument()
+  })
+
+  it('renders 0/total on a past call when the answered count is missing', () => {
+    // Reachable only at runtime against a response that omits the field, which
+    // the cast expresses. The fraction interpolates the value directly and React
+    // drops undefined, so without coalescing this rendered a bare "/41" beside a
+    // warning chip. This is the one input where retiring the score-presence
+    // proxy changes what a reader sees, so it is pinned here.
+    const missingAnswered = {
+      ...untouchedEntry,
+      questionsanswered: undefined,
+    } as unknown as ScoreProgress
+    render(<ProgressCell entry={missingAnswered} isCurrentCall={false} />)
+    expect(screen.getByText('0/41')).toBeInTheDocument()
+    expect(screen.getByText('Incomplete')).toBeInTheDocument()
     expect(screen.queryByText('Complete')).not.toBeInTheDocument()
   })
 
--- a/src/views/FismaTable/progressColumn.test.tsx
+++ b/src/views/FismaTable/progressColumn.test.tsx
@@ -307,18 +307,40 @@ describe('ProgressCell', () => {
     expect(screen.queryByText('Not updated')).not.toBeInTheDocument()
   })
 
-  it('keeps the legacy Not updated wording when questionsanswered is absent', () => {
-    // The field is now required on ScoreProgress, so absence can no longer be
-    // expressed through the type. It remains reachable at runtime against a
-    // backend that predates the field, which is exactly what the cell's
-    // fallback guards, so the cast is deliberate: it exercises the runtime
-    // path the type alone can no longer describe.
+  it('reads Not started on the current call when the answered count is absent', () => {
+    // The field is required on ScoreProgress, so absence is reachable only at
+    // runtime against a response that omits it, and the cast is how the test
+    // expresses that. Both branches now coalesce, so this reads as nothing
+    // answered rather than falling back to the retired "Not updated" wording.
     const withoutAnswered = {
       ...untouchedEntry,
       questionsanswered: undefined,
     } as unknown as ScoreProgress
     render(<ProgressCell entry={withoutAnswered} isCurrentCall={true} />)
-    expect(screen.getByText('Not updated')).toBeInTheDocument()
+    expect(screen.getByText('Not started')).toBeInTheDocument()
+    expect(screen.queryByText('Not updated')).not.toBeInTheDocument()
+  })
+
+  it('no longer renders the retired Not updated wording in any state', () => {
+    // The label had one emit site, so this is the guard against it coming back
+    // through a future conditional rather than a restatement of the cases above.
+    for (const entry of [
+      updatedEntry,
+      untouchedEntry,
+      { ...untouchedEntry, questionsanswered: 41 },
+      {
+        ...untouchedEntry,
+        questionsanswered: undefined,
+      } as unknown as ScoreProgress,
+    ]) {
+      for (const isCurrentCall of [true, false]) {
+        const { unmount } = render(
+          <ProgressCell entry={entry} isCurrentCall={isCurrentCall} />
+        )
+        expect(screen.queryByText('Not updated')).not.toBeInTheDocument()
+        unmount()
+      }
+    }
   })
 
   it('describes the carried-forward state in the tooltip', () => {
--- a/src/views/FismaTable/progressColumn.test.tsx
+++ b/src/views/FismaTable/progressColumn.test.tsx
@@ -6,6 +6,7 @@ import { progressSortValue, progressTooltip } from './progressHelpers'
 const updatedEntry: ScoreProgress = {
   fismasystemid: 1,
   questionsexpected: 41,
+  questionsanswered: 12,
   questionsupdated: 12,
   lastupdatedat: '2026-07-01T15:30:00Z',
   updatedsincestart: true,
@@ -14,6 +15,7 @@ const updatedEntry: ScoreProgress = {
 const untouchedEntry: ScoreProgress = {
   fismasystemid: 2,
   questionsexpected: 41,
+  questionsanswered: 0,
   questionsupdated: 0,
   lastupdatedat: null,
   updatedsincestart: false,
@@ -52,6 +54,7 @@ describe('progressSortValue', () => {
     const empty: ScoreProgress = {
       fismasystemid: 3,
       questionsexpected: 0,
+      questionsanswered: 0,
       questionsupdated: 0,
       updatedsincestart: false,
     }
@@ -69,6 +72,7 @@ describe('progressSortValue', () => {
     const empty: ScoreProgress = {
       fismasystemid: 3,
       questionsexpected: 0,
+      questionsanswered: 0,
       questionsupdated: 0,
       updatedsincestart: false,
     }
@@ -127,6 +131,7 @@ describe('progressTooltip', () => {
     const empty: ScoreProgress = {
       fismasystemid: 3,
       questionsexpected: 0,
+      questionsanswered: 0,
       questionsupdated: 0,
       updatedsincestart: false,
     }
@@ -157,12 +162,14 @@ describe('ProgressCell', () => {
     expect(screen.getByText('Updated')).toBeInTheDocument()
   })
 
-  it('renders the fraction and a Not updated chip for a carried-over system', () => {
-    // A pre-populated questionnaire has answers but no edits this cycle -
-    // the whole point of ztmf#299 is that this renders as NOT updated.
+  it('renders the fraction and a laggard chip for an unstarted system', () => {
+    // No answers and no edits this cycle, so the cell shows the honest
+    // fraction alongside the warning chip. The fixture carries zero answers,
+    // which is the "Not started" half of the wording split; the carried-answers
+    // half is covered below.
     render(<ProgressCell entry={untouchedEntry} />)
     expect(screen.getByText('0/41')).toBeInTheDocument()
-    expect(screen.getByText('Not updated')).toBeInTheDocument()
+    expect(screen.getByText('Not started')).toBeInTheDocument()
   })
 
   it('renders an em-dash when progress data is missing', () => {
@@ -176,6 +183,7 @@ describe('ProgressCell', () => {
     const empty: ScoreProgress = {
       fismasystemid: 3,
       questionsexpected: 0,
+      questionsanswered: 0,
       questionsupdated: 0,
       updatedsincestart: false,
     }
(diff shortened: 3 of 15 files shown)


jsos3-cms:
## Conversion checklist

Collected here because the ordering matters and the loose ends are spread across three other threads.

### Ordering, and the one trap

**Mark this ready BEFORE the final push, not after.** There is already a skipped workflow run against head `6251d7c`, and marking a PR ready does not reliably spawn a fresh run for a head that has already been seen. Doing it in the wrong order leaves this blocked on the required dev deployment with no failing check to explain why. Recovery if that happens is an empty commit, which a squash merge drops.

- [ ] Mark ready (before appending anything further)
- [ ] Append #654 once a collaborator approves it. The push lands as a non-draft `synchronize`, so it fires CI on the final head, which is the head that actually needs verifying
- [ ] Add `Closes #578` to the description in the same pass, see below
- [ ] Confirm CI green: Snyk plus the `ui` deploy. Draft runs report `skipping`, so a skipped result is not a pass
- [ ] Request approvals only once CI is green
- [ ] Merge

Note that any push dismisses an approval already collected here, so approvals are best gathered once the batch is final.

Also: when this deploys to dev it takes the shared lane from whichever PR holds it. That is harmless for anything folded in here, since those close without merging, but it will re-block any other PR that was relying on its own dev deployment.

### Closes lines to carry when #654 is appended

#654 declares `Closes #578`. Because #654 will close without merging, that keyword never fires from there, so #578 has to be declared on this PR instead.

- [ ] Add `Closes #578`

The three already declared (#643, #667, #668) cover the current batch.

### Loose ends after this merges

Everything folded in here closes unmerged, so nothing auto-closes from the source PRs.

- [ ] Close #672 as subsumed. Both halves of its second proposal are in #654 (`abb236c` and `1b814fe`). Its response-validation half is out of scope and stays with #619
- [ ] Close #646 pointing at #654 as the rehome that shipped. It has been left open as a draft deliberately so it could not merge as a duplicate, on the understanding it closes once the rehome lands
- [ ] Verify each `Closes` issue actually closed after the merge, rather than assuming. The last two combined PRs left ztmf#509 and ztmf#510 open for exactly this reason, and both had shipped

### Verification standard for additions

Each addition should be verified on the combined branch rather than trusting the source PR, since separately green PRs can still interact once stacked. Current state of this branch: `tsc` clean, `eslint` 0 errors with 2 pre-existing warnings in an untouched file, full suite 813 passed and 0 failed.

jsos3-cms:
## Checklist status: queue complete, converted, CI running

Superseding the checklist above rather than editing it, so the original ordering advice stays readable.

**Done**

- [x] Marked ready. It did fire a fresh run, `30955788829`, so the empty-commit workaround was not needed. Same on ztmf#521. The trap the checklist warns about did not materialise in either repo today
- [x] Appended #654, as a fast-forward. Its five commits are in, including the fix @danielbowne asked for before merge
- [x] Added `Closes #578`. All four now register: #578, #643, #667, #668
- [x] Queue confirmed complete. Every approved frontend PR is in, verified by patch-id rather than by reading the log: #650 1/1, #670 1/1, #671 2/2, #654 5/5, totalling the nine commits on the branch
- [x] Verified on the combined branch at this head: `tsc` clean, `eslint` 0 errors with the 2 pre-existing warnings in an untouched file, full suite **815 passed, 0 failed**

**In flight**

- Snyk and Snyk code both pass on the post-conversion run. `ui / Deploy` is still going. The dev deployment for `07f1ca3a` is already the newest on the shared lane, so this PR holds it
- Ignore the older `skipping` entries on the checks tab. Those are the draft-gated runs, and check-runs attach to the commit rather than the PR, so names appear more than once with different results

**Remaining, in order**

1. Confirm `ui / Deploy` green
2. Collect one approval. Nothing is outstanding on the code
3. Merge. Squash, matching the `Title (#N)` convention on `main`

**After merge**

1. Close #672 as subsumed, and name **#619** explicitly for its response-validation half so that half does not fall between the two tickets. @danielbowne asked for that pointer specifically. This is why #672 is deliberately not in the `Closes` list
2. Close #646 pointing at #654 as the rehome that shipped. It has been held as a draft precisely so it could not merge as a duplicate
3. Verify #578, #643, #667, #668 actually closed rather than assuming. Two earlier combined PRs left ztmf#509 and ztmf#510 open for exactly this reason, and both had shipped
4. #650, #670, #671 and #654 will read closed-without-merging. That is the expected outcome, not an abandonment

Independent of this PR: ztmf#521 is the backend train, and after it merges ztmf#519 closes pointing there with ztmf#513 to verify.

jsos3-cms:
@danielbowne requesting your review on the frontend release train. The queue is complete and this is what merges; the four source PRs close without merging.

**What is in it:** #650 and #670 from @tarratsco, #671 from @costacalvin, and #654 from @voidspooks with my follow-on commits. Nine commits, every author preserved, all signed. Confirmed the queue is complete by patch-id rather than by reading the log.

**CI:** both Snyk jobs pass on the post-conversion run `30955788829`; `ui / Deploy` is still running. Ignore the older `skipping` entries, those are the draft-gated runs and check-runs attach to the commit so names repeat.

**Verified on the combined branch,** not inherited from the parts: `tsc` clean, `eslint` 0 errors, full suite 815 passed and 0 failed. Three separately green PRs can still interact once stacked, so that is a result rather than an assumption.

Your review on #654 is already incorporated. You asked for the stale cross-reference in the past-call comment to go before merge, and that is `b69eb48`, which is in this batch.

Two things I would specifically like your eye on:

`Closes #578, #643, #667, #668` are declared here because each source PR closes unmerged and would never fire its own. If any of those look wrong to own from this PR, say so before it merges rather than after.

**#672 is deliberately not in the `Closes` list.** Its second half is resolved here by #654's commits; its response-validation half is the app-wide casting pattern and stays with #619. So it needs a manual close naming #619, which is the pointer you asked for. I did not auto-close it precisely so that half does not get dropped.

danielbowne (approved):
lgtm

Issue #526, opened 2026-07-14
rename CfactsRecordCard to SystemEnrichmentCard


jsos3-cms:
### Feature Type

UI Component

### Workflow Description

The card now renders ZTMF Insights data from the /systemenrichment endpoint (ztmf-ui#367), but the component and file are still named CfactsRecordCard. Rename to SystemEnrichmentCard (file + component + import) for consistency.

Defer until the old CFACTS read surface is retired (https://github.com/CMS-Enterprise/ztmf-misc/issues/214) to avoid churn while both surfaces coexist.

Part of the ZTMF Insights decoupling (https://github.com/CMS-Enterprise/ztmf-misc/issues/190). Rehome of [ztmf-misc#220](https://github.com/CMS-Enterprise/ztmf-misc/issues/220)

### Acceptance Criteria

- [ ] Criterion 1
- [ ] Criterion 2
- [ ] Criterion 3

tarratsco:
Closed by: https://github.com/CMS-Enterprise/ztmf-ui/pull/522

PR #679, opened 2026-08-05
Let select-all reach systems with no score rows

### LINKED ISSUE #678: [P0] Select-all cannot select systems with no score rows - the checkbox gate silently drops them from selection and export
### Description

On the Home dashboard, filter to an OpDiv and click the header select-all checkbox: every system that has no score rows in the viewed data calls stays unselected, and its checkbox cannot be checked manually either. Nothing in the UI indicates rows were skipped, so anything driven by the selection - the footer's "Download selected system answers" export - silently omits exactly the systems that most need attention: the ones that haven't started.

This is the same population #655 measured on a prod clone: the 524 live systems with no score rows for the current call that the "Not updated only" filter hides are the same systems the select-all cannot select. After the HHS onboarding and FY26 inventory loads, that population grows to roughly 1,100+ until their first data call closes - in that window, an OpDiv admin doing select-all-and-export gets a file missing most of their inventory, with no warning.

A second confusion compounds it: the Zero Trust Score column renders 0.00 both for a system with a genuine zero-score aggregate row (selectable) and for a system with no row at all (unselectable). Two rows that look identical behave differently, and the user has no way to tell why.

### Root cause

`src/views/FismaTable/FismaTable.tsx` (~line 1039) gates row selection on membership in the scores map:

```tsx
isRowSelectable={(params: GridRowParams) =>
  params.row.fismasystemid in scores
}
```

The `scores` map is built in `src/views/Home/Home.tsx` (lines 17, 43) from `/scores/aggregate?datacallid=N` per active call, so a system only has a key if it has at least one score row. MUI's `checkboxSelection` header checkbox honors `isRowSelectable`, so select-all skips non-selectable rows by design, and the grid renders no disabled affordance on them.

The gate presumably exists to keep the export meaningful (a scoreless system has no answers to download), but the export endpoint already takes explicit `fsids`, and the cost of the guard is invisible exclusion.

The Score column ambiguity comes from the same file (~lines 851-856): the valueGetter returns 0 when the map has no entry, identical to a real 0.

### Fix direction

1. Preferred: gate on questionnaire membership instead of score presence, or drop the gate. `/scores/progress?datacallid=N` returns a valid row (`questionsexpected: 40, questionsupdated: 0`) for exactly these systems (verified in #655), so progress membership is the honest signal for "this system is in the call and selectable." If #655's fix flows progress rows into the dashboard maps, this gate can key off the same data - the two fixes complement each other.
2. If the gate stays: render the disabled checkbox visibly disabled with a tooltip ("No scores in the viewed data call") and surface a skip count when select-all runs ("Selected 12 of 15 visible systems").
3. Either way, stop rendering absent-entry as 0.00 in the Score column - an em dash or "Not scored" distinguishes "never assessed" from "assessed at zero."

Refs #655, #662. Prior related: #375 (closed; the export's earlier selection mismatch in the other direction).

tarratsco:
## Summary

Closes #678 (frontend half). Backend half tracked in CMS-Enterprise/ztmf#526; this PR is sequenced to merge together with it.

On the Home dashboard, select-all could not reach systems with no score rows in the viewed data calls, and their checkbox could not be ticked manually either. Nothing signaled the skip, so anything driven by the selection silently omitted exactly the not-started systems that most need attention. Compounding it, the Zero Trust Score column rendered 0.00 both for a genuine zero-score row and for a system with no row at all, so two rows that looked identical behaved differently.

Root cause: `FismaTable` gated row selection on membership in the scores map (built from `/scores/aggregate`), so a system only counted if it had a score row. A not-started system has a `/scores/progress` row (40 expected, 0 updated) but no score yet, so it was never selectable. The Score column's `valueGetter` returned 0 for a missing entry, rendering it identically to a genuine zero.

Since #655, `progressMap` already includes these not-started systems and passes them to `FismaTable` as progress, so the fix needs no new fetch.

## Changes

* **`rowSelection.ts` (new):** `isSystemSelectable(id, scores, progress)` gates selection on membership in scores OR progress, so any system present in the viewed calls is selectable. The union (not progress alone) keeps scored rows selectable if `/scores/progress` fails and resolves to an empty map.
* **`scoreColumn.tsx` / `scoreHelpers.ts` (new):** `ScoreCell` renders "Not scored" for a system with no aggregate row, distinct from a real 0.00, keeping the tier-colored badge for scored cells. `scoreSortValue` sorts a never-assessed system below a genuine zero so the two never interleave.
* **`FismaTable.tsx`:** wired in the three helpers; the inline score `valueGetter`/`renderCell` collapse to one-liners and the now-unused `TIERS` import moved into `scoreColumn.tsx`. The score `valueGetter` now returns a number (was a mixed string/number), which is correct for the `type: 'number'` column; a side effect is that quick-filter text matches the raw value ("3.5") rather than the formatted "3.50".
* **`FismaTable.tsx` (export footer):** the export call target now falls back to a row's displayed call (`chosenCallMap`, which `buildDashboardMaps` fills for every row) before the global active-call default. A not-started system has no score-derived call, so an all-never-started selection previously fell through to the active call, which is the wrong call in a past-year view. This becomes a live correctness bug the moment ztmf#526 makes those systems return export rows, so it is fixed here rather than deferred. `systemCallMap` stays score-derived (the #467 call picker and export button depend on it); only the footer's per-row fallback changed.

## Acceptance criteria

* [x] Select-all and the per-row checkbox can select not-started systems (present in progress, no score row).
* [x] A never-assessed system reads "Not scored", not 0.00; a genuine zero score system still reads 0.00.
* [x] "Not scored" sorts below a real zero rather than interleaving.
* [x] Scored rows stay selectable if the progress fetch fails.
* [x] An export of only never-started systems targets the call those rows are displayed against, not the active call (correct in a past-year view).
* [ ] Export file actually contains not-started systems: backend-owned. `FindAnswers` inner-joins on scores, so the file gains those rows only once CMS-Enterprise/ztmf#526 lands. This PR is sequenced to merge with it.

## Testing

* **`rowSelection.test.ts` (4):** scored selectable, not-started selectable (the bug fix), absent-from-both not selectable, and the failed-progress-fetch fallback.
* **`scoreColumn.test.tsx` (5):** "Not scored" vs real 0.00, and sort ordering (never-assessed below zero).
* **`FismaTable.test.tsx` (5):** footer export targeting, including the not-started displayed-call fallback and the multi-call disable guard.
* Full `FismaTable` + Home suites green (99 passed); `tsc --noEmit` and `eslint` clean.
* Behavior-only; no visual change beyond the Score column's "Not scored" badge.

### DIFF
--- a/src/views/FismaTable/FismaTable.test.tsx
+++ b/src/views/FismaTable/FismaTable.test.tsx
@@ -53,10 +53,20 @@ beforeEach(() => {
   mockAxiosGet.mockResolvedValue(blobResponse)
 })
 
-function renderFooter(selectedRows: number[]) {
+function renderFooter(
+  selectedRows: number[],
+  extra: {
+    systemCallMap?: Record<number, number[]>
+    chosenCallMap?: Record<number, number>
+  } = {}
+) {
   return render(
     <MemoryRouter>
-      <CustomFooterSaveComponent {...baseProps} selectedRows={selectedRows} />
+      <CustomFooterSaveComponent
+        {...baseProps}
+        {...extra}
+        selectedRows={selectedRows}
+      />
     </MemoryRouter>
   )
 }
@@ -96,5 +106,29 @@ describe('CustomFooterSaveComponent download button', () => {
     // Must NOT be a bare export URL (no fsids = download everything)
     expect(calledUrl).not.toMatch(/\/export$/)
     expect(calledUrl).not.toMatch(/\/export\?$/)
+  })
+
+  it('targets a not-started system displayed call, not the active call', () => {
+    // A not-started system has no score-derived call (empty systemCallMap), but
+    // the dashboard shows it against call 7 (a past-year view). The export must
+    // target 7, not the active call 42 - otherwise a past-year export silently
+    // hits the wrong call once the backend starts returning rows for it.
+    renderFooter([3], { systemCallMap: {}, chosenCallMap: { 3: 7 } })
+    fireEvent.click(
+      screen.getByRole('button', { name: /download selected system answers/i })
+    )
+    const calledUrl: string = mockAxiosGet.mock.calls[0][0]
+    expect(calledUrl).toContain('/datacalls/7/export')
+    expect(calledUrl).not.toContain('/datacalls/42/export')
+    expect(calledUrl).toContain('fsids=3')
+  })
+
+  it('disables export when selected rows display different calls', () => {
+    // Two never-started rows shown against different calls have no single
+    // export target, so the button stays disabled rather than guessing.
+    renderFooter([3, 4], { systemCallMap: {}, chosenCallMap: { 3: 7, 4: 9 } })
+    expect(
+      screen.getByRole('button', { name: /download selected system answers/i })
+    ).toBeDisabled()
   })
 })
--- a/src/views/FismaTable/FismaTable.test.tsx
+++ b/src/views/FismaTable/FismaTable.test.tsx
@@ -131,4 +131,34 @@ describe('CustomFooterSaveComponent download button', () => {
       screen.getByRole('button', { name: /download selected system answers/i })
     ).toBeDisabled()
   })
+
+  it('targets a scored system score-derived call', () => {
+    // A scored system resolves through systemCallMap (calls it has scores in),
+    // not the chosen-call fallback. Pins that the not-started fallback did not
+    // displace the original path.
+    renderFooter([1], { systemCallMap: { 1: [7] } })
+    fireEvent.click(
+      screen.getByRole('button', { name: /download selected system answers/i })
+    )
+    const calledUrl: string = mockAxiosGet.mock.calls[0][0]
+    expect(calledUrl).toContain('/datacalls/7/export')
+    expect(calledUrl).toContain('fsids=1')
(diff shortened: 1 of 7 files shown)


MackOverflow (approved):
clean, no nits, we jsut need to pay attention to merge order.

jsos3-cms:
@tarratsco shipped. Your three commits merged in #682, the frontend code-freeze batch, as `a1c0f24`, authorship and signatures intact. #678 closed via the keyword declared on the batch.

Sequencing held as you designed it: the backend half, ztmf#528, went to prod first and was confirmed live at 16:36 Pacific before this merged at 16:36:41. So the window your unchecked acceptance criterion described, where select-all is clickable but the export is still scores-anchored and returns a header-only file, never opened in prod. Thanks for flagging that explicitly in the PR rather than leaving it implicit; it is what made the ordering enforceable.

Independent review confirmed the `valueGetter` change against MUI's own documentation for a `type: 'number'` column, including that quick-filter now matching `3.5` rather than `3.50` is what the Community filter implementation produces rather than an oversight. One minor gap left unticketed: nothing tests quick-filter matching against the Score column.

Closing by hand, since the batch squash-merged and GitHub cannot match this PR's content.

PR #432, opened 2026-06-24
make idp and Opdiv fields editable

jsos3-cms:
In-repo version of #431 by @costacalvin: makes the OpDiv and Auth (identity provider) fields editable in the add-user DataGrid row, so a new user can be assigned OpDiv(s) and an explicit IdP at create time instead of requiring follow-up actions.

This PR carries that change onto a branch in this repository so the org-secret-gated CI can run — Snyk and the dev deploy (via the AWS OIDC role) do not run on a fork PR. The contributor's commit is unchanged and authorship is preserved (`Calvin Costa`); this repo is squash-merge-only, so please keep @costacalvin credited as the author on merge.

Closes #406. Refs #431.

## Summary

- **OpDiv assignment at create** — the add-user row includes a multi-select for OpDivs. Selections are sent as parallel grant requests after the user POST. The selectable set is scoped to the acting admin: OPDIV_ADMIN sees only their own OpDiv(s); HHS-wide admins (OWNER, HHS_ADMIN) see all active, non-parent OpDivs.
- **IdP selector at create** — an `okta`/`entra` dropdown shown only to HHS-wide admins (`CONFIG.IDP_ENABLED && hasUnscopedRead`). Sent in the create body only when the selector is visible and a valid value is selected.
- **isCellEditable gates** — `opdivs` and `identity_provider` are editable only on new rows; existing users' OpDiv membership and derived IdP remain managed via the Assign OpDivs action.
- **Post-create IdP display** — when OpDivs are granted, the row's `identity_provider` is refreshed from the server-recomputed value after grants settle; when none are granted, the POST response value is used directly.

See #431 for screenshots and the full "OpDiv grant derivation overrides explicit IdP" note.

## Depends on

Backend CMS-Enterprise/ztmf#352 — widens the `identity_provider` override gate from `IsOwner()` to `HasUnscopedRead()` so HHS_ADMIN can set Entra at create, and ships the `/users/{userid}/assignedopdivs` endpoints this UI calls. Confirmed green in dev at the time this branch was opened.

## Test plan

- [ ] HHS_ADMIN: add-user row shows the OpDiv multi-select and IdP selector; create a user with no OpDiv and `entra` selected → user created with `entra`
- [ ] HHS_ADMIN: create a user with an OpDiv that derives `entra` plus an explicit `okta` selection → IdP resolves to `entra` after grants (OpDiv derivation wins)
[...]

PR #681, opened 2026-08-05
test(users): pin the cell-level edit gates for the user grid

jsos3-cms:
Covers the case behind #651, which is being closed as not reproducible. Behaviour-preserving: the predicate is extracted unchanged and the call site passes the same inputs.

## Why

#651 reported that an admin could inline-edit an existing user's OpDivs cell, have the commit report success, and lose the change. That does not reproduce: `isCellEditable` has gated the `opdivs` field to new rows only since #432 merged on 2026-06-29. I filed the issue from the column definition, which declares `editable: isAdmin`, and missed the gate 180 lines below it.

The write-path asymmetry behind the report is real, though. `processRowUpdate` writes grants only on the `newRow.isNew` branch, so there is no path that persists them for an existing user, and the existing-user path still sets the saved message and optimistically updates the row. **The gate is the only thing keeping that unreachable**, and nothing tested it:

- `isCellEditable` appeared exactly once in the repo, in the implementation.
- `src/views/UserTable/` has no test file at all, so there was no harness to notice a change to it.

So a refactor that reorganised the grid props could have removed the gate and restored a write that lies about having saved, with the entire suite still green. That is the gap this closes.

## What changed

`cellEditGuards.ts` holds the predicate, extracted with no behaviour change, following the pattern this repo already uses for `saveGuard.ts`, `confirmState.ts`, and `rowSelection.ts`: a pure function with its own test rather than logic embedded in a component. The grid now calls it with the same `assignableRoles` and `showIdpSelector` it computed before.

`cellEditGuards.test.ts` covers all four branches, 11 tests: `opdivs` and `identity_provider` locked on existing rows, `role` locked when the current role is outside the caller's assignable tier, and everything else deferred to the column definition. It includes the `isNew` absent case, not just `isNew: false`, since the field is optional.

One small type improvement falls out: the `role` branch previously returned `params.row.isNew || ...`, so its value was truthy rather than boolean. The helper returns a real boolean. Same behaviour through MUI, which reads it for truthiness.

## Verification

The two `opdivs` assertions were checked against a deliberately broken guard: flipping the branch to `return true` fails exactly those two and passes the other nine, and restoring it passes all 11. Verified by running it rather than assumed, since a test that survives the bug it names is worthless.

`tsc --noEmit` clean, `eslint ./src` 0 errors with the 2 pre-existing warnings in an untouched file, full suite **826 passed, 0 failed** (up 11).

## Not addressed here

Whether the inline OpDivs editor should eventually persist rather than be gated off is a product question, not this. The gate reflects a deliberate decision that memberships are managed through the Assign OpDivs action, which is the only path with the caller-scope narrowing and preserve boundary from #642 and #650. This just makes that decision enforceable.

### DIFF
--- a/src/views/UserTable/UserTable.tsx
+++ b/src/views/UserTable/UserTable.tsx
@@ -42,6 +42,7 @@ import { fetchOpDivs } from '@/utils/opdivs'
 import { fetchUserOpDivs, setUserOpDivs } from '@/utils/userOpdivs'
 import CONFIG from '@/utils/config'
 import EditOpDivCell from './EditOpDivCell'
+import { isUserCellEditable } from './cellEditGuards'
 import { parseApiError } from '@/utils/apiErrors'
 import { isAuthHandled, notify } from '@/utils/notify'
 import { useContextProp } from '../Title/Context'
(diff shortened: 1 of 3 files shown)


MackOverflow (approved):
Minor nits, non-blocking: 
Suggestions / minor observations
Call-site wiring is untested. The tests cover isUserCellEditable in isolation, but nothing asserts that UserTable's isCellEditable actually forwards field/row/ctx correctly. It's trivial glue, so this is acceptable — worth a one-line mention only if you want the pin to survive a bad refactor of the call site too. Not blocking.
EditableRow is a hand-rolled shape rather than a slice of the users type. This is a reasonable decoupling (the guard reads only two fields), but it can drift from the real row shape without a compile error. Fine as-is; just a note.
No performance concern (pure sync predicate, per-cell as before). No security concern — comments correctly frame this as defense-in-depth with server enforcement as the real boundary.

jsos3-cms:
Shipped in #682, the frontend code-freeze batch, merged as `a1c0f24`.

This is the coverage that made #651 safe to close as not reproducible: the `isCellEditable` gate for the `opdivs` field is now pinned by 11 tests, verified to fail against a guard that admits existing rows rather than assumed to.

Closing by hand, since the batch squash-merged.

PR #529, opened 2026-07-14
feat(questionnaire): ZTMF-insights justification field (#527) [impl]

jsos3-cms:
Rehome of #527 (by @costacalvin) onto the `impl` branch so it can run Analysis CI and deploy to the impl environment — fork PRs can't reach the org secrets those workflows need.

**Base is `impl`, not `main`, on purpose.** The impl line and main have diverged; this change is being taken to impl first, and the impl↔main divergence will be reconciled separately. #527's two commits are replayed onto the current `impl` head (authorship preserved), so the diff here is exactly those changes relative to impl — not a merge of main into impl.

Refs #527.

## Summary

Adds the ZTMF-insights justification field to the questionnaire: a new `JustificationField` component (component, context, and tests), plus error handling and HHS-specific conditionals in `QuestionnairePage`.

## Files

- `src/views/QuestionnairePage/JustificationField/` — new `JustificationField` component, `justificationContext`, and unit tests
- `src/views/QuestionnairePage/QuestionnairePage.tsx` — integrate the field, error handling, HHS conditionals
- `src/views/QuestionnairePage/QuestionnairePage.test.tsx` — page tests

## Deploy path

A PR into `impl` runs Analysis (lint / security / unit tests) only. Squash-merging into `impl` deploys the bundle to the impl environment. This is rehomed from a work-in-progress fork PR to exercise it in impl ahead of the main-line reconciliation.

PR #682, opened 2026-08-05
Code freeze batch 2 (frontend): select-all for unscored systems, isso_name contract, user-grid edit-gate coverage

### LINKED ISSUE #678: [P0] Select-all cannot select systems with no score rows - the checkbox gate silently drops them from selection and export

jsos3-cms:
Frontend release train for the code freeze. Approved PRs only, cherry-picked so each contributor's authorship and signature survive. Backend counterpart is CMS-Enterprise/ztmf#529, **which must merge first** — see below.

Closes #678

## What is in it

| Source | Author | Change | Closes |
| --- | --- | --- | --- |
| #679 | @tarratsco | Let select-all reach systems with no score rows; "Not scored" distinct from a real 0.00; export targets a row's displayed call | #678 |
| #680 | @jsos3-cms | Record the `isso_name` read/write asymmetry on the field | — |
| #681 | @jsos3-cms | Pin the cell-level edit gates for the user grid | — |

All three touch disjoint areas: `FismaTable`, `types.ts`, and `UserTable`. Ordering is by approval and carries no conflict risk.

#680 and #681 declare no closing issue because both record decisions on issues already closed: #680 captures the Option B outcome from CMS-Enterprise/ztmf-misc#262, and #681 adds the coverage that made #651 safe to close as not reproducible.

## Verification on the combined branch

`tsc --noEmit` clean. `eslint ./src` 0 errors, with the 2 pre-existing warnings in a file none of these touch. Full jest suite **839 passed, 0 failed**, 3 skipped across 61 suites.

Verified as a combination rather than trusting the parts. That matters more than usual here: #679 rewrites the Score column's `valueGetter` and #681 extracts a predicate out of a different grid, so a shared-fixture or shared-import interaction would only show up once stacked. It does not, but that is a result.

## Merge order, and this one is a correctness constraint

**ztmf#529 merges first, and this follows at least five minutes later.**

Not just deploy courtesy. #679 makes a never-started system's checkbox selectable and the export button clickable for such a selection. Until the backend's export change is live, that download silently produces a header-row-only file. The author disclosed exactly this in an unchecked acceptance criterion on #679, so it is a known and accepted window rather than a hidden one, but it is why the order is not interchangeable.

Landing them the other way round is not unsafe in the crash sense; it just turns "this cannot be done yet" into "this appeared to work and gave you nothing", which is the worse of the two.

## Deploy note

Opened ready rather than draft-then-ready on purpose. That transition fired two dev deploys two minutes apart on the backend side earlier today, and the second failed on an ECS stabilise timeout after racing the first. The frontend lane is a shared bucket rather than ECS, where the same race yields a served `index.html` referencing a bundle the other deploy already deleted, so a single run matters here too. If the deploy fails, let it settle and re-run **one**; do not push again in parallel.

## What is not in it

- #584 is a dependency bump with no approval.
- #676 is the nightly `main -> impl` sync, on a different base and not feature work.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -153,6 +153,18 @@ export type FismaSystemType = {
   // Extended metadata fields, typed and canonicalized by the backend.
   // Enum strings carry a canonical value or null; the tri-state booleans are
   // true / false / null = Unknown; cloud_service_model is a decomposed list.
+  //
+  // isso_name is asymmetric between read and write, and that asymmetry is the
+  // contract. A read may return either the stored override or a name the backend
+  // resolved from the ISSO's user record, and the payload does not say which. A
+  // write always stores. So echoing back a value the user did not edit converts a
+  // resolved name into a permanent stored override for that system, silently, and
+  // no server-side rule rejects it.
+  //
+  // The guarantee therefore lives in this client, by decision rather than by
+  // accident: send only the fields the user changed. buildExtendedDiff does that,
+  // and it needs a real baseline to work. Any move toward sending the whole
+  // object reintroduces the bug with every existing test still green.
   isso_name?: string | null
   hva?: boolean | null
   fips?: string | null
--- a/src/views/FismaTable/FismaTable.test.tsx
+++ b/src/views/FismaTable/FismaTable.test.tsx
@@ -53,10 +53,20 @@ beforeEach(() => {
   mockAxiosGet.mockResolvedValue(blobResponse)
 })
 
-function renderFooter(selectedRows: number[]) {
+function renderFooter(
+  selectedRows: number[],
+  extra: {
+    systemCallMap?: Record<number, number[]>
+    chosenCallMap?: Record<number, number>
+  } = {}
+) {
   return render(
     <MemoryRouter>
-      <CustomFooterSaveComponent {...baseProps} selectedRows={selectedRows} />
+      <CustomFooterSaveComponent
+        {...baseProps}
+        {...extra}
+        selectedRows={selectedRows}
+      />
     </MemoryRouter>
   )
 }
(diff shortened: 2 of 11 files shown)


danielbowne (approved):
LGTM

Issue #683, opened 2026-08-05
Stop deleting drafts on a DRAFT_VERSION mismatch (rollback destroys them)


jsos3-cms:
## Problem

`draftStore.ts:238` deletes a stored draft when its `DRAFT_VERSION` does not match, rather than simply declining to use it. That makes the loss permanent, and it is not necessary: the 14-day TTL already cleans up genuine orphans.

Two consequences follow, and the second is the one least likely to be anticipated:

- Bumping `DRAFT_VERSION` destroys every in-progress questionnaire draft on the next page load.
- The comparison is `!==`, not `<`, so **rolling back** a deploy that bumped the version makes the older code delete every draft the newer code wrote. A rollback is what you reach for when something has already gone wrong, which is the worst moment to also lose users' unsaved text.

No current bug. Nothing has bumped the version, and the deploys on 2026-08-05 were checked and do not touch this file. Filed to make the next bump safe, not to fix something broken.

Worth knowing why it matters at all: drafts are the only thing protecting free-text justifications across a deploy. The app has no route-level code splitting, so an already-loaded tab keeps working while static assets are replaced, and the draft save is debounced a second after the user stops typing. A user typing through a deploy is fine today. A version bump is the single change that would take that away, and it would leave no other symptom.

## Fix

Return null on a version mismatch without calling `clearDraft`. Two lines. The entry becomes unusable but recoverable, so rolling forward again restores it, and the TTL still reaps it if nobody does.

Deliberately not in scope: a migrator table, a guard test that fails an unmigrated bump, surfacing an eviction to the user, and a stronger comment on the constant. All defensible, none of them earning their weight against an event that may never happen. If a future change genuinely needs to alter the stored shape, revisit them then.

## Acceptance criteria

- [ ] A version mismatch no longer deletes the stored entry
- [ ] A draft written by a newer version survives a rollback to an older one and is usable again after rolling forward
- [ ] Matching-version behaviour unchanged, and the TTL and shape guards still evict as they do now

## Research question

Whether `DB_VERSION` on the `ztmf-draft-keys` IndexedDB store carries the same hazard. If a key-store version change discards the per-user CryptoKey, existing drafts become undecryptable, which is the same outcome by a different route and is not addressed by the fix above.

PR #685, opened 2026-08-06
fix(#605): persist the dev banner dismissal for signed-in users

### LINKED ISSUE #605: Banner state persistence and update
### Description

Can we persist banner preference? When dismissing it, it comes back if you open a new tab, or navigate away from the current page.

Also, the dev banner says testing stops July 17th. Should we update the banner to remove this?

### Steps to Reproduce

1. As unauthenticated user, select the "x" to dismiss the banner
2. Login via Okta/IDM
3. Yellow banner comes back.
4. Navigate to the site in a new tab, banner comes back.

### Expected Behavior

Banner preference persists for login, regardless of time on page or page.
Remove notes about past-date deadline

### Actual Behavior

Banner stays gone when I dismiss it as a logged in user

### Environment

Dev

### Acceptance Criteria

- [ ] Banner stays gone when I dismiss it as a logged in user
- [ ] No mention of prior date (July 17th)

### Severity

None

### Affected Component

_No response_

### Screenshots

_No response_

### Logs / Error Messages

```shell

```

### Workaround

_No response_

costacalvin:
Closes #605.

## Problem

`DevEnvironmentBanner` tracked its dismissed state in `useState(false)`, so the dismissal never outlived the document. Two things reset it:

- **A new tab.** Fresh document, fresh state. This is the behavior testers actually hit.
- **Signing in.** Not an SPA transition at all — `LoginPage` navigates via `window.location.href = '/login'` (an ALB rule, not a router route), so login is a full page load that wipes all in-memory state.

## Change

Record the dismissal in `localStorage`, as `{ v, t }` — a stamp of the banner copy plus a timestamp. A dismissal holds while **both** the copy is unchanged **and** the record is under a week old.

**Why the copy stamp:** replacing `DEV_BANNER_MESSAGE` re-surfaces the notice once for everyone who dismissed the previous one. Without it, a tester who dismissed stale copy would never see it corrected.

**Why the TTL:** the copy stamp only fires when the wording changes, and impl never sets `DEV_BANNER_MESSAGE` — its stamp is permanently `default`, so no redeploy and no copy change could ever bring the banner back there. A dismissal in impl would last until someone cleared site data. Dev has a smaller version of the same problem: `default` is reused whenever the override secret is cleared after a testing window, so a months-old dismissal can silently match again. The TTL bounds every environment regardless of copy.

A record that is unparseable, malformed, or expired is left in place rather than cleared, so a bad read cannot destroy state — the same hazard #683 is open for on draft storage.

The stamp is a non-cryptographic FNV-1a hash, documented as such. Hashing rather than storing the copy verbatim also keeps the value of a build-time secret off disk, even though the same text is rendered on screen.

Storage access is guarded — private browsing and disabled site data make `localStorage` throw, and the banner stays renderable and dismissable without it.

**Files:**
- `src/components/DevEnvironmentBanner/bannerDismissal.ts` (new) — `bannerVersion`, `isBannerDismissed`, `setBannerDismissed`
- `src/components/DevEnvironmentBanner/DevEnvironmentBanner.tsx` — seed state from storage when authenticated, re-read when the session starts, persist on close
- Tests for both

## Acceptance criteria

- [x] **Banner stays gone when I dismiss it as a logged in user** — holds across reloads and new tabs. **Note the deliberate narrowing:** "stays gone" is now "stays gone for a week, or until the copy changes." Unbounded dismissal was itself a defect — see the impl case above — so this is the intended reading rather than a shortfall. Flagging it because it is a change to the AC as literally written.
- [x] **No mention of prior date (July 17th)** — no longer live. That copy was never in the repo; it is injected at build time from the `DEV_BANNER_MESSAGE` repo secret, wired at `.github/workflows/ui.yml:69` and set only for the `dev` environment, and that testing window has passed. Nothing in the codebase carries the date. Recording it here so the indirection is documented for the next time the copy needs to change.

## Scope decisions

**Signed-out dismissals are in-memory only.** The issue's repro dismisses pre-login and expects it to persist through login, but that would let a tester who dismissed the generic pre-login marker silently lose the testing copy and its feedback link. Signed-in testers get one showing.

**Not scoped per user.** A different sign-in on the same browser inherits the dismissal. Nonprod access is individual accounts on individual laptops, so there is no shared-browser case; what remains is one person with two accounts in one profile, who already knows the environment. Revisit if nonprod ever gets shared or kiosk-style access.

**No cross-tab live sync.** A new tab respects the dismissal via a mount-time read, which is all the AC needs. Piggybacking on `sessionSync.ts` would require refactoring its `channel.onmessage` assignment into a dispatch, since a second consumer would clobber the logout listener.

## Risk

None to the data call. The entire component is gated on `CONFIG.IS_NONPROD` (`src/utils/config.ts:19`, derived from Vite `MODE`), so this code cannot run in production. No API, no schema, no data touched.

Note `build:impl` uses `--mode impl`, so the banner *does* render in impl — that is what makes the TTL worth having rather than theoretical.

## Verification

- `yarn test:ci` — 863 passed, 3 skipped, 0 failures
- `yarn lint:js` — 0 errors
- 32 tests in the banner directory. Coverage includes: dismissal survives a remount when signed in; is *not* persisted pre-login; a pre-login dismissal does not suppress the signed-in banner; a stored dismissal is picked up when the session starts mid-mount; changed copy re-shows; a record past the TTL re-shows while one inside it does not; four malformed-record shapes read as not-dismissed **and** are left in storage; a throwing `localStorage` breaks neither render nor dismissal.
- Assertions were checked by mutation rather than by eye — removing the TTL check, widening the TTL, clearing the key on a bad read, dropping the timestamp type guard, swapping `localStorage` for `sessionStorage`, and deleting the auth-flip effect each fail at least one test.

Note for reviewers: `yarn typecheck` reports 3 pre-existing `TS2786` errors in `EmailModal.tsx` and `QuestionnairePage.tsx`. They reproduce on a clean `main` and are unrelated to this change, but they do mean `yarn pre-push` is currently red.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/components/DevEnvironmentBanner/DevEnvironmentBanner.test.tsx
+++ b/src/components/DevEnvironmentBanner/DevEnvironmentBanner.test.tsx
@@ -24,6 +24,9 @@ beforeEach(() => {
   mockConfig.DEV_BANNER_MESSAGE = ''
   mockConfig.DEV_FEEDBACK_URL = ''
   mockConfig.DEV_CONTACT_EMAIL = ''
+  // setupTests does not reset storage between tests in a file, so a persisted
+  // dismissal would leak into every case that follows it.
+  localStorage.clear()
 })
 
 test('renders nothing in production', () => {
@@ -91,6 +94,70 @@ test('rejects a non-https feedback URL', () => {
 test('can be dismissed', async () => {
   const user = userEvent.setup()
   render(<DevEnvironmentBanner />)
+  expect(screen.getByText(/non-production environment/i)).toBeInTheDocument()
+  await user.click(screen.getByRole('button', { name: /close/i }))
+  expect(
+    screen.queryByText(/non-production environment/i)
+  ).not.toBeInTheDocument()
+})
+
+test('a signed-in dismissal survives a fresh page load', async () => {
+  const user = userEvent.setup()
+  const { unmount } = render(<DevEnvironmentBanner authenticated />)
+  await user.click(screen.getByRole('button', { name: /close/i }))
+  unmount()
+
+  // A remount stands in for a new tab or a reload - both start from scratch.
+  render(<DevEnvironmentBanner authenticated />)
+  expect(
+    screen.queryByText(/non-production environment/i)
+  ).not.toBeInTheDocument()
+})
+
+test('does not persist a dismissal made before sign-in', async () => {
+  const user = userEvent.setup()
+  const { unmount } = render(<DevEnvironmentBanner />)
+  await user.click(screen.getByRole('button', { name: /close/i }))
+  unmount()
+
+  render(<DevEnvironmentBanner />)
+  expect(screen.getByText(/non-production environment/i)).toBeInTheDocument()
+})
+
+test('a pre-login dismissal does not suppress the banner after sign-in', async () => {
+  const user = userEvent.setup()
+  mockConfig.DEV_BANNER_MESSAGE = 'OpDiv data loaded for testing.'
+  const { rerender } = render(<DevEnvironmentBanner />)
+  await user.click(screen.getByRole('button', { name: /close/i }))
+
+  // Signing in must still show the testing copy and its feedback link once.
+  rerender(<DevEnvironmentBanner authenticated />)
+  expect(screen.getByText('OpDiv data loaded for testing.')).toBeInTheDocument()
+})
+
+test('shows again for a signed-in user when the banner copy changes', async () => {
+  const user = userEvent.setup()
+  mockConfig.DEV_BANNER_MESSAGE = 'Testing runs through July 17th.'
+  const { unmount } = render(<DevEnvironmentBanner authenticated />)
+  await user.click(screen.getByRole('button', { name: /close/i }))
+  unmount()
+
+  mockConfig.DEV_BANNER_MESSAGE = 'Testing is ongoing.'
+  render(<DevEnvironmentBanner authenticated />)
+  expect(screen.getByText('Testing is ongoing.')).toBeInTheDocument()
+})
+
+test('renders and dismisses when storage is unavailable', async () => {
+  const user = userEvent.setup()
+  // Private browsing and disabled site data make localStorage throw.
+  jest.spyOn(Storage.prototype, 'getItem').mockImplementation(() => {
+    throw new Error('storage disabled')
+  })
+  jest.spyOn(Storage.prototype, 'setItem').mockImplementation(() => {
+    throw new Error('storage disabled')
+  })
+
+  render(<DevEnvironmentBanner authenticated />)
   expect(screen.getByText(/non-production environment/i)).toBeInTheDocument()
   await user.click(screen.getByRole('button', { name: /close/i }))
   expect(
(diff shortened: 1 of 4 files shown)


jsos3-cms (changes requested):
Nice implementation, and the scope write-up is thorough. Two ways a persisted dismissal could bite, and one mostly does not apply to us.

**Cross-user, largely negated.** The key is a single constant and `Title.tsx` passes only `authenticated`, so a different sign-in on the same browser inherits the dismissal. I did reproduce it. But the banner is gated on `IS_NONPROD` and never renders in production, and we do not have users sharing browsers, so what remains is one person with two accounts in one profile who already knows the environment. Optional. `src/views/QuestionnairePage/draftStore.ts` has the pattern if you want it: keys hashed by userid, plus a purge on mount from this same `Title.tsx`.

A dev dismissal does not leak into impl either, since the three environments are separate hostnames and localStorage is origin-scoped. That is the browser doing it rather than us, though. The environment appears nowhere in the key, so it would break silently if two environments ever shared a host.

**The one worth fixing: nothing bounds a dismissal in impl.** `bannerVersion()` returns the constant `'default'` whenever the override is empty, and only dev sets `VITE_DEV_BANNER_MESSAGE`. So impl's stamp is permanently `'default'`, no redeploy changes it, and no timestamp is stored to expire against. A dismissal there is permanent until someone clears site data.

A configurable TTL fits this better than deriving the stamp from the rendered copy, which only re-fires when the words change and would miss an impl redeploy that behaves differently under identical text. A week or so would mean working in impl all week shows the banner once, and coming back after a month shows it again. Deploy-keyed would work too, but it needs a build ID we do not expose today, and re-showing on every impl deploy would get dismissed reflexively.

Two notes if you take the TTL. Existing records carry no timestamp, so treat those as not dismissed and let the banner show rather than clearing the key, which is the trap #683 is open for. And `'default'` is reusable: when dev's override is cleared after a testing window the stamp returns to it, and testers who dismissed that months ago stay quiet, so dev has a smaller version of the same problem.

Smaller: swapping `localStorage` for `sessionStorage` leaves all twenty tests green, which would quietly break the new-tab requirement. One direct read-back assertion pins it.

Everything else checked out: storage errors fail safe and are genuinely tested, no flicker thanks to the lazy initializer, and `tsc`, `eslint` and all twenty tests pass.

*Edited: the cross-user point was originally framed as blocking, and the copy-derivation fix I first suggested is replaced by the TTL above.*

costacalvin:
Thanks @jsos3-cms, good catch on the unbounded dismissal. What changed:

TTL added. The record is now { v, t } and holds only while the copy is unchanged and it is under a week old. Seven days hardcoded.

Kept the copy stamp alongside it rather than replacing it, which is the one place I went a different way. It's what makes a wording change re-fire immediately instead of waiting out a week; the TTL covers time and the environments with no copy signal.

Malformed, expired, and pre-TTL records are left in storage, never cleared, per the #683 hazard. A bare pre-TTL stamp makes JSON.parse throw, the existing catch returns false, banner shows. There's a test, and a mutation that clears the key fails it.

sessionStorage is covered as of 3c80346, including direct read-back assertion on the key and stored shape, and swapping to sessionStorage now fails two tests.

No change for cross-user or environment-in-key, matching where we landed on both.

Now 863 tests, 32 in the banner directory, assertions checked by mutation rather than by eye. Verified locally: survives reload and new tab, backdating past seven days brings it back, expired record still in storage afterward.

Ready for another look.

jsos3-cms (approved):
LGTM!

costacalvin:
This only affects lower environments so we are good to merge

PR #687, opened 2026-08-07
Last Seen column in the user management grid

### LINKED ISSUE #675: Show last_seen in the users list — presentation TBD
### Description

The backend now serves `last_seen` on `/api/v1/users` (CMS-Enterprise/ztmf#522). Nothing displays it yet, and that is deliberate — the data is available so the team can decide how it should be shown rather than having a presentation chosen for them.

Until then the question is answerable by query, so this is not blocking anyone.

### What the field is

The timestamp of the most recent activity that user recorded in the events log, **including signing in**. Derived on read, so it is always current.

The backend now writes a session event at login, so someone who signs in and only reads still registers. That means "signed in but never touched anything" and "never signed in at all" are now distinguishable, which they were not before.

One caveat to respect in the wording: the log is not retrospective. For an account created before login events existed, `null` means "nothing since then", not "never". That ambiguity shrinks as people sign in, but the UI should not promise more than the data supports on day one.

### Why it is worth surfacing

Roughly 1,250 respondent accounts exist, nearly all created by the onboarding load rather than by signup. During the FY2026 call the recurring question has been "has this person actually shown up" — for chasing participation, for deciding whether an OpDiv is stalled or simply untrained, and for spotting accounts that were provisioned to the wrong address. Right now answering it means one of us running SQL.

It also covers stale privileged accounts, which we currently have no way to spot at all: an administrator who has not signed in for months is now visible without anyone writing a query.

### Some directions, none of them decided

- A **Last seen** column in the users table, sortable, with never-used rendered distinctly rather than as an empty cell
- A filter or quick toggle for "no recorded activity", since that is the actionable subset
- Relative rendering ("3 days ago") with the absolute timestamp on hover
- Surfacing it per OpDiv so an OpDiv admin can see their own team's engagement — this is the version that would take us out of the help-desk seat, and it needs the endpoint's OpDiv scoping to be confirmed for this field

### Constraints worth respecting

- `last_seen` is `null`, not absent, for accounts with no recorded activity; treat null as a real value rather than missing data
- Null is not proof of "never signed in" for older accounts — see the caveat above
- The users list is already OpDiv-scoped for OpDiv-tier admins, so this field inherits that scoping — worth verifying rather than assuming when the UI lands

danielbowne:
Closes #675.

Adds a sortable "Last Seen" column to the user management grid, reading the `last_seen` field the backend now serves inline on `GET /users` - no new requests, the column rides the existing list response. Verified against the local prod clone: 1,269 users, 295 with timestamps, and the new "No Activity Only" toggle matches the remaining 974.

| | |
|---|---|
| Cell render | relative time ("18 hours ago"), absolute timestamp on hover and in a visually hidden span for screen readers |
| Never-active rows | italic "No activity recorded" - deliberately not blank, and deliberately not "never logged in" |
| Sorting | nulls last in both directions |
| Stretch (from the issue) | "No Activity Only" toolbar switch |

Two wording/mechanism choices worth knowing about:

**Why "Last Seen" and a neutral empty state.** Activity tracking began 2026-08-06, so for accounts created before that, null means "nothing since tracking began", not "never signed in". The label and empty state are worded so the column never accuses a user who simply predates tracking. A test pins the wording so it can't drift into "Never logged in".

**Why the sort comparator is direction-aware.** We're on x-data-grid v6.19.4, which has no `getSortComparator` (that's v7). The grid negates a plain comparator's result when sorting desc, which would flip "nulls last" into "nulls first" the moment someone sorts by most-recent. `compareLastSeen` reads the active sort direction and pre-compensates so never-active rows always sink to the bottom. Alternatives considered: a plain comparator (wrong under desc) and upgrading to v7 (out of scope for this change). The negation behavior is verified against the installed grid source and simulated end-to-end in a unit test.

The toggle drives a controlled filter model with an `isEmpty` filter on the column. The community DataGrid only keeps one column-filter item, so turning the toggle on replaces the items array rather than appending (an appended item is silently discarded when a filter panel filter already exists); quick-filter text lives beside the items and is preserved. That transform is extracted to `withNoActivityFilter` and unit-tested, including the pre-existing-filter case.

## Rider: InsightsPanel helper extraction

`severityStyle`, `rollupControls`, their types, `asText`, and `INFERRED_PASS_SOURCES` moved verbatim from `InsightsPanel.tsx` to a new `controlRollup.ts`. This clears the two long-standing `react-refresh/only-export-components` lint warnings (a component file must only export components). No behavior change - the move was line-diffed and the only deltas are `export` keywords, an import, and a docstring. Test imports updated to the new module.

## Testing

- 38 tests across `lastSeen.test.ts` / `LastSeenCell.test.tsx` / the untouched `cellEditGuards.test.ts`: ISO parsing (including the backend's microsecond timestamps), relative formatting, the direction-aware comparator with a grid-negation simulation, null render, the hidden-span a11y text, and the filter-model transform.
- Full suite: 866 passing. Typecheck and lint clean, zero warnings repo-wide.
- Verified live against a prod clone: render, both sort directions, and the toggle, as an OWNER account. OpDiv scoping needs no frontend work - `last_seen` is computed inside the same fail-closed OpDiv-scoped query as the rows themselves.
- Deliberately not needed: no new endpoints or hooks, no changes to `/users/current` (which intentionally returns null for this field), no edit-gate changes (the column is not editable).
image

The one judgment call to scrutinize: the toolbar switch wiring is covered by unit tests on the extracted filter-model transform rather than a grid-level component test, matching how the edit gates in this table are already tested.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -371,6 +371,11 @@ export type users = {
   deleted?: boolean
   isNew?: boolean
   identity_provider?: 'okta' | 'entra'
+  // Most recent recorded activity (incl. sign-ins), list endpoint only.
+  // null = no activity recorded since tracking began (2026-08-06), which for
+  // older accounts is NOT the same as "never signed in" - keep any empty-state
+  // wording neutral. /users/current always returns null for this field.
+  last_seen?: string | null
 }
 
 export type datacall = {
--- a/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
+++ b/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.test.tsx
@@ -1,10 +1,9 @@
 import { render, screen, fireEvent } from '@testing-library/react'
 import InsightsPanel, {
   OptionInsightBadges,
-  severityStyle,
-  rollupControls,
   ISSO_DETERMINATION_DISCLAIMER,
 } from './InsightsPanel'
+import { severityStyle, rollupControls } from './controlRollup'
 import type { InsightPayload } from '@/types'
 import CONFIG from '@/utils/config'
 
--- a/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.tsx
+++ b/src/views/QuestionnairePage/InsightsPanel/InsightsPanel.tsx
@@ -14,6 +14,15 @@ import {
   asImpactLevel,
   FIPS_GOLD,
 } from '../fipsBaseline'
+import {
+  asText,
+  severityStyle,
+  rollupControls,
+  INFERRED_PASS_SOURCES,
+  type ControlRollup,
+  type ControlRollupState,
+  type ControlEvidence,
+} from './controlRollup'
 
 type Props = {
   payload: InsightPayload
@@ -1169,17 +1178,6 @@ function FeedCheckBlock({
   )
 }
 
-// Coerce an opaque payload value to renderable text. The payload is untrusted
-// JSON, so a field the type declares as a string could arrive as an object or
-// array; rendering that directly throws "Objects are not valid as a React
-// child" and (via the error boundary) blanks the whole panel. Coercing to a
-// string degrades gracefully instead. Returns undefined for nullish/empty so
-// existing truthiness guards still hide the element.
-function asText(v: unknown): string | undefined {
-  if (v == null || v === '') return undefined
-  return typeof v === 'string' ? v : String(v)
-}
-
 // Hardenize instance `detail` is inconsistent — a stringified JSON object
 // ({"Error message":"..."}), plain text ("Dangling DNS record: ..."), or empty
 // ({}). Surface something readable: JSON objects → their values joined; plain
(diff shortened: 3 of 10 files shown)


MackOverflow (dismissed):
The direction-aware sortComparator is the one thing to scrutinize (the author flags this too). Its correctness hinges on cellParams1.api.getSortModel() returning the newly-applied sort direction at the moment the comparator runs:

sortComparator: (v1, v2, cellParams1) => {
  const direction = (cellParams1.api.getSortModel() as GridSortItem[])
    .find((item) => item.field === cellParams1.field)?.sort ?? 'asc'
  return compareLastSeen(v1, v2, direction === 'desc' ? 'desc' : 'asc')
}
In v6, sortedRowsSelector runs the comparator against the already-updated state.sorting.sortModel, and getSortModel() reads that same state — so this is sound in v6.19.4, and api is present on GridSortCellParams in v6. The logic in compareLastSeen is a consistent total order under both asc and post-negation desc (I traced all four cases). ✅
But the wiring itself is untested. The 38 tests exercise compareLastSeen in isolation and simulate grid negation with -compareLastSeen(...) — they never assert that the column's comparator actually reads the live direction from getSortModel(). That glue is exactly the fragile part. Consider one integration-level test (render the grid, click the header twice, assert null rows stay at the bottom) — or at minimum acknowledge this is manual-verification-only.
Upgrade landmine: this whole approach exists to work around v6 lacking getSortComparator. It will silently invert on a v7 upgrade. The comment documents this well; ideally leave a // TODO(x-data-grid v7): replace with getSortComparator breadcrumb at the call site so the migration doesn't miss it.

danielbowne:
Good catch on the glue being the untested part. Pulled the wiring out of the column def into lastSeenSortComparator and added a grid-level test that renders a real DataGrid, clicks the header twice, and asserts the null rows stay at the bottom both ways, so the live getSortModel() read is exercised for real now instead of simulated with a hand-negated comparator. Also left the TODO(x-data-grid v7) breadcrumb at the definition so the getSortComparator migration doesn't get missed. be9dde0

danielbowne:
Stop finding good nits in my code Mack ;) Both fixed in be9dde0: the comparator wiring now has a real grid test (header clicked both ways, nulls stay at the bottom) and the v7 TODO breadcrumb is in place.

MackOverflow (approved):
looks good, no nits!

jsos3-cms:
Landed in `main` as part of #691, squash commit `b830d21`. Your four commits went in unchanged and `Closes #675` is re-declared on the batch, so ztmf-ui#675 is closed by the merge.

Closing as superseded rather than abandoned. Thanks @danielbowne.

PR #688, opened 2026-08-07
Scope the Scored / Systems tile to the selected call(s)

### LINKED ISSUE #658: Dashboard "Scored / Total Systems" tile mixes a call-scoped numerator with an all-systems denominator
## Summary

The first dashboard stat tile, "Scored / Total Systems", puts a data-call-scoped numerator over an all-systems denominator, so the ratio does not mean what it reads as. A large share of the apparent shortfall is systems that are not part of the selected data call at all, not systems that are behind on it.

## Detail

In `src/views/StatisticBlocks/StatisticsBlocks.tsx`:

* The numerator (`answeredSystems`) counts systems that have an entry in the `scores` map. That map is `scoreMap` from `buildDashboardMaps`, built from `/scores/aggregate?datacallid=...` for each call currently selected in the year picker, so it is scoped to the selection.
* The denominator (`totalSystems`) is `fismaSystems.length`. `fismaSystems` comes from `/fismasystems` in `Title.tsx`, which takes no `datacallid`. It is every system the user can see, across all calls and all time.

The two never line up. A system that has never been part of the selected call still lands in the denominator and can never reach the numerator, so the tile permanently reports it as unscored for a call it was never in.

The comment above the loop currently claims the opposite:

```
// Counted inside the loop so the numerator and denominator always cover the same systems.
```

That holds for the average (`avgSystemScore` divides by `scoredCount`, which is what ztmf-ui#633 fixed), but not for this tile, which sets `totalSystems` from `fismaSystems.length` outside that reasoning.

Scale of the gap is easy to see against the current open call: the not-started population for that call is in the dozens, while the tile's implied unscored count is in the hundreds.

## Why it matters

The tile is the first number on the dashboard and the one people quote for data-call completeness. Reading it as "this many systems still owe us answers" overstates the outstanding work by an order of magnitude, and it drifts silently, since there is no error state or visual cue when the number is wrong.

## Proposed fix

Scope the denominator to the systems that are actually in the selected call(s): systems with a progress row where `questionsexpected > 0`. `progressMap` from `buildDashboardMaps` is already exactly that set and is already passed down the same path.

The label needs to change with it, since "Total Systems" will no longer be the whole inventory. Something like "Scored / Systems in this data call" states the population it is actually measuring.

## Ordering

This should land after #657. Before that fix, `progressMap` was missing never-started systems entirely (they had a progress row but no score row, and map construction keyed off scores alone), so `progressMap` could not serve as a correct call-scoped denominator. #657 is what makes the proposed fix possible.

## Secondary, needs confirming first

The numerator guard is `if (entry && entry.score)`, a truthiness check, so a system scoring exactly 0.00 would be dropped from the count, the average, and the "Lowest System Score" tile. Whether that is reachable depends on whether ZTMF scoring can produce a genuine 0.00. Worth confirming before changing it, because `buildDashboardMaps` also collapses a null `systemscore` to 0 via `systemscore ?? 0`, and excluding *that* case is correct. If 0.00 is reachable, the guard needs to distinguish the two rather than just switching to a null check.

## Acceptance criteria

* [ ] The tile's numerator and denominator cover the same population: systems in the selected data call(s).
* [ ] A system not present in any selected call does not appear in the denominator.
* [ ] The tile label names the population it measures.
* [ ] Selecting a different year or call updates both halves consistently.
* [ ] Unit tests in `StatisticsBlocks.test.tsx` cover a system outside the selected call and a call-scoped denominator.
* [ ] The stale "numerator and denominator always cover the same systems" comment is corrected or made true.

tarratsco:
## Summary

Closes #658.

The dashboard's first stat tile, "Scored / Total Systems", put a call-scoped numerator over an all-systems denominator, so the ratio did not mean what it read as. The numerator (`answeredSystems`) counts systems with an entry in `scoreMap`, which `buildDashboardMaps` builds per selected call. The denominator was `fismaSystems.length` from `Title.tsx`, which takes no `datacallid` and represents every system the user can see across all calls and all time. A system that was never part of the selected call still landed in the denominator and could never reach the numerator, so the tile permanently reported it as unscored for a call it was never in. This overstated the outstanding work by roughly an order of magnitude in production, with no error state when it drifted.

## Changes

* **`StatisticsBlocks.tsx`:** added a `progress` prop and scoped the denominator to systems actually in the selected call(s): a progress row with `questionsexpected > 0`, the set that already includes never-started systems. Numerator, average, and high/low logic are unchanged. The stale "numerator and denominator always cover the same systems" comment is corrected. Label changed to "Scored / Systems in selected data calls", which reads correctly whether one call or a whole year is selected.
* **`Home.tsx`:** passes `progress={progressMap}` (already built there for the table).
* Documented why the numerator's truthy score guard is correct: backend system scores are floored at 1.0 (`AVG(answer + 1.0)` per pillar), so a falsey value only comes from an absent/null score coalesced upstream, which must not count as scored.

## Acceptance criteria

* [x] The tile's numerator and denominator cover the same population: systems in the selected data call(s).
* [x] A system not present in any selected call does not appear in the denominator.
* [x] The tile label names the population it measures.
* [x] Selecting a different year or call updates both halves consistently (both derive from `buildDashboardMaps`).
* [x] Unit tests cover a system outside the selected call and a call-scoped denominator.
* [x] The stale "same systems" comment is corrected.

## Testing

* **`StatisticsBlocks.test.tsx` (7 tests):** denominator scoped to system(s) in the selected call(s) with an out-of-call system excluded; a `questionsexpected: 0` (no-questionnaire) system excluded; a no-participation call reads 0 / 2 (not below scale); a total progress-fetch failure reads N / 0 rather than a false 100%; thousands separators intact.
* `tsc --noEmit` and `eslint` clean; Home suite (12) green.
* Behavior-only; the visible change is the corrected count and label.

### DIFF
--- a/src/views/Home/Home.tsx
+++ b/src/views/Home/Home.tsx
@@ -103,7 +103,7 @@ export default function HomePageContainer() {
   }
   return (
     <Box>
-      <StatisticsBlocks scores={scoreMap} />
+      <StatisticsBlocks scores={scoreMap} progress={progressMap} />
       <BreadCrumbs />
       <FismaTable
         scores={scoreMap}
--- a/src/views/StatisticBlocks/StatisticsBlocks.test.tsx
+++ b/src/views/StatisticBlocks/StatisticsBlocks.test.tsx
@@ -1,6 +1,6 @@
 import { render, screen } from '@testing-library/react'
 import StatisticsBlocks from './StatisticsBlocks'
-import type { FismaSystemType, SystemScoreEntry } from '@/types'
+import type { FismaSystemType, ScoreProgress, SystemScoreEntry } from '@/types'
 
 // StatisticsBlocks reads the full active-system list from the outlet context and
 // the selection-scoped score map from props. The bug in ztmf-ui#633 was the
@@ -22,6 +22,18 @@ const score = (
   score: n,
   tier,
 })
+
+// A progress row for a system in the selected call(s). `expected` is the number
+// of applicable questions; > 0 means the system is part of the call.
+const prog = (id: number, expected: number): ScoreProgress =>
+  ({
+    fismasystemid: id,
+    questionsexpected: expected,
+    questionsanswered: 0,
+    questionsupdated: 0,
+    lastupdatedat: null,
+    updatedsincestart: false,
+  }) as ScoreProgress
 
 // Read the big numeral out of a tile located by its label. The high/low tiles
 // fold an acronym into the label element, so callers pass a regex for those.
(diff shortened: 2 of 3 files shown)


jsos3-cms (approved):
Approving. I verified the claims by execution rather than inspection, and every concern I went in with turned out to be unfounded.

The effect's dependency array does include `progress` (`StatisticsBlocks.tsx:110`), so the denominator recomputes when the progress map arrives and when the selected call changes. There is exactly one call site for this component, `Home.tsx:106`, and it passes the prop, so the optional signature is not a live hazard today. And there is no division to go wrong: the tile concatenates two `toLocaleString()` numerals, so an `inCallCount` of zero renders as literal text rather than `NaN`, and the average divides by the scored count behind its own zero guard, untouched by this change.

Your truthy-score reasoning also checks out against the backend. System and pillar scores both derive from `AVG(COALESCE(a.score, 0) + 1.0)`, and `functionoptions.score` is fixed reference data in the 1 to 4 range with no write path, so a computed score is never below 1.0 and a missing one is absent from the response rather than present as zero. That closes the open question #658 raised before this PR existed.

Three non-blocking notes.

The strongest is a test gap rather than a defect, and it is worth two minutes while this is fresh. I mutated the dependency array back to `[fismaSystems, scores]` and all seven tests still passed, because each test renders once with its final props and React runs an effect on mount regardless of what its dependency array says. So the array is correct today and nothing pins it: a later refactor could drop `progress` and the suite would stay green, reintroducing exactly the stale-denominator bug this PR fixes. One test that renders with a stale or undefined `progress`, then rerenders with a populated map and asserts the tile updates, closes that permanently.

Second, "an outage reads as 0" holds for a total progress failure but not a partial one. `aggregateScores.ts` populates the score and progress maps independently, and `Home.tsx` catches per call id, so if one call's progress request fails while its score request succeeds, systems scored via that call land in the numerator and are absent from the denominator. The tile can then read something like 3 / 1. Degraded-backend only and cosmetic, so I would not hold anything for it.

Third, `questionsexpected === 0` is not exclusively the intentional no-questionnaire case. The backend's `expected` CTE is built from inner joins through `datacenterenvironments`, so a system whose environment value fails to resolve also lands at 0. Such a system is absent from the scores map too, so it drops out of numerator and denominator together and the ratio stays honest, but the true call population is undercounted by one. Pre-existing and not worsened here; a backend thread if anyone wants it.

Folding this into tonight's frontend batch.

jsos3-cms:
Landed in `main` as part of #691, squash commit `b830d21`. Your commit went in unchanged and `Closes #658` is re-declared on the batch, so ztmf-ui#658 is closed by the merge.

The one follow-up worth carrying forward is the test-hardening note in my review above: the effect's dependency array is correct today but no test pins it, since a single `render()` cannot exercise a dependency array at all. A `rerender()` test would close that permanently. Not urgent, and easiest while this is still fresh.

Closing as superseded rather than abandoned. Thanks @tarratsco.

Issue #546, opened 2026-07-15
Dashboard: use backend questionsanswered for past-call progress, retire the "has a score" proxy


voidspooks:
## Context

ztmf-ui#537 fixed the "past data call shows `0/40 Not updated`" bug on the Home dashboard. Because the backend had no real completion field at the time, that fix (PR #542) leans on a proxy: for a past call it shows a binary neutral **Complete** chip whenever the system **has a score** for that call.

The backend follow-up (ztmf#435 / PR ztmf#437) adds a real field to `GET /scores/progress`:

- **`questionsanswered`** — distinct applicable functions that have an answer row in that data call, regardless of whether they were touched this cycle. This is the answered/total a *closed* call needs. Carried-forward answers count here.
- `questionsupdated` is now derived from persisted `scores.status` (genuinely saved this cycle) rather than the events audit log; it stays the active-call numerator.

Once ztmf#437 ships, the frontend can retire the "score present ⇒ complete" proxy and render a real completion count for past calls.

## What to do

1. Add `questionsanswered: number` to the `ScoreProgress` type (`src/types.ts`).
2. In the past-call branch of `ProgressCell` (`src/views/FismaTable/progressColumn.tsx`), replace the `hasScore` proxy with the real field — render `questionsanswered / questionsexpected` for a past call instead of the binary Complete chip.
3. Keep the `isCurrentCall` gating from #542 as-is. Active calls still use `questionsupdated` (Updated / Not updated); past calls use `questionsanswered`. The framing decision stays; it just drives a real number now.
4. Update tests (`progressColumn.test.tsx`) and drop the now-unused `hasScore` plumbing if nothing else needs it.

## Open question (decide before implementing)

Do we want the richer fraction on a past call (`38/40`, distinguishes fully- from partially-answered historical calls), or keep the binary **Complete** chip (now backed by `questionsanswered > 0` / `=== questionsexpected` instead of the score proxy)? Mack/Daniel chose binary originally only because there was no field to do better — now there is. This decides how much of the past-call branch changes.

## Dependencies

- Blocked on **ztmf#437** being merged **and deployed** — the API must actually return `questionsanswered` before this can be wired.

## Acceptance criteria

- Past-call progress reflects `questionsanswered` from the API, not the "has a score" proxy.
- Active-call chip is unchanged (still `questionsupdated`, Updated / Not updated).
- The `0/0` N/A case and the Zero Trust Score column are unchanged.
[...]

PR #690, opened 2026-08-10
fix(#289): restore SaaS pillars on historical data calls

costacalvin:
Part of CMS-Enterprise/ztmf-misc#289 — phase 1 of 2. Pairs with CMS-Enterprise/ztmf#546, **which must deploy first.**

## Problem

The SaaS pillar filter from #430 is not data-call-aware, so it hides Devices and Applications on **every** cycle, closed ones included. Those cycles collected all 40 questions and their answers still exist, so a historical SaaS questionnaire rendered 25 questions for a cycle that recorded 40, with no indication the other 15 existed — answered history the reader had no way to see.

## Change

**Filtering now starts at FY26 and applies to every cycle after it** (`src/utils/reducedPillarScope.ts`), gating the existing `filterPillarsForSystem` call in `QuestionnairePage` and `QuestionnareModal`.

A threshold, not "is this the latest call" — a latest-only gate would flip FY26 back to the full question set the moment FY27 opened, silently undoing the fix for a cycle already reported at 25. Later cycles qualify by deadline, never by `datacallid`, since ids are not chronological. Deliberately not "at or after the earliest FY26 deadline" either: an FY26 call mis-dated before a closed cycle would drag that cycle into scope. Mirrors `saasPillarScopeSQL` in #546. Every unknowable input falls back to the full set — showing an extra pillar only over-discloses to the form's author, while hiding one loses an answer.

**Also fixes the pillar-scores radar plotting a missing comparison pillar as `0`.** Scores are floored at 1.0, so 0 is not a reachable score: it drew the comparison line into the centre of the chart and read as a total collapse rather than "not measured that cycle". `previous` is now left undefined, which recharts renders as a gap, matching what the tiles and the comparison table already do. **This is already reachable on `main`** — any system whose environment changed between cycles has pillars in one call and not the other — and becomes unavoidable once phase 2 reduces SaaS scoring.

**Files:**
- `src/utils/reducedPillarScope.ts` (new) — `reducedPillarScopeApplies`
- `src/components/PillarScoresModal/radarData.ts` (new) — `buildRadarData`, `findComparisonPillarScore`; the chart, tiles and accessible table now share one pillarid pairing instead of each inlining it
- `QuestionnairePage.tsx`, `QuestionnareModal.tsx` — gate the filter; the modal derives the gate with `useMemo` outside its fetch effect, since depending on the `datacalls` array would re-run the effect on any context refetch and reset an open modal to question 1
- Tests for all three

## Acceptance criteria

- [x] Closed historical calls unchanged — **this is the criterion `main` violates**, and the reason for the PR. Historical cycles now show all six pillars with their recorded answers.
- [x] SaaS systems present exactly 25 questions, as seen by the ISSO, from FY26 onward. Still a client-side filter over an API that returns 40; CMS-Enterprise/ztmf#545 moves it into the questions endpoint, after which `filterPillarsForSystem` comes out.
- [x] All other environments unchanged (40) — `filterPillarsForSystem` short-circuits for any category but `SaaS`; this PR changes only *when* it is called.

The issue's remaining criteria are owned elsewhere: the `x/25` denominator by CMS-Enterprise/ztmf#546 (`progressColumn.tsx` renders what `/scores/progress` returns), and the API-side question set plus SaaS scoring by CMS-Enterprise/ztmf#545.

## Risk

Low. No API, schema, or data changes. The one behavior change outside the SaaS gate is the radar's missing-pillar rendering, which replaces a misleading zero with a gap.

## Verification

- `pnpm test` — 888 passing. `vite build` clean, lint clean, typecheck unchanged (2 pre-existing errors in `EmailModal.tsx` and the `ArrowIcon` JSX types, both untouched).
- `QuestionnairePage`: hidden on FY26; all six pillars on an earlier cycle; still hidden on a cycle *after* FY26; still hidden on FY26 once a later cycle has opened. The last two fixtures give the later cycle a **lower** `datacallid`, so an id-ordered threshold fails them too.
- `reducedPillarScopeApplies`: 16 cases — mis-dated FY26 call, name trimming, several FY26 cycles, and every unknowable input. `buildRadarData`: 6 cases including pillarid-not-position matching.
- Mutation-checked: reverting the page gate fails the historical-cycle test; restoring `?? 0` fails 4 radar cases; `>=`→`>` fails 5 helper cases; ordering by `datacallid` fails 3.
- Against a prod-shaped dev stack: a SaaS questionnaire shows 25 on FY2026 and all 40 on FY2025 Q3; a non-SaaS system is unchanged.

## UI

Screenshots are attached to CMS-Enterprise/ztmf-misc#289 rather than here, since they show live system data. What they cover:

- FY26 data call — SaaS systems show 25 questions.
- Historical data calls — SaaS systems show 40 questions total.
- A historical SaaS questionnaire — all 40 questions present, with the recorded answers.
- FY26 export of a SaaS system — only the 25 in-scope questions.

## Reviewer notes

- **CMS is deliberately not exempt.** This filter is OpDiv-blind and #546 does not exempt CMS either, so the two agree. When CMS confirms, both sides change together — exempting one alone would give CMS systems a 40 denominator with 15 questions their ISSOs cannot see.

### DIFF
--- a/src/components/PillarScoresModal/PillarScoresModal.tsx
+++ b/src/components/PillarScoresModal/PillarScoresModal.tsx
@@ -38,6 +38,7 @@ import { tierStyle, TIERS } from '@/utils/tierStyles'
 import { sortDatacallsByDeadline } from '@/utils/sortDatacallsByDeadline'
 import { parseDatacallName } from '@/utils/datacallGrouping'
 import { useContextProp } from '@/views/Title/Context'
+import { buildRadarData } from './radarData'
 
 interface PillarScoresModalProps {
   open: boolean
@@ -148,19 +149,13 @@ const PillarScoresModal: React.FC<PillarScoresModalProps> = ({
     (comparisonScoreEntry?.pillarscores?.length ?? 0) > 0
 
   // Prepare radar chart data
-  const radarData = useMemo(() => {
-    if (!hasValidData || !latestScore?.pillarscores) return []
-    return latestScore.pillarscores.map((pillar) => {
-      const comparisonPillarScore = comparisonScoreEntry?.pillarscores?.find(
-        (p) => p.pillarid === pillar.pillarid
-      )?.score
-      return {
-        pillar: pillar.pillar,
-        current: pillar.score ?? 0,
-        previous: comparisonPillarScore ?? 0,
-      }
-    })
-  }, [hasValidData, latestScore, comparisonScoreEntry])
+  const radarData = useMemo(
+    () =>
+      hasValidData
+        ? buildRadarData(latestScore?.pillarscores, comparisonScoreEntry)
+        : [],
+    [hasValidData, latestScore, comparisonScoreEntry]
+  )
 
   // Focus management for accessibility
   useEffect(() => {
--- a/src/components/PillarScoresModal/PillarScoresModal.tsx
+++ b/src/components/PillarScoresModal/PillarScoresModal.tsx
@@ -38,7 +38,7 @@ import { tierStyle, TIERS } from '@/utils/tierStyles'
 import { sortDatacallsByDeadline } from '@/utils/sortDatacallsByDeadline'
 import { parseDatacallName } from '@/utils/datacallGrouping'
 import { useContextProp } from '@/views/Title/Context'
-import { buildRadarData } from './radarData'
+import { buildRadarData, findComparisonPillarScore } from './radarData'
 
 interface PillarScoresModalProps {
   open: boolean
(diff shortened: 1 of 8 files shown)


MackOverflow (approved):
Looks clean to me!

jsos3-cms:
Landed in `main` as part of #691, squash commit `b830d21`. Your three commits went in unchanged.

The ordering you asked for was honoured: CMS-Enterprise/ztmf#546 merged and completed its prod deploy first, so the backend was already scoping FY26 SaaS progress to 25 before this reached prod.

No closing keyword was re-declared, deliberately. This is phase 1 of 2 against CMS-Enterprise/ztmf-misc#289, which stays open for #545, and cross-repository keywords would not fire regardless.

Closing as superseded. Thanks @costacalvin.

PR #694, opened 2026-08-11
chore(#693): serve the reduced-pillar rule from the API and drop the client-side filter

### LINKED ISSUE #693: Remove the client-side reduced-pillar filter now that the API owns the rule (ztmf#549)
## Problem

The reduced-pillar rule (SaaS systems answer 25 questions, not 40 — no Devices, no Applications & Workloads) is currently implemented **twice on the frontend**, and CMS-Enterprise/ztmf#549 moves the authoritative version into the API. Once that merges, both frontend copies are dead code that duplicates a rule which now lives in seeded data.

Background: the rule originates in CMS-Enterprise/ztmf-misc#289. Backend phase 1 (CMS-Enterprise/ztmf#546) scoped the progress counts and the export. Backend phase 2 (CMS-Enterprise/ztmf#549, closing CMS-Enterprise/ztmf#545) seeds the rule into a `reducedpillarscopes` table and adds an optional `datacallid` query param to `GET /fismasystems/{id}/questions`, so the endpoint serves the reduced set directly for in-scope cycles.

The frontend got its own implementation first, in two pieces added at different times:

- `src/utils/filterPillarsForSystem.ts` (#430) — drops Devices and Applications when the system's resolved category is `SaaS`.
- `src/utils/reducedPillarScope.ts` (#691) — decides whether a given cycle is in scope, by matching data-call names against a hardcoded `['FY2026', 'FY26']` prefix list. Its own docstring already says *"Interim - phase 2 moves this into the questions API."*
- `SAAS_EXCLUDED_PILLARS` in `src/constants.ts` — the hardcoded pillar names both rely on.

Two consumers wire them together: `QuestionnairePage.tsx` and `QuestionnareModal.tsx`.

Leaving this in place is not neutral. The frontend copy hardcodes the cycle names and pillar names that ztmf#549 exists to turn into data, so the two can drift the moment a rule row is added for another environment — the frontend would keep showing 40 for a system the API has already reduced. It also means every future change to the rule needs a coordinated release across both repos, which is exactly the coupling the seeded table was meant to remove.

## Deploy ordering

⚠️ **This must not ship before CMS-Enterprise/ztmf#549 is deployed.** The pre-#549 backend ignores query params on that endpoint entirely and returns the full 40-question catalog. With the frontend filter removed and the API not yet honoring `datacallid`, SaaS ISSOs would be served all 40 questions — the original ztmf-misc#289 bug, reintroduced.

## Mechanical checklist

- [ ] Pass `?datacallid=<the cycle being viewed>` on the questions request in `QuestionnairePage.tsx`, and render the response as-is
- [ ] Same in `QuestionnareModal.tsx`, using the latest call it already resolves from `/datacalls/latest`
- [ ] Delete `src/utils/filterPillarsForSystem.ts` and its test
- [ ] Delete `src/utils/reducedPillarScope.ts` and its test
- [ ] Delete `SAAS_EXCLUDED_PILLARS` from `src/constants.ts`
- [ ] Remove the context values and derived state that fall dead with them (`systemCategory`, `toCategoryMap`, and the now-unused `datacenterEnvironments` / `datacalls` / `latestDataCallId` destructures), including the stale entries left in the effects' dependency arrays
- [ ] Rewrite `QuestionnairePage.saaspillars.test.tsx` — it currently mocks 40 questions and asserts the client filters them, which no longer describes anything the frontend does

## Acceptance criteria

- [ ] No cycle name, environment name, or pillar name related to the reduced-pillar rule remains anywhere in `src/` — the rule is entirely server-side
- [ ] A SaaS system on an FY26-or-later cycle shows 25 questions; the same system on a pre-FY26 cycle shows all 40, with its answered history intact
- [ ] A non-SaaS system shows 40 on every cycle
- [ ] The questionnaire requests the cycle **being viewed**, not the latest one — viewing a closed FY26 cycle while FY27 is open must not ask the API about FY27
- [ ] Tests pin the `datacallid` actually sent, not just the pillars rendered, so a regression to an unparameterized request fails rather than silently falling back to the full catalog
- [ ] A test proves the response is not re-filtered client-side: feed the full set for an in-scope SaaS cycle and assert every pillar renders
- [ ] The progress fraction (`xx/25`) is unaffected — it comes from `/scores/progress`, which has been scope-aware since ztmf#546
- [ ] `yarn typecheck`, `yarn lint`, and `yarn test` clean

## Out of scope

`ScoreDiffModal.tsx` also calls this endpoint, and should **keep** calling it unparameterized. It uses the response only as a `functionid → pillar` lookup, cached per system across cycles; a reduced set would leave excluded-pillar diff rows unmapped and dumped into the fallback flat group. It wants the full catalog.

costacalvin:
Closes #693. Frontend half of Phase 2 CMS-Enterprise/ztmf-misc#289; pairs with CMS-Enterprise/ztmf#549.

## Ordering dependency, read this before merging

⚠️ **This must land after CMS-Enterprise/ztmf#549 is deployed.** The pre-#549 backend ignores query params on `GET /fismasystems/{id}/questions` entirely and returns the full 40-question catalog. With the client-side filter removed here and the API not yet honoring `datacallid`, SaaS ISSOs would be served all 40 questions — the original ztmf-misc#289 bug, reintroduced. There is no partial-credit order: frontend-first is strictly broken, not merely less useful.

## Problem

The reduced-pillar rule — SaaS systems answer 25 questions, not 40 — was implemented twice on the frontend, in two pieces added at different times:

- `filterPillarsForSystem` (#430) dropped Devices and Applications when the system's resolved category was `SaaS`.
- `reducedPillarScope` (#691) decided whether a cycle was in scope by matching data-call names against a hardcoded `['FY2026', 'FY26']` prefix list. Its own docstring said *"Interim - phase 2 moves this into the questions API."*
- `SAAS_EXCLUDED_PILLARS` in `constants.ts` held the pillar names both relied on.

ztmf#549 makes the rule seeded data and adds an optional `datacallid` param to the questions endpoint, so the API serves the reduced set directly. Keeping the frontend copy is not neutral: it hardcodes the cycle and pillar names the seeded table exists to replace, so the two drift the moment a rule row is added for another environment — the frontend would keep showing 40 for a system the API has already reduced.

## Change

Both consumers pass the cycle being viewed and render the response as-is. `QuestionnairePage` uses `activeDataCallId`; `QuestionnareModal` uses the latest call it already resolves from `/datacalls/latest`. Deleted alongside the two utils and the constant: `systemCategory`, `toCategoryMap`, and the now-unused `datacenterEnvironments` / `datacalls` / `latestDataCallId` destructures.

**One fewer fetch, and one fewer flash.** `systemCategory` was in the fetch effect's dependency array and resolved asynchronously from `systemInfo`, so the page fetched questions twice and briefly painted six pillars before re-rendering to four. Both are gone.

**A column swap worth naming.** The filter keyed on the resolved *category*; the API keys on *scoring_key*. Those are separate columns in `datacenterenvironments`. Checked all 17 rows against the dev database: they hold the same value everywhere except `DECOMMISSIONED`, whose NULL `scoring_key` already matched no functions and still lands on the same empty state. A no-op today, but a real coupling if the two are ever set differently.

**Threshold semantics.** The frontend's rule was "FY26-named, or deadline strictly after the last FY26 deadline"; the API's is "deadline at or after the seeded anchor." These differ only for a non-FY26 call sharing FY26's exact deadline. No deadline ties exist in the data, and the frontend's version disappears either way, so there is one less place for the two to drift.

## Pillar score modal

Also included, since it only became visible once systems started carrying four pillars: the tiles in `PillarScoresModal` are `md={2}`, six to a row, so the container being left-aligned never showed while every system had six. Four tiles fill 8 of 12 columns and left a two-tile gap on the right instead of sitting centered. `justifyContent="center"` is a no-op at six and also fixes `sm`, where four tiles at `sm={4}` wrap to a row of three plus a lone tile that hung left. `xs` is unaffected — two per row divides evenly either way.

## Acceptance criteria

- [x] No rule-related cycle name, environment name, or pillar name remains anywhere in `src/` — the rule is entirely server-side
- [x] A SaaS system on an FY26-or-later cycle shows 25 questions; the same system on a pre-FY26 cycle shows all 40, with its answered history intact
- [x] A non-SaaS system shows 40 on every cycle
- [x] The questionnaire requests the cycle **being viewed**, not the latest one
- [x] Tests pin the `datacallid` actually sent, not just the pillars rendered
- [x] A test proves the response is not re-filtered client-side
- [x] The progress fraction (`xx/25`) is unaffected — it comes from `/scores/progress`, scope-aware since CMS-Enterprise/ztmf#546, and no frontend code computes a denominator
- [x] `yarn lint` and `yarn test` clean; `yarn typecheck` reports the same 3 pre-existing errors as `main` (React JSX-element type mismatch in `EmailModal` and `ArrowIcon`), none from this change

## Testing

`QuestionnairePage.saaspillars.test.tsx` no longer described anything the frontend does — it mocked 40 questions and asserted the client filtered them. Rewritten to pin what the frontend now owns. The mock is keyed on both the system and the `datacallid`, and returns the **full** catalog when the param is absent, so a page that forgets it renders Devices and fails rather than passing by accident.

| Case | Pins |
| --- | --- |
| current / later / pre-anchor cycle, SaaS | the set rendered **and** the `datacallid` sent |
| viewed cycle vs latest | viewing closed FY26 while FY27 is open must not ask about FY27 |
| non-SaaS control × 3 cycle positions | only environments with a seeded rule row are reduced |
| full set served on an in-scope cycle | no surviving client-side filter |

The modal test asserts the container's computed `justify-content`; verified non-vacuous by reverting the component change, which fails it with `Expected: "center" / Received: ""`.

Full suite: 899 passed, 3 skipped (`main.test.tsx`, pre-existing), 0 failed.

Also exercised by hand against a local stack running ztmf#549: SaaS system 40 → 25 with the param and 40 without, pre-FY26 cycle unchanged at 40, and a non-SaaS control at 40 on the in-scope cycle.

## Out of scope

`ScoreDiffModal` also calls this endpoint and deliberately keeps calling it unparameterized. It uses the response only as a `functionid → pillar` lookup, cached per system across cycles; a reduced set would leave excluded-pillar diff rows unmapped and dumped into the fallback flat group. It wants the full catalog.


MackOverflow (approved):
Looks good to me, no nits!

PR #695, opened 2026-08-11
Fix questionnaire opening a different data call than the row's Data Call column

### LINKED ISSUE #662: Data Call Progress column and questionnaire icon can name different calls for a score-less system in a past-year view
### Description

On the Home dashboard, for a system that has a progress row but **no score** in the selected call, the **Data Call Progress column** and the **questionnaire icon** can resolve to different data calls. The column labels one call; clicking the questionnaire icon opens a different one. This is the exact failure `rowCall.ts`'s own docstring warns about (the column and the thing it opens naming different calls).

**Root cause.** The two surfaces resolve the row's call through different paths:

- **Column:** `resolveRowCallId` (`src/views/FismaTable/rowCall.ts`), whose first rung returns `chosenCallMap[fismasystemid]` (`FismaTable.tsx:899`). For a score-less system this is the chosen (per `buildDashboardMaps`) call.
- **Icon:** reads `systemCallMap[fismasystemid]`, which is correctly empty for a score-less system (the map is score-derived), so it falls through to `rowCallObjs[0] ?? datacalls.find(d => d.datacallid === activeDataCallId)` (`FismaTable.tsx:988-989`).

`activeDataCallId` collapses to the newest (open) call whenever more than one call is toggled on, and selecting a year toggles all of that year's calls. So the two agree whenever the open call is the newest selected call, but diverge in a past-year view: the column names the past call while the icon opens the current/open call.

**Not introduced by #657.** This behaves the same on `main`. Before the #655 fix (PR #657), score-less systems had no `chosenCallMap` entry either, and the column fell through to the same `activeDatacallIds`-based rung, producing the same past-vs-open split. #657 did not introduce this; it only made these systems visible in more views.

**Suggested fix.** Give the questionnaire icon the same fallback the column uses: resolve its target through `chosenCallMap` (or `resolveRowCallId`) rather than `activeDataCallId`, so both surfaces name the same call. Small change, but its own scope (touches #467 call-resolution, not the #655 count fix).

Refs #655.

### Steps to Reproduce

1. Log in as an admin/owner.
2. Select a **past year** in the data-call picker (so `activeDataCallId` points at the open call, not the year on screen).
3. Find a system that has a questionnaire (progress row) but no score in the selected past call.
4. Note the call named in the Data Call Progress column.
5. Click the questionnaire icon for that row.

### Expected Behavior

The questionnaire icon opens the same call the Data Call Progress column names.

### Actual Behavior

The column names the past call; the questionnaire icon opens the current/open call. The two disagree.

### Environment

Prod, Dev. Browser independent - logic is client-side (`FismaTable.tsx`, `rowCall.ts`).

### Acceptance Criteria

- [ ] For a score-less system in a past-year view, the questionnaire icon opens the same call the Data Call Progress column names.
- [ ] No regression in the common case (open call is the newest selected call).
- [ ] Test covering the past-year divergence case.

### Severity

Low - pre-existing, and `isNotUpdated` returns `false` for past calls, so the "Not updated only" filter never surfaces these rows in the view where the divergence is visible.

### Affected Component

Home dashboard - `src/views/FismaTable/FismaTable.tsx`, `src/views/FismaTable/rowCall.ts`

tarratsco:
## Summary

Closes #662.

On the Home dashboard, the Data Call Progress column and the questionnaire icon could resolve to different data calls for a system with a progress row but no score in the selected call. The column labels one call, while clicking the questionnaire icon opens a different one: the exact failure `rowCall.ts`'s docstring warns about.

Root cause: the two surfaces resolved the row's call through different paths.
* **Column:** `resolveRowCallId` (`chosenCallMap`-first).
* **Icon:** read the score-derived `systemCallMap`, which is empty for a score-less system, then fell back to `activeDataCallId`.

`activeDataCallId` collapses to the newest (open) call whenever more than one call is toggled on, and selecting a year toggles all of its calls. So the two agree when the open call is the newest selected call, but diverge in a past-year view: the column names the past call while the icon opens the open call. Pre-existing on main; #657 only made these systems visible in more views.

## Changes

* **`rowCall.ts` (`resolveQuestionnaireCall` new):** resolves the questionnaire icon's target through the same `resolveRowCallId` path the column uses, then looks up the `datacall` object. The column label and the call the icon opens now always name the same call.
* **`FismaTable.tsx`:** the questionnaire icon's `onClick` calls `resolveQuestionnaireCall` instead of the old `rowCallObjs[0] ?? activeDataCallId` fallback. The multi-call picker gate (`rowCallObjs.length > 1`) is untouched, so multi-scored systems still get the picker. (`openQuestionnaire` already defaults to `activeDataCallId` only when the call is undefined.)
* **`rowCall.test.ts`:** tests for the past-year divergence and the fallback rungs, plus an invariant test that `resolveQuestionnaireCall` matches `resolveRowCallId` for every input shape.

## Scope note

Beyond the literal score-less case, this also aligns a related pre-existing edge (flagged in review): a system scored in an older call but displayed against a newer progress-only call reads "Not scored" and names the newer call in the column. The icon now opens that same call rather than the hidden older score. Both are the same bug, and the single fix resolves them together.

## Acceptance criteria

* [x] For a score-less system in a past-year view, the questionnaire icon opens the same call the Data Call column names.
* [x] No regression in the common case (open call is the newest selected call).
* [x] Test covering the past-year divergence.

## Testing

* **`rowCall.test.ts`:** score-less past-year case opens the column's call (not the active one); the scored-but-displayed-newer case opens the displayed call; newest-in-view fallback matches the column; active-call-in-view unregressed; and an invariant test pinning icon == column across all input shapes.
* Full `FismaTable` suite green (94 passed); `tsc --noEmit` and `eslint` clean.
* Manually verified in dev against a two-call FY25 past-year view: on main, `MOCK-GARRISON`'s row named "FY25 ZTM" but the questionnaire breadcrumb opened "FY26 ZTM"; on this branch both read "FY25 ZTM".
* Behavior-only; no visual change.

### DIFF
--- a/src/views/FismaTable/FismaTable.tsx
+++ b/src/views/FismaTable/FismaTable.tsx
@@ -55,7 +55,11 @@ import { scoreSortValue } from './scoreHelpers'
 import { isSystemSelectable } from './rowSelection'
 import { ProgressCell } from './progressColumn'
 import { progressSortValue } from './progressHelpers'
-import { resolveRowCallId, datacallNameComparator } from './rowCall'
+import {
+  resolveRowCallId,
+  resolveQuestionnaireCall,
+  datacallNameComparator,
+} from './rowCall'
 import { fetchOpDivs } from '@/utils/opdivs'
 import { toCategoryMap } from '@/utils/dataCenterEnvironments'
 import { parseDatacallName } from '@/utils/datacallGrouping'
@@ -973,8 +977,14 @@ export default function FismaTable({
                     openQuestionnaire(
                       params.row.fismasystemid,
                       params.row.fismaacronym,
-                      rowCallObjs[0] ??
-                        datacalls.find((d) => d.datacallid === activeDataCallId)
+                      resolveQuestionnaireCall(
+                        params.row.fismasystemid,
+                        chosenCallMap,
+                        systemCallMap,
+                        datacalls,
+                        activeDataCallId,
+                        activeDatacallIds
+                      )
                     )
                   }}
                   color="inherit"
--- a/src/views/FismaTable/rowCall.test.ts
+++ b/src/views/FismaTable/rowCall.test.ts
@@ -1,5 +1,9 @@
 import type { datacall } from '@/types'
-import { resolveRowCallId, datacallNameComparator } from './rowCall'
+import {
+  resolveRowCallId,
+  resolveQuestionnaireCall,
+  datacallNameComparator,
+} from './rowCall'
 
 const call = (datacallid: number, deadline: string): datacall => ({
   datacallid,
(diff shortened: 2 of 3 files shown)


jsos3-cms (approved):
**GO.** Clean root-cause fix.

`resolveQuestionnaireCall` (`rowCall.ts:109-126`) routes the questionnaire icon through the same `resolveRowCallId` that already drives the Data Call column, deleting the old `rowCallObjs[0] ?? activeDataCallId` fallback rather than special-casing around it. That fallback was the mechanism — `rowCallObjs` is score-derived, so a score-less system fell straight through to `activeDataCallId`, which collapses to the newest open call whenever a whole year is toggled on.

It also picks up the related case the description calls out: a system scored in an older call but displayed against a newer one now opens the displayed call, matching the column.

I checked specifically for stale-closure risk, since that's the usual source of this bug class, and there isn't any. The handler computes inline at click time, and the `columns` array is a plain literal rebuilt every render, so there's no stale dependency-array trap. Confirmed against React Router's docs that `navigate(to, { state })` stores a fresh state object per call with no merge-with-previous behavior, so consecutive opens can't leak a prior call's id forward.

## Testing

Ran the full suite on `origin/main` as a control before attributing anything to this branch, since this suite is known to fail a chunk of tests under parallel load on a clean main. Both sides came back clean this run:

| | origin/main | this branch |
|---|---|---|
| Tests | 918 passed, 3 skipped | 923 passed, 3 skipped |
| Failures | 0 | 0 |

The +5 delta is exactly the new `resolveQuestionnaireCall` cases. `tsc --noEmit` and `eslint` are clean on the changed files, and `jest src/views/FismaTable` gives 94 passed / 0 failed, matching the description.

The invariant test asserting `resolveQuestionnaireCall(...)?.datacallid === resolveRowCallId(...)` across five input shapes is the strongest part of this. It pins the property that actually matters — icon and column can never disagree — rather than re-testing individual scenarios, so a future edit that reintroduces a second resolution path fails loudly instead of silently.

## Non-blocking

Nothing renders the grid, clicks the icon and asserts what `navigate` actually receives. The wiring at `FismaTable.tsx:977-988` into `openQuestionnaire` is covered only by the manual verification described in the PR. Unit coverage on the underlying logic is strong enough that I wouldn't hold the merge for it, but it'd be a good addition whenever that file next gets touched.

## Merge ordering

No file overlap with #694, so no conflict in either direction. They are complementary on the same data path though: #694 makes `QuestionnairePage` pass `location.state.datacallid` as `?datacallid=` to the questions endpoint, and this PR is what makes that id correct in the past-year and score-less cases. Worth landing this one first so the id #694 starts trusting is right — #694's own ordering constraint against backend ztmf#549 is separate and unaffected.

PR #696, opened 2026-08-11
chore(deps): phase 0 — version floors, unused dependency removal, and Storybook teardown

jsos3-cms:
Phase 0 of the dependency work. Two commits, both mechanical, no application source touched.

## Why this is broken into phases at all

The dependency backlog here spans everything from a stale lockfile entry to a four-major MUI jump. Landing that as one change would mean a single revert button for changes with wildly different blast radii, and a reviewer having to hold all of it at once.

So it is split by **what breaks if it is wrong**, not by package:

- **Phase 0 (this PR)** — changes that cannot alter application behavior. Version floors already permitted by existing ranges, and packages nothing imports. If any of this is wrong, the build fails; it cannot fail silently at runtime.
- **Phase 1** — patch bumps that touch declared versions, including unpinning `react-router` so its patches can flow. Still no API changes, but it edits versions the app actually resolves against.
- **Phase 2** — dev tooling majors (husky, lint-staged, testing-library, eslint). Breaks CI or git hooks if wrong; never the shipped bundle.
- **Phase 3** — framework majors (TypeScript, Vite, React, MUI, Apollo). Real migration surface, real user-visible risk, each needing its own PR and its own deploy.

The point of the split is that a reviewer of this PR does not need to think about MUI, and a reviewer of the MUI work does not need to re-litigate lockfile hygiene.

## What phase 0 does

**Commit 1 — raise two version floors.**

| Package | Before | After |
|---|---|---|
| `axios` | `^1.16.0` (locked 1.16.1) | `^1.19.0` (locked 1.19.0) |
| `tar` (resolution) | `^7.5.16` (locked 7.5.19) | `^7.5.21` (locked 7.5.22) |

Both were already permitted by their existing ranges — the lockfile had simply drifted behind. Raising the declared floors as well as the lockfile means a future resolution cannot land back on an older patch. That matters most for `tar`, where a resolution entry sitting below the current patch reads as pinned-safe while permitting an older resolution.

Note `tar` is a resolution-only entry, not a direct dependency, so `yarn up tar` fails with "doesn't match any packages referenced by any workspace" — it needs `yarn up -R tar`.

**Commit 2 — drop unused runtime dependencies and their orphaned types.**

`@apollo/client`, `@fortawesome/fontawesome-free`, `@hookform/resolvers`, `@mui/lab`, `@popperjs/core`, `classnames`, `clipboard-copy`, `jwt-decode`, `react-animate-height`, `react-cookie`, `react-dropzone`, `react-helmet`, `react-hook-form`, `react-is`, `react-popper`, `react-virtuoso`, `uswds`, `yup`.

Yarn additionally pruned `@types/react-helmet` and `@types/react-is`, which existed only to type packages in that list, so the declaration count is **twenty, not eighteen**. 55 packages and 22.32 MiB out of the install; runtime dependency count 33 → 15.

**This does not shrink the bundle.** Vite only bundles what is imported, so unused dependencies were never shipping. The win is install time, upgrade surface, and the number of advisories the tree can attract — not payload.

## What phase 0 deliberately does not do

- No version bump that crosses an API boundary.
- No change to any file under `src/` other than deleting the one story file in commit 3.
- **Storybook is removed** — see commit 3. This was originally carved out as a separate decision, and the discussion on this PR settled it: remove now, reintroduce deliberately if a future redesign wants it.

**Commit 3 — remove Storybook.**

It has never been runnable from a clean clone. `.gitignore` has excluded `.storybook/main.ts`, `.storybook/preview.tsx` and `.storybook/README.md` since the initial repo import, so the config was never committed and `storybook dev -p 6006` had nothing to start from for anyone cloning. It presumably worked for whoever set it up, on their machine only.

Removed: the ten supporting packages, the three non-functional scripts (`sb`, `storybook`, `build-storybook`), the Storybook block and plugin import in `eslint.config.mjs`, the single story file `src/hooks/stories/useAlert.stories.tsx` (unmodified since the initial import), and the three `.gitignore` entries — keeping those would silently re-hide the config if anyone reintroduces it.

This is the one part of the PR that touches `src/`, and it is a deletion.

## Adversarial verification

I tried to break this rather than confirm it, and two checks actually caught something:

1. **An import-pattern search is not sufficient.** A regex over import statements reports `react-dom` and `core-js` as unreferenced. Both are genuinely used — `react-dom/client` in `main.tsx:6`, `core-js/stable/atob` in `authLoader.ts:1` and `Title.tsx:14`. Subpath imports defeat the obvious check, so the removal list came from a substring search over every tracked file, and both packages stayed.

2. **Two `tsconfig.json` `types` entries were referencing packages nothing imports.** `yup` was one. No import search of any kind would have found it. `yarn typecheck` failed with `TS2688: Cannot find type definition file for 'yup'`, and the `types` entry is removed alongside the package. The other was `uuid` — and it is the sharper case, because neither `uuid` nor `@types/uuid` has ever been a *declared* dependency here. The only thing satisfying that entry was a transitive `@types/uuid` pulled in by `@storybook/addon-actions`, so removing Storybook exposed it. Both are the case for running the whole gate rather than trusting the search.

Also checked and clean: no SCSS/CSS `@import` of `uswds` or Font Awesome; no `index.html` references; no `jest.config.cjs` or `declaration.d.ts` references; no Vite aliases pointing at a removed package.

**Peers kept deliberately:** `@emotion/react` and `@emotion/styled` have no direct import but are required MUI peers.

**Transitive check:** `react-is` was a direct dependency at `^19.1.1` against React 18.3.1. After removal it still resolves for its real consumers — `@mui/base` takes 18.2.0, `@mui/material` takes 19.1.1 — so nothing was stranded, and the odd direct-dependency skew is gone.

## Deploy safety

The gate CI runs, plus the two build modes CI does not exercise on a PR:

| Check | Result |
|---|---|
| `yarn install --immutable` | clean — no package.json/lockfile drift |
| `yarn typecheck` | clean |
| `yarn lint` | clean |
| `yarn test:ci` | 66 suites, **904 passed**, 3 skipped, **0 failures** |
| `yarn build:dev` | built |
| `yarn build:prod` | built |
| `yarn build:impl` | built |

All three build modes are verified because this branch is what eventually deploys to each of them, and a PR only exercises `build:dev`.

The suite is two suites smaller than before (`filterPillarsForSystem.test.ts` and `reducedPillarScope.test.ts` were deleted by #694, which merged while this was open) — this branch has been updated onto `7e9d8f8` so it is verified against current `main`.

## Relationship to #697 and epic #179

Worth stating explicitly, because three things are moving in the same area.

**Epic #179** owns Dependabot *configuration* — `dependabot.yml`, the auto-merge workflow, CODEOWNERS, branch protection — across all three repos. This PR is not part of that and does not touch any of it.

**#697** is that automation's grouped security PR on this repo. The only overlap with this PR is **`axios`**: #697 targets `^1.18.0`, this targets `^1.19.0`. Everything else is disjoint — `tar` is not in #697's group, and removing unused packages is not something Dependabot can do at all.

**Merge this one first.** #697 rebases automatically, and against the higher floor here its `axios` entry becomes a no-op, leaving it as a clean set of exactly the bumps this PR does not cover.

I deliberately did **not** strip `axios` from this PR to force a cleaner split. Doing so would mean either reverting the lockfile to a known-vulnerable 1.16.1, or leaving a declared floor of `^1.16.0` against a locked 1.19.0 — a mismatch this PR exists to eliminate. Letting the automation rebase is the cheaper resolution.

## Rollback

Each commit reverts independently and neither touches application code, so a revert restores the previous resolution set exactly. Nothing here needs a data or infrastructure rollback.


jsos3-cms:
## Tier 0c — Storybook needs a decision, not a cleanup

Surfaced while doing the dependency removal in this PR, and deliberately left out of it. Recording it here so the exclusion is explained where the diff is, and so the decision has somewhere to happen. Nothing here blocks this PR — it is follow-on scope for the team.

### What is actually true today

**Storybook cannot run in this repo.** There is no `.storybook/` directory, so `storybook dev -p 6006` has no config to start from. The three scripts in `package.json` — `sb`, `storybook`, `build-storybook` — are non-functional as they stand.

Ten packages support it:

`storybook`, `@storybook/react`, `@storybook/react-vite`, `@storybook/addon-essentials`, `@storybook/addon-interactions`, `@storybook/addon-links`, `@storybook/blocks`, `@storybook/test` (all `^8.6.18`), plus `eslint-plugin-storybook` `^0.11.6` and `react-docgen-typescript` `^2.2.2`.

There is exactly **one story file**: `src/hooks/stories/useAlert.stories.tsx`. Its only commit is `2e63a91` "initial import from previous repo" (2024-07-29, Alex Tran) — it has never been modified since the repo was created.

`eslint.config.mjs` imports `eslint-plugin-storybook` and applies its recommended rules to `**/*.stories.{ts,tsx}`, so the config is wired even though the tool cannot start.

### Why this was not folded into the dependency cleanup

Everything else in phase 0 was provably unused: nothing imports it, and if that judgement were wrong the build would fail. Storybook is different — `useAlert.stories.tsx` is real source that imports `@storybook/react`, so removing the package breaks `yarn typecheck` unless that file is deleted at the same time.

That makes removal a three-part change touching a source file, the lint config and the scripts, resting on an assumption nobody has stated: that Storybook is not coming back. Bundling that into a lockfile-hygiene PR would have hidden the assumption inside a diff that otherwise cannot change behavior.

### The decision

Roughly, one of:

**A — Remove it.** Delete the ten packages, `useAlert.stories.tsx`, the Storybook block in `eslint.config.mjs`, and the three scripts. Cleanest tree; drops ten packages from the upgrade surface, including a Storybook 8 → 10 major that otherwise sits in phase 3. Loses the one story, though git history keeps it.

**B — Restore it.** Add a `.storybook/` config so the tooling works again, and treat the single story as the seed of a component library. Costs the config work plus keeping Storybook current — it is currently two majors behind (8.6.18 against 10.5.7).

**C — Leave it.** No work now. Cost is that ten unused packages stay in the tree attracting advisories and upgrade pressure, and the next person hits the same confusion.

### Worth knowing before choosing

This is **dev tooling only** — nothing here ships in the bundle or touches a deploy, so there is no urgency and no risk to any environment either way. The cost of leaving it is the ten packages' share of the dependency backlog, not any user-facing exposure.

Whichever way it goes, the three non-functional scripts should be either fixed or dropped, since a script that cannot run is a trap for the next person who tries it.

I have no strong preference between A and B. I would rank C last only because it is the option that guarantees someone rediscovers this.

MackOverflow:
My votes would be keep things simple with A

tarratsco:
I agree with Mack, we can keep it simple with A. Then during the re-design if we wanted to take advantage of storybook, we have the option.

jsos3-cms:
Option A it is — Storybook removed in `a5a5b81`, thanks both.

Two things surfaced while doing it that are worth recording, since neither was in the original writeup.

**It was gitignored, not missing.** `.gitignore` has excluded `.storybook/main.ts`, `.storybook/preview.tsx` and `.storybook/README.md` since the initial repo import. So the config was never committed — Storybook presumably worked for whoever set it up, on their machine, and has been unrunnable for everyone cloning since. I removed those three entries too; leaving them would silently re-hide the config if anyone reintroduces Storybook later, which is exactly the trap that produced this situation.

**It was propping up an unrelated tsconfig entry.** `tsconfig.json` listed `uuid` in its `types` array, and the only thing satisfying that was a transitive `@types/uuid` from `@storybook/addon-actions`. Neither `uuid` nor `@types/uuid` has ever been a declared dependency here and nothing imports it, so the entry was stale — it just happened to resolve. Typecheck caught it as TS2688 once Storybook was gone.

Removed in full: the ten packages, the three non-functional scripts, the Storybook block and plugin import in `eslint.config.mjs`, `src/hooks/stories/useAlert.stories.tsx`, the `.gitignore` entries, and the stale tsconfig entry.

Gate is green with exit codes checked rather than inferred: `yarn install --immutable`, `typecheck`, `lint`, `test:ci` (66 suites, 904 passed, 3 skipped, 0 failures), and `build:dev`, `build:prod` and `build:impl` all exit 0.

To @tarratsco's point about keeping the option open — nothing here forecloses it. Reintroducing Storybook later is `yarn add` plus a committed `.storybook/`, and it would start from a working config rather than an ignored one.

costacalvin (approved):
LGTM! Rebuilt local off of this branch and everything runs smoothly! Also verified Dev envhad a successful deploy. Removed deps all valid as well

Issue #677, opened 2026-08-05
Automated accessibility (508/WCAG) scan with isolated stack, modeled on Emberfall


jsos3-cms:
## Update, August 2026

The approach has changed since this was specced. Instead of assembling the scan in-repo from `accessibility-insights-action` plus a Playwright/axe-core harness, the scanning engine was built as a standalone tool: **Frostfall**, a single-binary accessibility scanner modeled on Emberfall (same YAML config style, same CLI/GitHub Action distribution, axe-core as the engine). It is in private development and pre-release hardening now; this issue becomes the tracker for wiring it into ztmf-ui once released, the same way Emberfall is wired into the backend.

What the tool already covers from the acceptance criteria below, out of the box:

- Scans real rendered routes including states behind interaction (modals, menus), not just page loads
- Baseline file with new-versus-known diffing (the `.a11y-baseline.json` item), so reports only surface regressions
- Severity thresholds, critical/serious surfaced prominently
- Self-contained HTML report with per-violation screenshots, uploadable as a PR artifact
- Non-blocking by default; enforcement is a config opt-in, which replaces the `continue-on-error` sprint-validation plan with something cleaner
- Isolated static-build serving (no dev-DB risk, same isolation rule as Emberfall)

What remains as ztmf-ui/ztmf integration work when it lands, replacing the original staged plan:

- [ ] `make test-a11y` target following the `install-emberfall` / `test-e2e` pattern (binary install with checksum verification, isolated `compose-test.yml` stack, JWT via `make generate-jwt`)
- [ ] CI workflow on PRs, report as artifact, non-blocking initially
- [ ] Commit `.frostfall.yml` covering the route set (dashboard, system detail, questionnaire, data calls, scores, admin) plus interaction-state tests (pillar scores modal is already validated in testing)
- [ ] Establish and commit the baseline
- [ ] README/onboarding docs for the target and the dev-DB isolation rule

Auth-gated coverage note: local scans work today via the existing `VITE_LOCAL_DEV` bypass; the tool's login-flow support is still in development and gates the full Stage 2 route set in CI.

The original spec is collapsed below for history. Its staged plan (accessibility-insights-action, test-a11y-public, in-repo Playwright harness) is superseded by the checklist above; its checkboxes are flattened so they do not count as open tasks.

---

<details>
<summary><b>Original spec (May 2026, superseded)</b></summary>

### Feature Type

General Feature

### Feature Type

General Feature

### Workflow Description

[...]

PR #700, opened 2026-08-12
chore(deps): phase 1 — security floors, react-router unpin, and unreferenced dependency removal

jsos3-cms:
Phase 1 of the dependency work, following #696. Two commits: raise the remaining security floors, then remove what nothing references.

Grouping the removals here rather than in phase 2 keeps the theme coherent — this PR is *stop carrying things*, and phase 2 becomes purely the large framework upgrades.

## What it clears

Eighteen of the twenty-one advisories currently open on this repo.

| Package | Scope | Alerts | Action |
|---|---|---|---|
| `brace-expansion` | runtime | 3 | resolution → `^1.1.18` |
| `js-yaml` | dev | 4 | scoped → `^3.15.1` and `^4.3.1` |
| `immutable` | runtime | 2 | resolution → `^4.3.9` |
| `lodash-es` | dev | 2 | resolution → `^4.18.0` |
| `postcss` | dev | 2 | resolution → `^8.5.23` |
| `react-router` | runtime | 1 | declared → `7.18.2` (with `react-router-dom`) |
| `glob` | dev | 1 | scoped → `^10.5.0` (10.x only) |
| `shell-quote` | dev | 1 | resolution → `^1.9.0` |
| `@babel/core` | dev | 1 | resolution → `^7.29.6` |
| `ip` | dev | 1 of 2 | resolution → `^2.0.1` |

## How this subsumes #697

#697 is the Dependabot security group for this repo. Everything it does correctly is done here, and three things it does are either dead or wrong:

**Its `axios` bump is already dead.** #696 raised axios to `^1.19.0`, past #697's `^1.18.0` target.

**Its `react-router` fix is incomplete.** It raises `react-router` to 7.18.2 but leaves `react-router-dom` at 7.18.0. Since `react-router-dom@7.18.0` depends on `react-router: '7.18.0'` **exactly**, the vulnerable 7.18.0 stays resolved underneath it — the alert closes while the vulnerable code still ships, and the tree carries two copies of a context-providing router in an app that imports from both packages. This PR bumps both, and the lockfile now contains a single `react-router@npm:7.18.2`.

**Its `@babel/core` bump is a major that fixes nothing.** It raises the declared `@babel/core` from `^7.29.7` to `^8.0.0`. But `^7.29.7` already resolved to 7.29.7, which is above the patched 7.29.6 — the declared dependency was never vulnerable. The alerted copy is a transitive 7.23.9 reached via `^7.11.6` / `^7.12.3`, and those ranges cannot resolve to 8.x. So that bump takes on a Babel major for zero security benefit and leaves the vulnerable copy untouched. This PR uses a `^7.29.6` resolution instead, which moves the transitive copy and takes no major.

It also covers eight advisories #697 does not touch at all: `@sigstore/core`, `glob`, `ip`, `lodash-es`, `sigstore`, and the extra `brace-expansion` advisory published today.

**Recommended handling for #697: leave it open, let it close itself.** Closing it now would only churn — Dependabot regenerates while the alerts persist, which is what happened when #584 was closed and returned as #697 hours later. Once this merges and the advisories resolve, it has nothing left to file.

## Why two of these are scoped rather than pinned by name

`glob` and `js-yaml` each have two major lines in the tree, and a blanket resolution would drag the other line across a major.

`glob` 7.2.3 arrives from `@jest/reporters` via `^7.1.3` and is **outside** the vulnerable range (`>= 10.2.0, < 10.5.0`), so only the 10.x descriptors are moved. Forcing 7.x to 10.5.0 would hand Jest a glob with no callback API.

`js-yaml` is vulnerable on both lines under separate advisories, so each is patched independently — `^3.13.1` → `^3.15.1` and `^4.1.x` → `^4.3.1`. A blanket `^4.3.1` would break the 3.x consumer, since js-yaml 4 dropped `safeLoad`.

The trade-off is real and worth stating: descriptor-scoped entries match specific declared ranges, so a future consumer declaring some other range will not be covered. That is the cost of not forcing a major on an unrelated package.

## What is deliberately left open

**`sigstore` and `@sigstore/core` cannot be fixed by resolution.** Patching needs `sigstore` 2.3.1 → 4.1.1 and `@sigstore/core` 1.1.0 → 3.2.1, but every consumer declares an incompatible range — `libnpmpublish@9.0.4` and `pacote@17.0.6` want `sigstore@^2.2.0`; `@sigstore/sign@2.3.2`, `@sigstore/verify@1.2.1` and `sigstore@2.3.1` want `@sigstore/core@^1.0.0`/`^1.1.0`. Forcing either would violate all of them. The real fix is upgrading the `semantic-release` chain that pulls them, which belongs with the other dev-tooling majors rather than here.

**One of the two `ip` advisories has no published patch** (`first_patched_version: none`), so no version bump can resolve it. Dev-scope. It needs either a dismissal with rationale or dropping the dependent.

So this leaves 3 of 21, each for a stated reason rather than an oversight.

## Commit 2 — remove unreferenced dev dependencies and the dead codegen config

Eight packages with no reference anywhere in tracked source, config, scripts or hooks, plus the config file one of them existed for.

| Removed | Why |
|---|---|
| `graphql` + `codegen.ts` | the config imports `@graphql-codegen/cli`, which is not installed, and targets a `localhost:3000/graphql` endpoint that does not exist. The Apollo client it paired with went in #696 |
| `dotenv` | not imported anywhere, including `vite.config.ts`. Vite reads `.env` natively |
| `camelcase`, `pretty-quick` | no references; formatting already runs through prettier via `lint:other` and lint-staged |
| `jest-date-mock` | not in `config/jest/setupTests.ts`, referenced nowhere |
| `jest-environment-node` | `jest.config.cjs` sets `testEnvironment: 'jsdom'`, and no file carries a `@jest-environment node` docblock |
| `babel-plugin-transform-import-meta`, `@types/babel__core` | there is no babel config in the repo; transforms run through `@swc/jest` |

These were checked against the places a plain search misses — `jest.config.cjs`, the `release` and `commitlint` blocks in `package.json`, the husky hooks, tsconfig's `types` array, and peer requirements. That is where most apparently-unused packages in this repo turn out to be wired, and it is how `yup` and `uuid` were caught in #696.

**`@axe-core/react` is kept** despite having no current reference, because #677 plans automated accessibility scanning and it is the plausible starting point.

**Three left alone deliberately, as decisions rather than cleanups:**

- **commitlint** — installed and configured, but there is no `commit-msg` hook and no workflow invokes it, so it enforces nothing today. That matters because semantic-release derives version bumps from conventional commits. Either wire it up or drop it; both are choices, not cleanup.
- **rimraf** — used only by the `clean` script. Every workflow is `ubuntu-latest`, so `rm -rf dist` would do, unless someone develops on Windows.
- **source-map-explorer** — no script, no references. A periodic bundle investigation that `yarn dlx` can run without a declared dependency.

## Verification

Exit codes checked rather than inferred from output:

| Check | Result |
|---|---|
| `yarn install --immutable` | exit 0 |
| `yarn typecheck` | exit 0 |
| `yarn lint` | exit 0 |
| `yarn test:ci` | exit 0 — 66 suites, **904 passed**, 3 skipped, 0 failures |
| `yarn build:dev` / `build:prod` / `build:impl` | all exit 0 |

Resolved versions confirmed against the lockfile afterwards, including that the 7.x `glob` and 3.x `js-yaml` lines were left where they were, and that only one `react-router` copy remains.

Tracked in CMS-Enterprise/ztmf-misc#295.


costacalvin (approved):
LGTM!

PR #701, opened 2026-08-14
Hoist opdivs into the shared Outlet context (dedupe GET /opdivs)

### LINKED ISSUE #558: Hoist opdivs into shared Outlet context (dedupe GET /opdivs across 5 call sites)
## Problem

`fetchOpDivs` (`src/utils/opdivs.ts`) is now called independently in **five** places, each firing a fresh `GET /opdivs` on mount with no shared cache:

| Call site | arg |
|---|---|
| `FismaTable.tsx` | `fetchOpDivs(true)` |
| `EditSystemModal.tsx` | `fetchOpDivs(false)` (active only) |
| `UserTable.tsx` | `fetchOpDivs(true)` |
| `OpDivAdmin.tsx` | `fetchOpDivs(true)` |
| `SystemDetailPage.tsx` | `fetchOpDivs(true)` (added in #557) |

The OpDiv list is small, static reference data. Refetching it on every page mount is redundant.

## Fix

Fetch once in `Title.tsx` and distribute via `useContextProp()`, the same pattern already used for `datacenterEnvironments` (#459) and `fismaSystems`. Consumers read the cached list off the Outlet context instead of hand-rolling their own fetch + abort + error handling.

### Wrinkle to handle

The call sites don't all want the same set: `EditSystemModal` uses `fetchOpDivs(false)` (active-only, for the assignment dropdown) while the rest use `true` (include inactive, so a system tied to a since-deactivated OpDiv still resolves its name). The shared context should fetch the `true` superset and let consumers filter to `active` client-side where they need active-only.

## Scope

- `Title.tsx` — fetch once, add to Outlet context value
- `Title/Context.ts` — add `opdivs` to the context type
- Five consumers above — drop the local `fetchOpDivs` effect, read from `useContextProp()`

## Notes

- Not a defect — the per-mount fetches all have correct abort/error/auth handling. This is a dedupe/cleanup.
- Surfaced as the "Possible Follow-up" in #557; filing so it's tracked.

costacalvin:
## Summary

Closes #558.

`fetchOpDivs` was called independently in six places, each firing a fresh `GET /opdivs` on mount with no shared cache. Fetches it once in `Title.tsx` and distributes it through `useContextProp()`, the pattern already used for `datacenterEnvironments` (#459) and `fismaSystems`. Not a defect fix — every per-mount fetch had correct abort/error/auth handling.

## Changes

* **`Title.tsx` / `Title/Context.ts`:** one `fetchOpDivs(true, signal)` gated on an active session; adds `opdivs`, `opdivsLoaded`, and `refreshOpdivs` to the context. `EditSystemModal` takes `opdivs` as a prop since it renders outside the Outlet.
* **Six consumers** drop their local fetch effect. `UserTable` loses four `useState`s and a 43-line effect.
* **`utils/opdivs.ts`:** `?? []` on the response — the backend serializes an empty result as JSON `null` on some endpoints (CMS-Enterprise/ztmf#346), and a `null` would now throw in every consumer at once.
* **`UserTable/opdivDerivations.ts` (new):** `UserTable`'s four OpDiv views moved out unchanged so they can be tested.

## The sixth call site, and `opdivsLoaded`

`QuestionnairePage` was added after #558 was filed so it isn't in the issue's table, but converting it is the difference between "deduped five sites" and "there is one fetch."

It needed one extra thing. Its insights gate keys on `null` meaning *not loaded yet* and blocks Next/Complete while pending, so the panel can't pop in after the user has advanced. Context `opdivs` starts `[]`, which is indistinguishable from loaded-and-empty — reading it directly would have opened the gate early. Hence `opdivsLoaded`, set in a `finally()` so a failed lookup resolves to "no insights OpDivs" rather than blocking forever, matching the old `.catch`.

## Re: the isolation-pattern comment (@jsos3-cms)

Applied, structurally rather than by copying #570's fix. The #532 poison needed a **mode-dependent** value: the Show Decommissioned toggle swaps the endpoint and replaces the shared `fismaSystems` array, so the picker inherited whatever the dashboard showed. `opdivs` can't do that — the context always holds the `includeInactive` superset, no consumer's UI state changes what is fetched, and no setter is exposed (only `refreshOpdivs`, which refetches the same superset). Each consumer narrows at point of use.

Worth recording: this branch was rebased off a stale WIP whose `UserTable` read `fismaSystems` from context and deleted the dedicated `/fismasystems` fetch — i.e. it would have reverted #570 and reopened #532. Resolving in `main`'s favour is what kept it out.

## Behaviour changes worth naming

* **Read-only admins now see OpDiv codes instead of raw ids.** Incidental, and a fix: `main` gated the fetch on `isAdmin` (write-tier) while `UserTable` admits `hasAdminRead`, so those roles hit the `?? id` fallback. Pinned by a test so it's deliberate.
* **A failed load is a toast, not a `console.error`.** With one fetch site there's no second mount to self-heal on, so an empty list persists for the session — Save stuck on the system form, OpDiv admin grid blank. Fixed text rather than `parseApiError`, which returns a generic message that tells the user nothing when fired from the shell. Cost: on the questionnaire route OpDivs feed only the insights gate, whose old catch deliberately failed silent. Accepted — the list is load-bearing on four other routes.
* **A cross-session OpDiv create no longer shows up without a reload.** `EditSystemModal` used to refetch on every open, so a newly created OpDiv appeared the next time you opened the form. Reading the shared list means it now appears on the next reload instead. Inherent to the dedupe, and narrow in practice: first-party creates are covered, since all three OpDiv-admin write paths call `refreshOpdivs`, which updates the list the modal reads. Only an OpDiv created in *another* session is affected. (Raised in review by @tarratsco.)

## Acceptance criteria

* [x] Fetched once in `Title.tsx`, added to the Outlet context value and to the context type
* [x] All five listed consumers drop their local `fetchOpDivs` effect (six deduped, not five)
* [x] The `includeInactive` superset is fetched; consumers needing active-only filter client-side
* [x] `EditSystemModal`'s active-only dropdown unchanged for the user
* [x] Exactly one `GET /opdivs` call site remains in `src/`
* [x] `yarn lint`, `yarn typecheck`, `yarn test` clean

## Testing

928 passed, 3 skipped (`main.test.tsx`, pre-existing). Three files had no coverage of the paths this touches, so they get it here:

| File | Pins |
| --- | --- |
| `Title/TitleOpdivs.test.tsx` (new, 5) | fetched **once** with `includeInactive` and reaching the context; skipped when unauthenticated; `refreshOpdivs` refetches; failure toasts; `opdivsLoaded` settles on failure |
| `utils/opdivs.test.ts` (new, 8) | envelope unwrap, the `active_only=false` param contract, `null` → `[]`, reject paths, `createOpDiv`/`updateOpDiv` |
| `UserTable/opdivDerivations.test.ts` (new, 10) | parent/inactive exclusion, OPDIV-tier narrowing, unscoped tiers keep the full set, read-only-admin labelling |

`Title.test.tsx` now mocks `fetchOpDivs` — the axios stub returned `undefined` from `get()`, so every test there was silently running the error path.

Two traps worth a reviewer's eye:

* The `QuestionnairePage` suites had **opposite** OpDiv defaults, and `OPDIV_ROWS[0]` is the fixture system's own OpDiv — putting it in the shared `makeCtx` default would have flipped the entire carried-forward suite to insights-enabled while still passing. Shared default is `[]`; the justification block opts in.
* `opdivDerivations.ts` is a move, not a rewrite. Verified with a throwaway differential test — previous inline bodies vs the new helpers over 5000 randomised inputs plus an edge matrix (empty lists, duplicate ids, every role × undefined/empty/matching/non-matching grants × both `isAdmin` values). Identical on every case.

## Out of scope

* **Dashboard datacall switch has no loading indicator.** `Home.tsx` initialises `loading` to `true` and never resets it when `activeDatacallIds` changes, so switching calls shows the previous call's scores until the refetch lands. Same symptom as the OpDiv grid overlay fixed here, unrelated cause. Filing separately.
* **Test-mock duplication.** The three `Title` suites now carry ~14 near-identical `jest.mock` calls each; this PR added the third. Extraction is awkward because `jest.mock` hoists per file.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -284,9 +284,10 @@ export type editSystemModalProps = {
   onClose: (data: FismaSystemType) => void
   system: FismaSystemType | null
   mode: string
-  // Datacenter-environment vocabulary for the dropdown. Passed from Title
-  // (the modal renders outside the outlet, so it can't read context).
+  // Datacenter-environment and OpDiv vocabularies for the dropdowns. Passed
+  // from Title because the modal renders outside the Outlet and can't read context.
   datacenterEnvironments: DataCenterEnvironment[]
+  opdivs: OpDiv[]
 }
 
 export type datacallModalProps = {
--- a/src/utils/opdivs.test.ts
+++ b/src/utils/opdivs.test.ts
@@ -1,0 +1,98 @@
+import MockAdapter from 'axios-mock-adapter'
+
+// Replace the app's axios instance with a bare one. The production module
+// reads import.meta.env at load time, which throws under @swc/jest.
+jest.mock('@/axiosConfig', () => {
+  const axios = require('axios').default
+  return { __esModule: true, default: axios.create({ baseURL: '/api/v1/' }) }
+})
+
+import axiosInstance from '@/axiosConfig'
+import { fetchOpDivs, createOpDiv, updateOpDiv } from './opdivs'
+import type { OpDiv } from '@/types'
+
+const mock = new MockAdapter(axiosInstance)
+afterEach(() => mock.reset())
+
+const ROWS: OpDiv[] = [
+  {
+    opdiv_id: 1,
+    code: 'CMS',
+    name: 'Centers for Medicare & Medicaid Services',
+    is_parent: false,
+    active: true,
+    system_delegate_enabled: false,
+  },
+  {
+    opdiv_id: 2,
+    code: 'RETIRED',
+    name: 'Deactivated OpDiv',
+    is_parent: false,
+    active: false,
+    system_delegate_enabled: false,
+  },
+]
+
+describe('fetchOpDivs', () => {
+  it('unwraps the { data: [...] } envelope', async () => {
+    mock.onGet('/opdivs').reply(200, { data: ROWS })
+    await expect(fetchOpDivs()).resolves.toEqual(ROWS)
+  })
+
+  it('defaults to active-only and sends active_only=false for the superset', async () => {
+    mock.onGet('/opdivs').reply(200, { data: ROWS })
+
+    await fetchOpDivs()
+    expect(mock.history.get[0].params).toBeUndefined()
+
+    await fetchOpDivs(true)
+    expect(mock.history.get[1].params).toEqual({ active_only: false })
+  })
+
+  it('coerces a null payload to an empty array', async () => {
+    mock.onGet('/opdivs').reply(200, { data: null })
+    await expect(fetchOpDivs(true)).resolves.toEqual([])
+  })
+
+  it('resolves an empty list as-is - empty is an answer, not a failure', async () => {
+    mock.onGet('/opdivs').reply(200, { data: [] })
+    await expect(fetchOpDivs(true)).resolves.toEqual([])
+  })
+
+  it('rejects on a server error', async () => {
+    mock.onGet('/opdivs').reply(500)
+    await expect(fetchOpDivs()).rejects.toBeDefined()
+  })
+
+  it('rejects on a network error', async () => {
+    mock.onGet('/opdivs').networkError()
+    await expect(fetchOpDivs()).rejects.toBeDefined()
+  })
+})
+
+describe('createOpDiv', () => {
+  it('posts the input and returns the created row', async () => {
+    mock.onPost('/opdivs').reply(201, { data: ROWS[0] })
+    await expect(createOpDiv({ code: 'CMS', name: 'CMS' })).resolves.toEqual(
+      ROWS[0]
+    )
+    expect(JSON.parse(mock.history.post[0].data)).toEqual({
+      code: 'CMS',
+      name: 'CMS',
+    })
+  })
+})
+
+describe('updateOpDiv', () => {
+  it('puts to the id-scoped path (204, no body)', async () => {
+    mock.onPut('/opdivs/2').reply(204)
+    await expect(
+      updateOpDiv(2, {
+        code: 'RETIRED',
+        name: 'Deactivated OpDiv',
+        active: false,
+      })
+    ).resolves.toBeUndefined()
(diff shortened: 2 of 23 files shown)


tarratsco (approved):
Approved!! Read the full diff and traced the risky paths; the refactor holds up well.

What I checked:

* Insights gate migration is faithful: `opdivsLoaded` set in `finally()` reproduces the old `.catch` semantics exactly (a failed lookup settles to "no insights OpDivs" instead of blocking Next/Complete forever), and the `TitleOpdivs` test pins the failure case specifically.
* `opdivDerivations` is a true move: the `isAdmin` gate, caller-scope narrowing, and code/label maps all survive intact, and the read-only-admin labeling change is deliberate and pinned by a test.
* The #570/#532 isolation stands: `UserTable`'s dedicated `/fismasystems` fetch is untouched, so the rebase hazard described in the body was in fact avoided.
* `?? []` in `fetchOpDivs` is load-bearing now that one fetch feeds the shared context: good that it landed with a test.
* Test coverage lands where the actual risk is (the provide side, previously untestable by construction), and the `Title.test.tsx` axios-stub fix is a nice catch of a pre-existing blind spot.

Two optional notes up  to your judgement, neither blocking:

1. `refreshOpdivs` has no in-flight sequencing, so two rapid OpDiv-admin writes could resolve out of order and briefly leave the older list (last-write-wins). This is the same characteristic as the old per-page fetches and it self-heals on the next write or reload, so a one-line comment acknowledging it is plenty. If OpDiv admin ever grows bulk operations, a request-sequence counter is a small fix at that point.
2. Worth a line in the body: `EditSystemModal` used to refetch on every open, so an OpDiv created in another session now will not appear until a reload. This is inherent to the dedupe, and first-party mutations are covered by `refreshOpdivs` (just naming the tradeoff).

